PR #18, opened 2022-04-11
Module addresses

tbwebb22:
Adds DAO module addresses to the data layer


src/daoData/useModuleAddresses.ts
```diff
@@ -0,0 +1,37 @@
...
+) => {
+  const [moduleAddresses, setModuleAddresses] = useState<string[]>();
+
+  useEffect(() => {
+    if (!daoContract || !accessControlContract) {
+      setModuleAddresses(undefined);
+      return;
+    }
+
+    const filter = accessControlContract.filters.ActionRoleAdded();
```
adamgall, line 17: I've got a hunch that because this doesn't use any indexed parameters, then using it with the `queryFilter` below is going to be very inefficient. Something to look out for as this gets ran on public networks.

adamgall, line 17: Should there also be logic here to _remove_ modules that have had all of their action roles removed?

src/daoData/useModuleAddresses.ts
```diff
@@ -0,0 +1,37 @@
+import { useState, useEffect } from "react";
+
+import { DAO, AccessControl } from "../typechain-types";
+
+const useModuleAddresses = (
```
adamgall, line 5: In here, can you please set up realtime listeners so that newly added (or removed) modules are captured and updated in realtime, without requiring a browser reload?

adamgall (changes requested):

adamgall (approved):

PR #31, opened 2022-04-12
Bugfix 20

christopherdancy:


src/web3/transactions.ts
```diff
@@ -26,23 +26,24 @@ export const useTransaction = () => {
...
           toast.dismiss(toastId);
           if (txReceipt.status === 0) {
             toast(failedMessage);
+            if (stoppedCallback) stoppedCallback();
           } else if (txReceipt.status === 1) {
             toast(successMessage);
             if (waitingCallback) waitingCallback();
           } else {
             toast(failedMessage);
+            if (stoppedCallback) stoppedCallback();
```
adamgall, line 42: Please spend some more time in this function.

Is this `else` block ever going to get hit? I don't think so, according to the docs here: https://docs.ethers.io/v5/api/providers/types/#providers-TransactionReceipt

What concept is being handled with this variable name `stopped`? I'm not sure what "stopped" means in the context of a transcation.

christopherdancy: I still need to go through and clean the entire transaction page up. My suggestion is this should be another PR.

adamgall: Agreed

christopherdancy: Will create this now

adamgall: @christopherdancy https://github.com/fractal-framework/fractal-interface/issues/33

adamgall (approved):
Approving with the understanding that the `useTransaction` hook will be cleaned up, tracking here: https://github.com/fractal-framework/fractal-interface/issues/33

src/web3/transactions.ts
christopherdancy: Oh cool - will delete the one I created

PR #47, opened 2022-04-20
Proposals

tbwebb22:
This PR renders all proposals that have been created for a given DAO on the DAO summary page. 

A useProposals hook has been added in the data layer which gathers data from emitted events (historical and realtime), and additional data from the GovernorModule. 

This hook creates an array of ProposalData objects, one for each proposal, with the following data:

export type ProposalData = {
  number: number;
  id: BigNumber;
  startBlock: BigNumber;
  endBlock: BigNumber;
  startTime: Date | undefined;
  endTime: Date | undefined;
  startTimeString: string | undefined;
  endTimeString: string | undefined;
  proposer: string;
  description: string;
  state: number | undefined;
  stateString: string | undefined;
  forVotesPercent: number | undefined;
  againstVotesPercent: number | undefined;
  abstainVotesPercent: number | undefined;
};

The ProposalCard components still need to be styled in a future PR. Currently, the ProposalCard components render on the DAO summary page as shown in the below screenshot:

image

### DIFF
--- a/src/components/DAOs/DAO/Proposals/ProposalCard.tsx
+++ b/src/components/DAOs/DAO/Proposals/ProposalCard.tsx
@@ -1,6 +1,5 @@
 import { ProposalData } from "../../../../daoData/useProposals";
 import { useState, useEffect } from "react";
-import { BigNumber } from "ethers";
 import { Link } from "react-router-dom";
 
 function ProposalCard({ proposal }: { proposal: ProposalData }) {
@@ -21,7 +20,7 @@ function ProposalCard({ proposal }: { proposal: ProposalData }) {
       ?.add(proposal.againstVotes)
       .add(proposal.abstainVotes);
 
-    if (totalVotes.eq(BigNumber.from("0"))) {
+    if (totalVotes.eq(0)) {
       setForVotesPercent(0);
       setAgainstVotesPercent(0);
       setAbstainVotesPercent(0);
@@ -30,19 +29,19 @@ function ProposalCard({ proposal }: { proposal: ProposalData }) {
 
     setForVotesPercent(
       proposal.forVotes
-        .mul(BigNumber.from("1000000"))
+        .mul(1000000)
         .div(totalVotes)
         .toNumber() / 10000
     );
     setAgainstVotesPercent(
       proposal.forVotes
-        .mul(BigNumber.from("1000000"))
+        .mul(1000000)
         .div(totalVotes)
         .toNumber() / 10000
     );
     setAbstainVotesPercent(
       proposal.abstainVotes
-        .mul(BigNumber.from("1000000"))
+        .mul(1000000)
         .div(totalVotes)
         .toNumber() / 10000
     );
--- a/src/components/DAOs/DAO/Proposals/ProposalCard.tsx
+++ b/src/components/DAOs/DAO/Proposals/ProposalCard.tsx
@@ -4,17 +4,12 @@ import { Link } from "react-router-dom";
 function ProposalCard({ proposal }: { proposal: ProposalData }) {
 
   return (
+    <Link to={`proposals/${proposal.number}`} className="underline">
     <div className="flex flex-col bg-gray-300 m-2 max-w-lg">
       <div>Proposal #{proposal.number}</div>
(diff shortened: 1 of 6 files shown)


src/components/DAOs/DAO/Proposals/ProposalCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+      proposal.abstainVotes === undefined
+    ) {
+      return;
+    }
+
+    const totalVotes = proposal.forVotes
+      ?.add(proposal.againstVotes)
+      .add(proposal.abstainVotes);
+
+    if (totalVotes.eq(BigNumber.from("0"))) {
```
adamgall, line 24: you can write this as

```
totalVotes.eq(0)
```

basically, once you have a `BigNumber` instance, the parameter to math operations on it (like `.eq(x)` or `.mul(x)`) just has to be `BigNumberish`, which includes numbers and strings.

src/components/DAOs/DAO/Proposals/ProposalCard.tsx
```diff
@@ -0,0 +1,66 @@
...
+        .div(totalVotes)
+        .toNumber() / 10000
+    );
+    setAbstainVotesPercent(
+      proposal.abstainVotes
+        .mul(BigNumber.from("1000000"))
+        .div(totalVotes)
+        .toNumber() / 10000
+    );
+  }, [proposal]);
```
adamgall, line 49: Please move these percentage calculations down into the data layer, or wherever the `ProposalData` object is being loaded/created.

src/components/DAOs/DAO/Proposals/ProposalsList.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const [{ proposals }] = useDAOData();
+
+  return (
+    <div>
+      <H1>
+        <EtherscanLink address={address}>
+          <span className="break-all">{address}</span>
+        </EtherscanLink>{" "}
+        Proposal List
+        <div className="grid grid-cols-3 gap-4">
```
adamgall, line 16: How does this look on various screen sizes?

src/components/DAOs/DAO/Summary.tsx
```diff
@@ -29,13 +30,23 @@ function ValidDAO({ address }: { address: string }) {
...
             Create Proposal
           </Link>
         </div>
         <div>name: {name}</div>
         <div>access control address: {accessControlAddress}</div>
+        <div className="grid grid-cols-3 gap-4">
+        {proposals?.map((proposal) => (
+          <ProposalCard key={proposal.number} proposal={proposal} />
+        ))}
+        </div>
```
adamgall, line 49: Could you just render the `ProposalsList` component here?

Why are there multiple places in the app (This DAO Summary page, and the Proposals index page) where the list of all proposals is displayed?

adamgall (changes requested):
Question

* Why is this branch targeting `develop` instead of `master`?

Request

* Can you please add some information / details / screenshots to the body of this PR, to give some context that someone (I) might need to assist in a review?
* Please link the relevant issues to this PR.

General comments

* Not going to fret about it now, but I find the use of `?` (when digging into an object) to be a smelly pattern. IMO, better to explicitly check if the data we expect is there, and _if not_, do something specific (like show a "loading" component or a "nothing here" component).

Thanks!

src/components/DAOs/DAO/Proposals/ProposalsList.tsx
```diff
@@ -0,0 +1,16 @@
+import { useDAOData } from "../../../../daoData";
+import ProposalCard from "./ProposalCard";
+
+function ProposalsList() {
+  const [{ proposals }] = useDAOData();
+
+  return (
+    <div className="grid grid-cols-1 md:grid-cols-2 lg:grid-cols-3 gap-4">
+      {proposals?.map((proposal) => (
```
adamgall, line 9: Just noting, we'll want to explicitly deal with UI when there are no proposals loaded yet

adamgall (approved):
Looks great, thank you @tbwebb22

PR #50, opened 2022-04-25
Proposal details

tbwebb22:
This PR adds components / routes to support a proposal details page. This page shows details about the currently selected proposal, and the vote counts of the proposal as shown in the screenshot below.

image

### DIFF
--- a/src/components/DAOs/DAO/Proposals/ProposalCardDetailed.tsx
+++ b/src/components/DAOs/DAO/Proposals/ProposalCardDetailed.tsx
@@ -1,15 +1,10 @@
 import { ProposalData } from "../../../../daoData/useProposals";
 import useDisplayName from "../../../../hooks/useDisplayName";
-import { useEffect } from "react";
 import CopyToClipboard from "../../../ui/CopyToClipboard";
 import Clock from "../../../ui/svg/Clock";
 
 function ProposalCardDetailed({ proposal }: { proposal: ProposalData }) {
   const proposerDisplayName = useDisplayName(proposal.proposer);
-
-  useEffect(() => {
-    console.log(proposal);
-  }, [proposal]);
 
   return (
     <div className="flex flex-col bg-gray-600 m-2 max-w-lg py-2 rounded-md">
--- a/src/components/DAOs/DAO/Proposals/ProposalCardDetailed.tsx
+++ b/src/components/DAOs/DAO/Proposals/ProposalCardDetailed.tsx
@@ -30,10 +30,7 @@ function ProposalCardDetailed({ proposal }: { proposal: ProposalData }) {
         <CopyToClipboard textToCopy={proposal.proposer} />
       </div>
       <div className="flex flex-row mx-4 text-gray-50">
-        Proposal ID:{" "}
-        {`${proposal.id.toString().substring(0, 4)}...${proposal.id
-          .toString()
-          .slice(-4)}`}
+        Proposal ID: {proposal.idSubstring}
         <CopyToClipboard textToCopy={proposal.id.toString()} />
       </div>
     </div>
--- a/src/components/DAOs/DAO/Proposals/ProposalCardDetailed.tsx
+++ b/src/components/DAOs/DAO/Proposals/ProposalCardDetailed.tsx
@@ -1,6 +1,8 @@
 import { ProposalData } from "../../../../daoData/useProposals";
 import useDisplayName from "../../../../hooks/useDisplayName";
 import { useEffect } from "react";
+import CopyToClipboard from "../../../ui/CopyToClipboard";
+import Clock from "../../../ui/svg/Clock";
 
 function ProposalCardDetailed({ proposal }: { proposal: ProposalData }) {
   const proposerDisplayName = useDisplayName(proposal.proposer);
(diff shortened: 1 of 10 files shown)


src/components/DAOs/DAO/Proposals/ProposalDetails.tsx
```diff
@@ -0,0 +1,39 @@
...
+import ProposalCardDetailed from "./ProposalCardDetailed";
+import ProposalVotes from "./ProposalVotes";
+
+function ProposalDetails() {
+  const params = useParams();
+  const [{ proposals }] = useDAOData();
+  const [proposal, setProposal] = useState<ProposalData>();
+
+  useEffect(() => {
+    if (proposals === undefined || params.proposalNumber === undefined) return;
```
adamgall, line 14: Please `setProposal(undefined)` if this condition evaluates to true, before `return`ing.

src/components/DAOs/DAO/Proposals/ProposalCardDetailed.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useEffect } from "react";
+import CopyToClipboard from "../../../ui/CopyToClipboard";
+import Clock from "../../../ui/svg/Clock";
+
+function ProposalCardDetailed({ proposal }: { proposal: ProposalData }) {
+  const proposerDisplayName = useDisplayName(proposal.proposer);
+
+  useEffect(() => {
+    console.log(proposal);
+  }, [proposal]);
```
adamgall, line 12: Please remove `console.log`s

src/components/DAOs/DAO/Proposals/ProposalCardDetailed.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <div className="flex flex-row mx-4">
+        <div className="text-gray-50">Created By:</div>
+        <div className="text-gold-500 ml-1">{proposerDisplayName}</div>
+        <CopyToClipboard textToCopy={proposal.proposer} />
+      </div>
+      <div className="flex flex-row mx-4 text-gray-50">
+        Proposal ID:{" "}
+        {`${proposal.id.toString().substring(0, 4)}...${proposal.id
+          .toString()
+          .slice(-4)}`}
```
adamgall, line 41: Can this logic please be moved out of this view layer, maybe into the `ProposalData` type?

src/components/DAOs/DAO/Proposals/ProposalDetails.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+    setProposal(proposals[parseInt(params.proposalNumber, 10)]);
```
adamgall, line 16: Is it implied that proposal numbers are always auto-incrementing? Something about this line doesn't sit quite right with me. Like, why are we using a property of a proposal to index into an array of proposals? Maybe this is fine but can you comment on this?

src/components/DAOs/DAO/Proposals/ProposalDetails.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  if (proposal === undefined) {
+    return <div>Proposal not found</div>;
+  }
+
+  return (
+    <div className="flex">
+      <ProposalCardDetailed proposal={proposal} />
+      <ProposalVotes proposal={proposal} />
+    </div>
```
adamgall, line 35: We'll want to come back and clean up the styling here (widths of things mostly), but that can happen in a later "style cleanup" PR.

adamgall (changes requested):

src/components/DAOs/DAO/Proposals/ProposalDetails.tsx
tbwebb22: So the proposals themselves (on-chain) don't have a number associated with them, just the ID. So to give the proposal a human readable number (counting upwards from zero), we are just using the order that the proposal created events are emitted to assign proposal numbers. 

We could add a proposal number to the ProposalData type, but it would still just be the same number as the index of the proposal in the proposals array.

adamgall: Where does the ID come from? Is it unique? What's the reason for needing a human readable number? Couldn't we just reference these things by their ID?

tbwebb22: The ID gets emitted in the proposal created event. They are unique, but are a full uint256, so not very human readable. 
I believe the design team decided that we should assign a human readable proposal number to each proposal to make things more user friendly.

adamgall: Got it. Ok, this makes sense.

I would still ask for the code that this comment thread references to be changed.

There's still this implicit assumption that the Proposal Number matches up with the index of that proposal in the `proposals` array. That works now, adds an implied constraint to the order of proposals in that array.

Can you please update this code so that you "search" through the `proposals` array for the proposal with the requested number?

Something like:

```js
const proposalNumber = parseInt(proposal.proposalNumber);
const proposal = proposals.find(p => p.proposalNumber === proposalNumber);
if (proposal === undefined) {
  // idk deal with it somehow
  return;
}
setProposal(proposal);
```

tbwebb22: Cool, addressed this in the most recent commit

adamgall (approved):

PR #58, opened 2022-04-26
Style search

christopherdancy:
image

image

image


src/components/DAOs/Search/DAOSearch.tsx
```diff
@@ -73,25 +77,34 @@ function DAOSearch() {
...
                   search
-                </Button>
+                </button>
               </div>
             </div>
+            {
+            searchFailed === false ? 
+            <p className="text-gray-50 pt-1 text-sm">Use a valid Fractal ETH address or ENS domain</p> : 
+            null
+            }
```
adamgall, line 100: can be written as

```
{ searchFailed === false && (
  <p className="text-gray-50 pt-1 text-sm">Use a valid Fractal ETH address or ENS domain</p>
)}
```

src/components/DAOs/Search/DAOSearch.tsx
```diff
@@ -73,25 +77,34 @@ function DAOSearch() {
...
+                  error = {searchFailed}
                   onChange={setSearchAddressInput}
                 />
               </div>
-              <div className="ml-2">
-                <Button
-                  disabled={false}
+              <div className="ml-1">
+                <button
+                  className="px-6 py-1 mx-2 border border-gold-500 rounded bg-gold-500 text-black-300 capitalize"
```
adamgall, line 89: Can this button (basically, its styling) be abstracted up the tree, so that it can be reused?

I anticipate that this app will settle upon a handful of different "looks" for buttons, and those styles should only be implemented once.

Styling a button like this, within another component, means that if we want a similar looking button elsewhere it will need to be re-styled elsewhere too.

src/components/DAOs/Search/DAOSearch.tsx
adamgall: but actually... see the engulfing comment

src/components/DAOs/Search/DAOSearch.tsx
```diff
@@ -73,25 +77,34 @@ function DAOSearch() {
...
+            null
+            }
           </form>
-          <Search searchAddress={searchAddress} />
-        </div>
+          <Search 
+          setSearchFailed={setSearchFailed}
+          searchAddress={searchAddress} 
+          />
+        </InputBox>
```
adamgall, line 106: Seems like there are two different components that are responsible for displaying the text beneath `InputAddress` component.

That seems messy. It makes it harder to get the styles/spacing consistent (see the little sizing/spacing jumps that happen when you first search for an invalid DAO), and puts multiple "things" in charge of one "thing".

Please consolidate this logic into one place. For example, what I would do, is move _all_ of the text display into the `Search` (and then `SearchingDAO`) component.

adamgall (changes requested):

adamgall:
@christopherdancy also please attach whatever relevant issues to this PR

src/components/ui/SearchButton.tsx
```diff
@@ -0,0 +1,13 @@
...
+      <button
+        className="px-6 py-1 mx-2 border border-gold-500 rounded bg-gold-500 text-black-300 capitalize"
+      >
+        search
+      </button>
+    </div>
+  );
+}
+
+export default SearchButton;
```
adamgall, line 13: Much closer, but still some conceptual problems here with "reusability".

First, you've named this component `SearchButton`, and hardcoded the word "search" into it. The intention with my previous request was to make it so that this "flavor" of button (meaning really just the colors and borders etc) could be reused anywhere in the app, regardless of what that button functionally represents in that context.

Next, the outer wrapping `div` with a left margin seems out of place. That's an assumption relating to "how this button is placed next to other elements" that I don't think this Button should make.

Look at the `Button.tsx` component. It makes no assumption about how it's being used, or the label on it, or how it should exist next to other components.

Perhaps move the `SearchButton` code into this existing `Button` component, and export multiple "flavors" of buttons from this file.

See https://github.com/decent-labs/decent-token-vesting/blob/master/src/components/ui/Input.tsx for an example from another project in which I did this for `input` components. There's a "base" `Input` component which isn't even exported, but implements functionality that all of the different "flavors" of inputs will use. Flavors in this "input" context represent inputting numbers, addresses, token amounts.

Something like that is what I'm looking for here.

src/components/ui/InputAddress.tsx
```diff
@@ -0,0 +1,31 @@
...
+    disabled={disabled}
+    placeholder={placeholder}
+    borderColor={error ? "border-red" : "border-black-100"}
+    onChange={e => onChange(e.target.value)}
+    onKeyDown={undefined}
+  />
+  )
+}
+
+export default InputAddress
```
adamgall, line 31: Please move this component into the existing `Input.tsx` file.

src/components/ui/Input.tsx
```diff
@@ -1,78 +1,43 @@
...
-  onChange: (newValue: string) => void,
-}) {
-  return (
-    <Input
-      title={title}
-      status={status}
+    <input
+      type={type}
+      min={min}
+      className={`${width} border ${borderColor} border-gray-20 bg-gray-400 rounded py-1 px-2 shadow-inner text-gray-50 ${disabled ? "disabled" : ""}`}
```
adamgall, line 27: Don't make `width` be an input to this component. Always make the `input` extend to fill its container. If you want to constrain the width of an `Input`, then do so at the place where it's being used (by, for example, wrapping the input in a `div` and giving that a width).

src/components/DAOs/Search/SearchingDAO.tsx
```diff
@@ -22,36 +20,49 @@ function SearchingDAO({
...
 }: {
   searchAddress: string | undefined,
   loading: boolean,
   validAddress: boolean,
   address: string | undefined,
   addressIsDAO: boolean,
   validDAOComponent: React.ReactNode,
+  setSearchFailed?: React.Dispatch<React.SetStateAction<boolean>>,
 }) {
+  if (loading !== true && searchAddress === undefined || searchAddress === "") {
```
adamgall, line 33: Warnings on this line, please use parentheses to make order of operations explicit

src/components/DAOs/Search/SearchingDAO.tsx
```diff
@@ -22,36 +20,49 @@ function SearchingDAO({
...
+    if(setSearchFailed) {
+      setSearchFailed(false);
+    }
+    return (
+      <p className="text-gray-50 pt-1 text-sm">Use a valid Fractal ETH address or ENS domain</p>
```
adamgall, line 38: There is still some spacing consistency issues as the text changes from this `p` tag, to the `SearchError` component.

src/components/ui/SearchButton.tsx
christopherdancy: With a new standard being developed by Taylor. This should be handled in a separate PR to keep consistency.

src/components/DAOs/Search/SearchingDAO.tsx
adamgall: @christopherdancy this is still a warning in the Typescript compiler

christopherdancy: for some reason I am not getting this error. Did you restart typescript server?

PR #59, opened 2022-04-26
Breadcrumbs

### LINKED ISSUE #42: Implement Breadcrumbs
https://github.com/icd2k3/use-react-router-breadcrumbs

As a part of this effort, it should include an "extended" header so that navigation (breadcrumbs) exists anywhere the header exists.

It would also be nice if that new "extended" component had a parameter to selectively hide the breadcrumbs when necessary (for example, on the home page).

adamgall:
# Implements Breadcrumbs

Closes #42 

* Breadcrumbs only show up when within a loaded DAO
  * `anyExcludeMatch` captures this -- don't even show the Breadcrumb bar if any of the given locations match
* When breadcrumbs _are_ displayed, "start" them at the "loaded DAO" level (exclude `Home`, and `DAOs`)
  * `excludePaths` captures this
* Don't display certain paths (e.g. `Proposals`) when showing breadcrumbs
  * path-specific `null` breadcrumbs capture this -- e.g. `{ path: '/daos/:address/proposals', breadcrumb: null },`
* Show dynamic information in Breadcrumbs -- e.g. DAO Name instead of the DAO's address (which is in the URL)
  * custom Breadcrumb render components capture this -- e.g. `DAOName` and `ProposalId` components
* Make current path white text, previous paths gold links

No Breadcrumb 👇
Screen Shot 2022-04-26 at 2 41 49 PM

No Breadcrumb 👇
Screen Shot 2022-04-26 at 2 42 09 PM

Just DAO Name on DAO homepage 👇
Screen Shot 2022-04-26 at 2 42 23 PM

Show "Details" on the DAO Details page 👇
Screen Shot 2022-04-26 at 2 42 26 PM

Show "New Proposal" on the New Proposal page (notice it's not `fr > Proposals > New`) 👇
Screen Shot 2022-04-26 at 2 42 30 PM

Show the Proposal Number on the Proposal detail page (notice it's not `fr > Proposals > 0`) 👇
Screen Shot 2022-04-26 at 2 42 34 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "react-scripts": "5.0.0",
         "react-toastify": "^8.2.0",
         "typescript": "^4.6.2",
+        "use-react-router-breadcrumbs": "^3.2.1",
         "util": "^0.12.4",
         "web-vitals": "^2.1.4",
         "web3modal": "^1.9.5"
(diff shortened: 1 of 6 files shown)


christopherdancy (approved):
Works as expected GTG

PR #74, opened 2022-04-28
Uplift search button

christopherdancy:
In this PR:
I extended Taylor's Button component to included styling for color. I do not like this pattern due to each Button component requiring its own styling... This is against the idea of reusability. 

I have also removed the Search button as required by this issue criteria.


tbwebb22 (approved):
Looks good to me

PR #80, opened 2022-04-28
Cast votes

tbwebb22:
This PR includes:
 - A new CastVote component that is rendered on the ProposalDetails page which allows a user to vote on an open proposal
 - The CastVote component only allows a user to vote if the proposal is open, and they have not already voted
 - The CastVote component shows a user what their vote was if they have already voted on the proposal
 - A useCastVote hook that handles the transaction
 - Additional hooks in useProposals which handles getting the user vote data

The styling of the CastVote component should be refined in a future PR

image


src/components/DAOs/DAO/Proposals/CastVote.tsx
```diff
@@ -0,0 +1,77 @@
...
+  // Vote Enum
+  // Against => 0
+  // For => 1
+  // Abstain => 2
+  const [newVote, setNewVote] = useState<number>();
+  const [pending, setPending] = useState<boolean>(false);
+  const [voteButtonString, setVoteButtonString] = useState<string>();
+
+  useEffect(() => {
+    if(proposal.stateString === "Closed") {
```
adamgall, line 19: might be less error prone (in the future, as we localize this app) to use the `state` instead of `stateString` to do comparison here.

src/components/DAOs/DAO/Proposals/DelegateVote.tsx
```diff
@@ -0,0 +1,38 @@
...
+          helperText="The address to delegate your votes to"
+          disabled={false}
+        />
+        <Button onClick={() => delegateVote()}>Delegate</Button>
+      </div>
+    </>
+  );
+}
+
+export default DelegateVote;
```
adamgall, line 38: Should this file be here? It's not being referenced anywhere else in this PR.

src/components/DAOs/DAO/Proposals/ProposalDetails.tsx
```diff
@@ -43,6 +44,7 @@ function ProposalDetails() {
     <div className="flex">
       <ProposalCardDetailed proposal={proposal} />
       <ProposalVotes proposal={proposal} />
+      <CastVote proposal={proposal} />
```
adamgall, line 47: Can you please try to make this look more like the rendering?

It should be below the Proposal Details
The font in the buttons is wrong
The arrow should be right aligned
There shouldn't be lines between all of the buttons
etc

Screen Shot 2022-04-28 at 4 23 04 PM

src/daoData/useDeployDAO.ts
```diff
@@ -139,7 +139,7 @@ const useDeployDAO = ({
           ), // Timelock Impl
           abiCoder.encode(["string"], [""]),
           abiCoder.encode(["uint64"], [BigNumber.from("0")]), // vote extension
-          abiCoder.encode(["uint256"], [BigNumber.from("6545")]), // voteDelay - 1 day
+          abiCoder.encode(["uint256"], [BigNumber.from("0")]), // voteDelay - 1 day was 6545 blocks
```
adamgall, line 142: What's the significance of this changing from "6545" to "0"?

src/components/DAOs/DAO/Proposals/CastVote.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const castVote = useCastVote({
+    proposalId: proposal.id,
+    vote: newVote,
+    setPending,
+  });
+
+  return (
+    <>
+      <Pending message="Casting Vote..." pending={pending} />
+      <ConnectModal />
```
adamgall, line 37: There is going to be an effort soon to get rid of the full screen semi-transparent modal things, so let's just remove it from here now.

adamgall: Same with that Connect Modal

src/components/ui/SelectVoteButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+      type={type}
+      {...rest}
+    >
+      {children}
+      {selected ? <Check /> : null}
+    </button>
+  );
+}
+
+export default SelectVoteButton;
```
adamgall, line 34: I'd love to see this custom button inherit from a more global Button that has global styles defined.

src/components/ui/SelectVoteButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+          ? "border-gold-500 bg-black-500 text-gold-500"
+          : "border-gray-25 bg-chocolate-500 text-gray-25"
+      } ${disabled ? "opacity-50" : ""}`}
+      disabled={disabled}
+      onClick={onClick}
+      type={type}
+      {...rest}
+    >
+      {children}
+      {selected ? <Check /> : null}
```
adamgall, line 29: Can be written as

```
{selected && <Check />}
```

src/daoData/useProposals.ts
```diff
@@ -133,21 +156,60 @@ const useProposals = (moduleAddresses: string[] | undefined) => {
...
+        startTimeString: proposal.startTimeString,
+        endTimeString: proposal.endTimeString,
+        proposer: proposal.proposer,
+        description: proposal.description,
+        state: proposal.state,
+        stateString: proposal.stateString,
+        forVotesPercent: proposal.forVotesPercent,
+        againstVotesPercent: proposal.againstVotesPercent,
+        abstainVotesPercent: proposal.abstainVotesPercent,
+        userVote: userProposalVote ? userProposalVote.vote : undefined,
```
adamgall, line 205: Much like `state` and `stateString`, probably a good idea to have `userVote` and `userVoteString` here.

Elsewhere I saw some logic that is checking the value of `userVote` to do things (ex: line 43 of CastVote.tsx), and it's not very future proof (for when we go and localize this app). Better to use actual data for comparisons, and english strings (in the future, any language strings) _just_ for display purposes.

adamgall (changes requested):

adamgall (changes requested):
Forgot to mention in preview review, please merge `master` into this branch

src/components/DAOs/DAO/Proposals/CastVote.tsx
```diff
@@ -0,0 +1,77 @@
...
+          onClick={() => setNewVote(2)}
+          selected={newVote === 2 || proposal.userVote === "Abstain"}
+          disabled={proposal.stateString !== "Open" || proposal.userVote !== undefined}
+        >
+          Abstain
+        </SelectVoteButton>
+        <hr className="mx-2 my-2 border-gray-200" />
+        <Button
+          onClick={() => castVote()}
+          disabled={newVote === undefined || proposal.stateString !== "Open" || proposal.userVote !== undefined}
```
adamgall, line 67: I was able to cast a vote, even though I had no tokens and my votes weren't delegated anywhere.

Apparently the contracts just let it happen without reverting the transaction.

But, our UI should be smarter than that.

Please only make sure this button (probably all the buttons) has additional checks to only allow them to be interactive if it makes sense for that user.

adamgall (changes requested):

src/components/DAOs/DAO/Proposals/DelegateVote.tsx
tbwebb22: Nope, that shouldn't have been there. Fixed

src/daoData/useDeployDAO.ts
tbwebb22: This changes it so that we can vote on a newly created proposal immediately instead of waiting a day, for testing purposes.

tbwebb22: Of course we will want to change this back for prod

adamgall: Should (could?) this be a user input when making a new Proposal?

Or

Why not leave it at "no delay" on mainnet?

src/components/DAOs/DAO/Proposals/CastVote.tsx
tbwebb22: Removed both

src/components/DAOs/DAO/Proposals/CastVote.tsx
tbwebb22: Updated so that user cannot vote if they have a vote power of 0 for this proposal

src/daoData/useDeployDAO.ts
tbwebb22: This parameter is specific to the Governor module, so it can't be changed per proposal. 

We could leave it at no delay on mainnet for sure. I think it could be argued though that its good to have a delay between when a proposal is created and voting can begin.

src/components/DAOs/DAO/Proposals/ProposalDetails.tsx
tbwebb22: Update styling according to new screen from design team:

Current styling:
image

Design team screen:
image

adamgall: Looking much better thank you.

First thing that stands out is that the "Cast Vote" and "Results" components should definitely be sharing some container-level styling (so that the title and horizontal line are on the same latitude.

But we can tackle that in a later PR. Let's get this merged.

src/daoData/useDeployDAO.ts
adamgall: ohhh i'm sorry, i didn't realize this was in the `useDeployDAO` hook. so this happens when a DAO is deployed. got it.

then I ask one more thing here. throw the comment `TODO: set back to 6545 before mainnet` in here.

So that we can search the codebase for `TODO` later and deal with things.

src/components/DAOs/DAO/Proposals/ProposalDetails.tsx
```diff
@@ -40,9 +41,12 @@ function ProposalDetails() {
   }
 
   return (
-    <div className="flex">
+    <div className="flex flex-col">
       <ProposalCardDetailed proposal={proposal} />
+      <div className="flex flex-row">
+      <CastVote proposal={proposal} />
       <ProposalVotes proposal={proposal} />
+      </div>
```
adamgall, line 49: I was expecting this to look cramped on mobile... but honestly it't not too bad!

src/components/DAOs/DAO/Proposals/CastVote.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <div className="flex mx-2 my-2 text-gray-25">Cast Vote</div>
+        <hr className="mx-2 my-2 border-gray-200" />
+        <SelectVoteButton
+          onClick={() => setNewVote(1)}
+          selected={newVote === 1 || proposal.userVote === 1}
+          disabled={
+            proposal.state !== 1 ||
+            proposal.userVote !== undefined ||
+            proposal.userVotePower === undefined ||
+            proposal.userVotePower.eq(0)
```
adamgall, line 47: Extract this out to a function (or better, some react state :)) cause it's the same for all buttons.

edit: ah fuck it let's just leave as is for now and handle later

adamgall: In a future PR, let's give some TLC to concepts across the data layer around "when should things be able to be `undefined`"

adamgall (approved):

PR #81, opened 2022-04-29
User is able to queue a transaction after a successful vote

christopherdancy:
- Completed Tasks:
- A Queue transaction component was created and added to the Proposal detail page.
- A user has the ability to queue a transaction. This transaction will update the state of the proposal.
- Included with this Main feature:

1. Deeper understanding of state is present within requests. Users will know if a proposal is active, succeeded, failed, queued or is executed.
2. applied content Box styling.
3. Created a hook which updates once the transaction is created


adamgall (approved):
SHIP IT

PR #113, opened 2022-04-29
Create/form button

Da-Colon:
## Overview

Creates New Button Components. Creates a new Button component that handles different button state (hover, disabled, loading) export styled components Primary, Secondary as well as a plain Text button. 

## Notes

- Added `classnames` library as discussed earlier in call. For reference https://www.npmjs.com/package/classnames .
  - Well need to run `npm run start` as a result
- Icons will be passed in as a JSX.Element from the `.svg` component folder

### Primary Button
- Disabled
image

- Active
image

- Loading
image

- All of the screenshots of buttons will be on Search Fractal page for viewing of the Different Button styles. 
### Secondary button
image

### Text Button
image

- Primary with Icon right
image

- Primary with Icon left
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "@types/react-dom": "^17.0.13",
         "@ukstv/jazzicon-react": "^1.0.0",
         "@walletconnect/ethereum-provider": "^1.7.3",
+        "classnames": "^2.3.1",
         "ethers": "^5.5.4",
         "process": "^0.11.10",
         "react": "^17.0.2",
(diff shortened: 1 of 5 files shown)


Da-Colon:
@adamgall updated with changes to loader so that the loading 'text' is absolute and sits over the 'hidden' label

Da-Colon:
I should also add instead of the word 'Loading' I used 3 asterisks `* * *`

adamgall (approved):
Looking very good thank you @Da-Colon

PR #114, opened 2022-04-29
Execute transaction

tbwebb22:
This PR adds functionality to support executing a proposal that has been passed and queued.

This includes:
 - A ProposalExecute component that gets rendered on the ProposalDetailed component when the proposal is able to be executed
 - Adding a proposal eta parameter to the proposal data object for checking whether a proposal is able to be executed
 - A useExecuteTransaction hook that handles the execute transaction

image


src/hooks/useBlockNumber.ts
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    const timer = setInterval(() => {
+      provider.getBlockNumber()
+      .then((blockNumber) => {
+        setBlockNumber(blockNumber)
+      })
+      .catch(console.error);
+    }, 1000);
+    return () => clearInterval(timer);
```
adamgall, line 21: I don't think you'd want to set a timer in here, especially not one that fires every second, since blocks show up on average like 12-15 seconds.

You can set an event listener on the `provider` to get notified when blocks come in.

Lil something like this: https://github.com/decent-labs/decent-token-vesting/blob/master/src/data/time.ts#L4

src/hooks/useBlockTimestamp.ts
```diff
@@ -0,0 +1,29 @@
+import useBlockNumber from "./useBlockNumber";
+import { useEffect, useState } from 'react';
+import { useWeb3 } from "../web3";
+
+const useBlockTimestamp = () => {
+  const [blockTimestamp, setBlockTimestamp] = useState<number>();
```
adamgall, line 6: Perhaps the initial state of this should be Javascript's time, so there isn't that moment at the start of the app where time is undefined.

src/hooks/useBlockTimestamp.ts
```diff
@@ -0,0 +1,29 @@
...
+  const { provider } = useWeb3();
+  const blockNumber = useBlockNumber();
+
+  useEffect(() => {
+    if(provider === undefined || blockNumber === undefined) {
+      setBlockTimestamp(undefined);
```
adamgall, line 12: Perhaps the initial state of this should be Javascript's time, so there isn't that moment at the start of the app where time is undefined.

src/hooks/useBlockTimestamp.ts
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    const timer = setInterval(() => {
+      provider.getBlock(blockNumber)
+      .then((block) => {
+        setBlockTimestamp(block.timestamp)
+      })
+      .catch(console.error);
+    }, 1000);
+    return () => clearInterval(timer);
```
adamgall, line 23: Seems overkill to make this `getBlock` network request every second. Perhaps, every second you just increase the current time by 1 second, and let the `blockNumber` updating be your trigger to make that network request to get the new block's timestamp.

Right now, the app will only "know what time it is" on the order of every block updating the time by like 15 seconds at a time. The way I described above, the app could count up second by second, and "realign" itself with every block.

Something like this: https://github.com/decent-labs/decent-token-vesting/blob/master/src/data/time.ts#L32

adamgall: I would also recommend you move these hooks into the global data layer, so that if multiple components need to know the time, this code and these network requests and listeners are only instantiated once.

src/daoData/useProposals.ts
```diff
@@ -80,7 +81,10 @@ const getTimestampString = (time: Date | undefined) => {
   );
 };
 
-const getBlockTimestamp = (provider: providers.BaseProvider | undefined, blockNumber: number) => {
+const getBlockTimestamp = (
+  provider: providers.BaseProvider | undefined,
+  blockNumber: number
+) => {
```
adamgall, line 87: I know nothing really changed here, but perhaps this function could be turned into a hook and placed into the `hooks` directory. It could then use the `useWeb3` hook to grab the `provider` so that could be removed from the input params.

src/components/DAOs/DAO/Proposals/ProposalExecute.tsx
```diff
@@ -0,0 +1,41 @@
...
+import PrimaryButton from "../../../ui/PrimaryButton";
+import useBlockTimestamp from "../../../../hooks/useBlockTimestamp";
+
+function ProposalExecute({ proposal }: { proposal: ProposalData }) {
+  const [show, setShow] = useState<boolean>(false);
+  const blockTimestamp = useBlockTimestamp();
+
+  useEffect(() => { 
+    if(proposal.eta === undefined || blockTimestamp === undefined) {
+      return;
```
adamgall, line 13: `setShow(false)` before return to reset the default state here please.

src/components/DAOs/DAO/Proposals/ProposalExecute.tsx
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    if(proposal.eta !== 0 && proposal.eta < blockTimestamp) {
+      setShow(true);
+    }
```
adamgall, line 18: `else { setShow(false) }`

or

`setShow(proposal.eta !== 0 && proposal.eta < blockTimestamp)`

src/components/DAOs/DAO/Proposals/ProposalExecute.tsx
```diff
@@ -0,0 +1,41 @@
...
+        Proposal ready for execution
+      </div>
+      <div className="flex flex-grow justify-end mx-4">
+        <PrimaryButton onClick={executeTransaction}>
+          Execute Proposal
+        </PrimaryButton>
+      </div>
+    </div>
+  );
+}
```
adamgall, line 39: Is there no indication to the user that a proposal has _already_ been executed?

adamgall (changes requested):

adamgall:
Please also add relevant Issues to this PR

src/hooks/useBlockTimestamp.ts
tbwebb22: Why would we want to increment the time every second, ultimately all we care about is the timestamp of the last mined block, since that determines when state has updated last, correct?

adamgall: I'm just trying to think ahead a bit.

It might be good for the webapp to have a clock that _does_ increment every second.

Imagine the scenario:

* blocks come every 15 seconds
* block `n` is mined _right now_, timestamp is 100
* there's a proposal that _can be_ executed at timestamp 105
* if all we do is look at last block's timestamp, then the webapp won't allow user to make the transaction until block `n + 1` is mined (at timestamp 115), and the earliest block that transaction will get into is block `n + 2` (at timestamp 130).
* if we optimistically keep track of the time per second, the transaction button will enable at timestamp 105, user makes transaction, it could get into block `n + 1` (timestamp 115)

src/components/DAOs/DAO/Proposals/ProposalExecute.tsx
tbwebb22: If a proposal has already been executed, then this component should just return null.
However, the proposal state will be updated to "Executed" and shown on the Proposal Detail card.

src/daoData/useProposals.ts
adamgall: ignore this, too complicated for now

src/hooks/useBlockTimestamp.ts
tbwebb22: Updated hook to increment time every second, and still get timestamp updates as new blocks are found

src/daoData/daoData.ts
```diff
@@ -48,7 +52,9 @@ export const useDAODatas = () => {
     proposals,
     governorModuleContract,
     tokenContract,
-    tokenData
+    tokenData,
+    currentBlockNumber,
```
adamgall, line 56: This `currentBlockNumber` property probably doesn't (yet) need to be included in the exported data object (don't think it's being used in any components), but it's not doing any harm.

adamgall (approved):

PR #118, opened 2022-04-30
Proposals Update

Da-Colon:
## Overview

Updates Proposals to match designs styles.

## Note
- I did add a shadow to the ContentBox, not in designs. let me know if there are any concerns

## Changes
- Updates Fonts and spacing to match designs
- Adds StatusBox Component to handle displaying status text

## Screenshot
image

### DIFF
--- a/src/components/DAOs/DAO/Proposals/ProposalCard.tsx
+++ b/src/components/DAOs/DAO/Proposals/ProposalCard.tsx
@@ -25,8 +25,8 @@ function ProposalCard({
   const proposerDisplayName = useDisplayName(proposal.proposer);
   return (
     <Link to={`proposals/${proposal.number}`}>
-      <div className={`rounded-lg bg-gray-600 p-4 border border-chocolate-400 hover:border-gold-300 hover:shadow-sm hover:shadow-gold-300 my-2`}>
-        <ContentBox>
+      <div className="my-1">
+        <ContentBox isLightBackground>
           <div className="flex flex-row">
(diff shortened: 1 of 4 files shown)


src/components/DAOs/DAO/Proposals/ProposalCard.tsx
```diff
@@ -25,21 +13,18 @@ function ProposalCard({
   const proposerDisplayName = useDisplayName(proposal.proposer);
   return (
     <Link to={`proposals/${proposal.number}`}>
-      <div className={`rounded-lg bg-gray-600 p-4 border border-chocolate-400 hover:border-gold-300 hover:shadow-sm hover:shadow-gold-300 my-2`}>
-        <ContentBox>
+      <div className="my-1">
+        <ContentBox isLightBackground>
```
adamgall, line 17: not sure if i like this... seems more like an "inconsistent background" issue than something that the ContentBox should be concerned with

adamgall: might want to revisit this after #123 is completed... let's just make a mental note lol

adamgall (approved):

src/components/DAOs/DAO/Proposals/ProposalCard.tsx
Da-Colon: Sure thing. The `InputBox` has the same styles. I did it this way as it's technically a `ContentBox` not a `InputBox`. but we can definitely revisit

PR #126, opened 2022-05-02
Content Banner Update

Da-Colon:
## Overview

Creates Content Banner and adds to Screens

## Changes

- updates Create DAO 'Token Details' with content banner
- updates Cast Vote Component with Content Banner

## Screenshots

image

image


src/components/ui/ContentBanner.tsx
```diff
@@ -0,0 +1,16 @@
+import Info from "./svg/Info";
+
+interface ContentBannerProps {
+  description: string;
+}
```
adamgall, line 5: Might be premature optimization right now, but would be cool to allow this to be a `children` prop that could be any `React.ReactNode`, instead of just a `string`. Will approve for now though.

adamgall (approved):

PR #133, opened 2022-05-02
updates TokenDetails input validations

Da-Colon:
## Overview
- adds `isWholeNumber` prop to string to prevent chars other than numbers
- adds props to allow for input customization if needed
- updates tokenSupply validation
- updates token allocation validation to happen individually on Change

## Screenshots
image


src/components/DAOs/New/TokenAllocationInput.tsx
```diff
@@ -25,20 +28,22 @@ const TokenAllocationInput = ({ index, tokenAllocation, updateTokenAllocation, r
 
   return (
     <>
-      <input
-        className="col-start-1 col-span-4 md:col-span-5 w-full border border-gray-200 bg-gray-400 rounded py-1 px-2 text-gray-50 focus:outline-none"
-        type="string"
+      <Input
+        containerClassName="col-start-1 col-helperspan-4 md:col-span-5 w-full"
```
adamgall, line 32: `col-helperspan-4`?

adamgall (changes requested):
Super small lil thang.

Other than this, I haven't given the PR a very thorough review, I'm sorry.

I figure, this seems like enough of a solid move in the right direction that any other tweaks we'll learn about from future usage. I want to just keep the merge cadence high right now.

adamgall (approved):

src/hooks/useAddress.ts
```diff
@@ -62,6 +63,33 @@ const useAddress = (addressInput: string | undefined) => {
   }, [provider, addressInput]);
 
   return [address, validAddress, loading] as const;
-}
+};
 
-export default useAddress;
\ No newline at end of file
+export const checkAddress = async (provider: any, addressInput?: string): Promise<boolean> => {
```
sethhrbek, line 68: could this be `isAddressValid`? `check` feels a little vague (also not sure if this app follows the `is`/`has` pattern for booleans).

src/hooks/useAddress.ts
```diff
@@ -62,6 +63,33 @@ const useAddress = (addressInput: string | undefined) => {
...
+
+  if (ethers.utils.isAddress(addressInput)) {
+    return true;
+  }
+  try {
+    const resolvedAddress = await provider.resolveName(addressInput);
+    if (!resolvedAddress) {
+      return false;
+    }
+    return true;
```
sethhrbek, line 89: Could this be `return !!resolvedAddress`

src/hooks/useAddress.ts
```diff
@@ -62,6 +63,33 @@ const useAddress = (addressInput: string | undefined) => {
...
-export default useAddress;
\ No newline at end of file
+export const checkAddress = async (provider: any, addressInput?: string): Promise<boolean> => {
+  if (addressInput === undefined) {
+    return false;
+  }
+
+  if (!provider || addressInput.trim() === "") {
+    return false;
+  }
```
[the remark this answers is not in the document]
sethhrbek: If you don't need to check provider, this case and previous case could be combined to just 

```
if (!addressInput) {
    return false;
  }
```

A an empty string with spaces `"   "` will return false with the `ethers.utils.isAddress` call.

src/components/DAOs/New/TokenDetails.tsx
```diff
@@ -77,7 +126,7 @@ const TokenDetails = ({
         <Input type="text" value={name} onChange={(e) => setName(e.target.value)} label="Token Name" helperText="What is your governance token called?" />
       </InputBox>
       <InputBox>
-        <Input type="text" value={symbol} onChange={(e) => setSymbol(e.target.value)} label="Token Symbol" helperText="Max: 5 chars" disabled={false} />
+        <Input type="text" value={symbol} onChange={onTokenChange} label="Token Symbol" helperText="Max: 5 characters" disabled={false} />
```
sethhrbek, line 129: Should `InputProps` have a `max` field added to set this to `5`? or is that handled somewhere else?

src/components/DAOs/New/TokenDetails.tsx
```diff
@@ -53,22 +54,70 @@ const TokenDetails = ({
...
+      tokenAllocations.map(async (tokenAllocation: { address: string; amount: number }, index: number) => {
+        if (errorMap.get(index)?.address === tokenAllocation.address) {
+          // do nothing if address at index hasn't changed since last render
+          return;
+        }
+        if (tokenAllocation.address.trim().length) {
+          // validates address as valid eth address or ENS domain as sets/removes error
+          const isValidAddress = await checkAddress(provider, tokenAllocation.address);
+          if (!isValidAddress) {
+            return setError(index, tokenAllocation, "Invalid address");
```
sethhrbek, line 114: would it be beneficial to type these error strings into their own constants in case the app needs to check them elsewhere in the app? I.e. `if (error === INVALID_ADDRESS) do something`

src/components/DAOs/New/TokenDetails.tsx
```diff
@@ -53,22 +54,70 @@ const TokenDetails = ({
...
+   * @param error
+   */
+  const removeError = (key: number) => {
+    const errors = new Map(errorMap);
+    errors.delete(key);
+    setErrorMap(errors);
+  };
+
+  useEffect(() => {
+    if (!tokenAllocations || !provider) return;
```
sethhrbek, line 103: In the other `useEffect` in this file, `tokenAllocations` is explicitly checked for `undefined` -- is there a reason the check here is different?

src/hooks/useAddress.ts
sethhrbek: I see a lot of "return true" / "return false" as opposed to things like `return !!myVar` in the app, in which case if that is the pattern can ignore this comment

PR #141, opened 2022-05-03
Proposal listener fix

tbwebb22:
This PR fixes the bug in which newly created proposals were not always added automatically to the proposals list.

This was done by checking if the proposal start block is in the future. If so, then the user vote power is set to undefined.

These changes also have the side effect of making the proposal states update automatically as new blocks are received, since currentBlockNumber is now a dependency of the useProposalsWithoutVotes hook.

### DIFF
--- a/src/daoData/useProposals.ts
+++ b/src/daoData/useProposals.ts
@@ -110,19 +110,28 @@ const getBlockTimestamp = (
   });
 };
 
+const getUserVotePower = (governorModule: GovernorModule, account: string, proposalStartBlockNumber: number, currentBlockNumber: number | undefined) => {
+  if(currentBlockNumber === undefined || proposalStartBlockNumber >= currentBlockNumber) {
+    return undefined;
+  }
+
+  return governorModule.getVotes(account, proposalStartBlockNumber);
+}
+
 // Get proposal data that isn't included in the proposal created event
 const getProposalData = (
   provider: providers.BaseProvider | undefined,
   governorModule: GovernorModule,
   proposal: ProposalDataWithoutVotes,
-  account: string
+  account: string,
+  currentBlockNumber: number | undefined,
 ) => {
   return Promise.all([
     governorModule.proposalVotes(proposal.id),
     governorModule.state(proposal.id),
     getBlockTimestamp(provider, proposal.startBlock.toNumber()),
     getBlockTimestamp(provider, proposal.endBlock.toNumber()),
-    governorModule.getVotes(account, proposal.startBlock),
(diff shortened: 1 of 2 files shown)


adamgall (commented):
Does this mean that all proposal data is requested every block?

tbwebb22:
Yes, thinking now maybe I should split useProposalWithoutVotes hook into two separate ones, one for static data that doesn't change, and one for dynamic data that could change with a new block.

adamgall:
The description of this PR is no longer correct right? The last paragraph

tbwebb22:
Yes, description updated.

src/daoData/useProposals.ts
```diff
@@ -228,16 +243,65 @@ const useUserVotes = (governorModule: GovernorModule | undefined) => {
...
+    });
+  }, [governorModule, account, currentBlockNumber, proposalsWithoutUserData]);
+
+  return userVotePowers;
+};
+
+const useProposalsWithoutUserData = (
+  governorModule: GovernorModule | undefined
+) => {
   const { account, provider } = useWeb3();
```
adamgall, line 297: `account` is not being used anymore (except for checking if it's `undefined`, and as a array param to the `useEffect` hook), so it should be removed.

src/daoData/useProposals.ts
```diff
@@ -110,54 +114,65 @@ const getBlockTimestamp = (
...
+    proposal.state = state;
+    proposal.startTime = startTime;
+    proposal.endTime = endTime;
+    proposal.startTimeString = getTimestampString(startTime);
+    proposal.endTimeString = getTimestampString(endTime);
+    proposal.stateString = getStateString(proposal.state);
+    proposal.eta = eta.toNumber();
+
+    return proposal;
+  });
```
adamgall, line 175: Please `.catch()` here and `console.error` the error

src/daoData/useProposals.ts
```diff
@@ -228,16 +243,65 @@ const useUserVotes = (governorModule: GovernorModule | undefined) => {
...
+          const newUserVotePower: UserVotePower = {
+            proposalId: proposal.id,
+            votePower: newUserVotePowerValues[index],
+          };
 
-const useProposalsWithoutVotes = (governorModule: GovernorModule | undefined) => {
+          return newUserVotePower;
+        })
+      );
+    });
```
adamgall, line 288: Please `.catch()` in here and `console.error`

adamgall (changes requested):
Minor feedback requests, thank you sir

adamgall (approved):

PR #143, opened 2022-05-03
Quick Fix: Pending Screen

Da-Colon:
## Overview
Notice a few other absolute/fixed elements  would hover above this screen. so add a z-index of 50


adamgall (approved):
Sure, but this component is going to get removed soon anyway

PR #144, opened 2022-05-03
Header Mobile update

Da-Colon:
## Overview

Updates Header menu not to flicker when opening on mobile

## Changes
- Also updates Disconnect text to be red
- removes transition effect when on small screens
- keeps address/menu right aligned when on small screens

## Screenshots
image

image

### DIFF
--- a/src/components/ui/Header/HeaderMenu.tsx
+++ b/src/components/ui/Header/HeaderMenu.tsx
@@ -48,11 +48,11 @@ const HeaderMenu = () => {
             <Transition
               show={open}
               enter="transition ease-out duration-100"
-              enterFrom="transform opacity-0 scale-95"
-              enterTo="transform opacity-100 scale-100"
-              leave="transition ease-in duration-75"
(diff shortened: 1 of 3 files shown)


Da-Colon:
@adamgall I was able to keep the address when I moved the menu to stay at the top. I can still have it remove address if you think this will just look better.

adamgall (commented):
Kind of nitpicky, and maybe this is just a Chrome "mobile device view" thing, but the menu seems to overflow off to the right. Notice in the screenshot how we don't see the rounded corners of the menu on the right side of the screen: they're off screen.

Screen Shot 2022-05-03 at 2 11 52 PM

Da-Colon:
@adamgall good eye. I'll get this fixed.

Da-Colon:
@adamgall after looking at it locally it appears normal in my browser. I have seen this bug, and can't reproduce consistently and opening a new tab in the browser and going back in dev tools fixes it.

adamgall (approved):

PR #147, opened 2022-05-03
Mobile: Welcome Screen

Da-Colon:
## Overview
updates buttons to fit on mobile

## Changes
- replaces with new Primary/Secondary Buttons

## Screenshots
image

### DIFF
--- a/src/components/Home.tsx
+++ b/src/components/Home.tsx
@@ -1,38 +1,20 @@
 import { Link } from "react-router-dom";
 import ContentBox from "./ui/ContentBox";
+import { PrimaryButton, SecondaryButton } from "./ui/forms/Button";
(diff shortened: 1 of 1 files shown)


adamgall (commented):
What about PrimaryButton and SecondaryButton were causing them to be so wide?

Da-Colon:
@adamgall wasn't the New Buttons. These were still using a 'ColoredButton' Component within the `Home.tsx` that wasn't set with a `width` so default to `auto` and that allowed the buttons to extend past the parent containers. 

Replaced with new buttons with className props to allow for these to stay within parent

adamgall:
Oh holy shit I read the git changes backwards, my fault.

adamgall (approved):

PR #163, opened 2022-05-04
Allow intermediate states in useModuleAddresses to be undefined

### LINKED ISSUE #137: State's `moduleAddresses` is empty array, not undefined, when DAO isn't loaded
It should be undefined though.

adamgall:
Closes #137 

Cleans up `useModuleAddresses` to allow for intermediate states to be undefined, which leads to a cleaner flow through the data building pipeline complete with early exits along the way.

Also includes a small fix to set `tokenDelegatee` to `undefined` in token information if early exit is happening


Da-Colon (approved):
Looks good. Don't see any problems locally

PR #165, opened 2022-05-04
Search only with account

### LINKED ISSUE #164: Don't allow user to Search for DAO if account not connected
- Disable the Search button
- show a toast indication they have to connect wallet, like in the New DAO workflow.

adamgall:
Closes #164 

Extracts the "persistent toast" out of Create a DAO components and into it's own component, which can easily be added anywhere else, like the Search component here.

Also disables Search button unless account is connected.

Screen Shot 2022-05-04 at 12 32 57 AM

Screen Shot 2022-05-04 at 12 35 13 AM


Da-Colon (approved):
This is great. plug and play anywhere in the app we need it

PR #166, opened 2022-05-04
No reload on disconnect

adamgall:
# Important

- This should be reviewed/merged _after_ #165 

This doesn't explicitly close any issues, but it does create a smoother app experience when disconnecting an account, because it no longer requires a window reload.

### DIFF
--- a/src/web3/index.tsx
+++ b/src/web3/index.tsx
@@ -1,8 +1,8 @@
 import React, { createContext, useContext } from 'react';
 
-import { Web3Custom, useProvider, defaultWeb3 } from './providers';
+import useProvider, { Web3Custom, defaultWeb3 } from './providers';
 
-const createWeb3Root = (context: React.Context<Web3Custom>) => {
+const createWeb3Root = (context: React.Context<readonly [Web3Custom, () => void, () => void]>) => {
   const Web3Root = ({ children }: { children: React.ReactNode }) => {
     const web3Provider = useProvider();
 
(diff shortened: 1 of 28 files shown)


src/web3/index.tsx
```diff
@@ -1,8 +1,8 @@
 import React, { createContext, useContext } from 'react';
 
-import { Web3Custom, useProvider, defaultWeb3 } from './providers';
+import useProvider, { Web3Custom, defaultWeb3 } from './providers';
 
-const createWeb3Root = (context: React.Context<Web3Custom>) => {
+const createWeb3Root = (context: React.Context<readonly [Web3Custom, () => void, () => void]>) => {
```
Da-Colon, line 5: It might be a good idea, for a future task to add Documentation to the useWeb3 hook and it's components. So anyone new can understanding how to use it. 

useWeb3

Da-Colon (changes requested):
I'm going to mark as changes request, IMO we need to get some documentation in for these components. But my request can be overruled, and a task can be created for this to be done at a later time.

adamgall:
I totally agree with you, this code is even still above my head a bit (even
though I wrote it).

I gotta figure out a better way to structure/document it.

adamgall:
@Da-Colon do you at least see what those changes entail as you look at the
rest of the code in this PR?

Da-Colon (approved):

Da-Colon:
@adamgall We can talk about this and add a card but I do see a simple way for a quick addon here.
image

`/web3/index.tsx`
```js
const useWeb3: () => readonly [Web3Custom, Connect, Disconnect] = () => {
  return useContext(web3Context);
};
```

Da-Colon (approved):
This works. looks good

PR #167, opened 2022-05-04
Only allow connected accounts to view DAO

### LINKED ISSUE #157: Don't allow /daos/:address to load if no account connected

adamgall:
Closes #157 

This UX manifests itself in two scenarios:

- When attempting to load a DAO URL directly, and a user doesn't have their wallet connected
- When viewing a DAO, and then the user manually disconnects their wallet


adamgall:
I could seriously use some extra eyes on this from @da-colon @tbwebb22 @sethhrbek

adamgall:
I could seriously use some extra eyes on this from @Da-Colon @tbwebb22 @sethhrbek. Let me know if you fellas are available for a screen share on Wednesday

adamgall:
After discussing with the fellas, came up with a solution that involved tracking when an account was being loaded, which is different than the two states of "connected account" and "no connected account".

So implementation-wise, force navigate back home if we're done "loading the account" and there is no account there.

adamgall:
This is ready for review :pray:

Da-Colon (approved):
👍

PR #168, opened 2022-05-04
Delegate Vote Address validation

Da-Colon:
## Overview
adds address and ens validation hook to Delegate Vote

## Note
There isn't currently a full way to test this in action as ENS Domains are currently only mainnet, and this screen isn't currently accessible on mainnet

## Screenshot
image


adamgall (approved):
LGTM

PR #169, opened 2022-05-04
fixed background

christopherdancy:
PR info:

Very simple tailwind change - added bg-fixed to keep bg static.

Before:
image

After:
image


Da-Colon (approved):
nice solution! looks good!

PR #174, opened 2022-05-04
Give an order for how DAO Name should display in breadcrumb

### LINKED ISSUE #138: When DAO is loading, show Address in breadcrumb
Currently, when a DAO is loading, the breadcrumb will show `...` in place of the DAO's name. Perhaps it would be better to display the DAO's address instead of `...`, and switch to the DAO's name once that is loaded.

adamgall:
First loaded name, if not that then verified DAO address, if not that then address from url params, if not that then "..."

Closes #138


Da-Colon (approved):
LGTM

PR #177, opened 2022-05-04
Add proposal listeners

tbwebb22:
This PR updates the useProposals hook to support:
 - Add voteCounts to ProposalData object
 - Only show the ProposalExecute component when the proposal is queued
 - Update voteCounts and votePercentages as new vote cast events are emitted
 - Update proposal state from pending to active when the proposal start block passes


adamgall (approved):
Lot going on here, but seems to work as expected!

SHIP IT

Da-Colon (approved):

PR #180, opened 2022-05-04
Delegate Vote Input Bug Fix

### LINKED ISSUE #175: Entering valid address on Delegate component flashes red "invalid address" error
This error manifests itself when either clicking the "Self" button, or pasting in a fully valid address into the input.

Since the address is in a fully valid state from the get-go, ideally there's no error displayed at all.

Seems like a better UX would be to delay showing an error until we're sure the address is invalid. Right now it seems like the error is being shown while the component/hook/whatever is figuring out if the address is valid or not.

Da-Colon:
## Overview
fixes #175 
error message i now handle by state/useEffect. It was being caused by `newDelegatee.trim() === ""` as part of the conditional.

## Screenshots
N/A


adamgall (approved):
works for me!

PR #184, opened 2022-05-05
Handle Network Change on DAO Route

### LINKED ISSUE #182: When network changes, recognize if loaded DAO is no longer valid and respond appropriately
Currently, if a DAO is loaded, and user changes their network, the UI kinda hangs because the loaded DAO is no longer valid (the address on the new network _likely_ is not a valid DAO on the new network).

The app should recognize this and probably kick the user back home, giving them an informational toast on what just happened.

Da-Colon:
closes #182 

## Overview
Adds a useEffect to fire when networkId is changed. reloading component and running address validating again. This would cause the user to be navigating back to "/" with toast error of "fractal doesn't exists"

### DIFF
--- a/src/components/DAOs/DAO/index.tsx
+++ b/src/components/DAOs/DAO/index.tsx
@@ -78,9 +78,9 @@ function Search() {
 function DAO() {
   const location = useLocation();
   const navigate = useNavigate();
-  const [{ account, accountLoading }] = useWeb3();
+  const [{ account, accountLoading, chainId }] = useWeb3();
   const [, setAddress] = useDAOData();
-
+  
   const [validatedAddress, setValidatedAddress] = useState((location.state as { validatedAddress: string } | null)?.validatedAddress);
   useEffect(() => {
     if (!location || !location.state) {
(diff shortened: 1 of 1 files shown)


src/components/DAOs/DAO/index.tsx
```diff
@@ -78,9 +78,9 @@ function Search() {
 function DAO() {
   const location = useLocation();
   const navigate = useNavigate();
-  const [{ account, accountLoading }] = useWeb3();
+  const [{ account, accountLoading, chainId }] = useWeb3();
   const [, setAddress] = useDAOData();
-
+
```
adamgall, line 83: eh

src/components/DAOs/DAO/index.tsx
```diff
@@ -103,6 +103,8 @@ function DAO() {
 
   // when this component unloads, setAddress back to undefined to clear app state
   useEffect(() => () => setAddress(undefined), [setAddress]);
+  // if network changes remove address validation.
+  useEffect(() => () => setValidatedAddress(undefined), [chainId])
```
adamgall, line 107: ok cool this seems simpler than i was expecting, nice.

question tho: should `setValidatedAddress(undefined)` happen on the return function of the `useEffect`? Why not have the line like this instead

```typescript
useEffect(() => setValidatedAddress(undefined), [chainId]);
```

Da-Colon: We want this to fire only on the `unmount` of the component. This means this will only be trigged when the the network is changed on unmount. Without that this will run once on mount. which would cause the `Search` component to mount and run the checks again (event when this page is navigated to via Search).

adamgall: oh duh that's right, thanks for explainming

adamgall (approved):

PR #186, opened 2022-05-05
Toast pending

christopherdancy:
PR Details
- Removed Pending Modal from the code base
- Replaced w/ transaction Toast handler 
- Utilizes Pending / Account checks to disable and reenable buttons.
- Before Deploy 
image
- After Deploy
image

- The same was done for the Create Proposal transaction.
- Also clears form state during  successful transaction


Da-Colon (approved):
LGTM

src/daoData/useCreateProposal.ts
```diff
@@ -48,17 +50,18 @@ const useCreateProposal = ({
...
+      pendingMessage: "Creating Proposal...",
       failedMessage: "Proposal Creation Failed",
       successMessage: "Proposal Created",
       successCallback: () => {
+        clearState();
         navigate(`/daos/${daoAddress}`,);
       },
       rpcErrorCallback: (error: any) => {
         console.error(error)
       },
```
adamgall, line 62: This is unnecessary because the error is already being caught and logged in `useTransaction`. We'd only want to utilize this `rpcErrorCallback` if there was actually some logical handling to do.

PR #202, opened 2022-05-05
Blockchain data context

adamgall:
Create a new top-level context for holding blockchain-specific data, such as current block number and current (blockchain) timestamp.

Originally these were in the `daoData` context, but they're not specific to DAOs so they've been moved.

As always, read through the commits in order to see the logical progression of changes.


Da-Colon (commented):
@adamgall (minor) do you think maybe these files `blockchainData` and `daoData` and moved to a `/data` or `/context` directory?

adamgall:
> @adamgall (minor) do you think maybe these files `blockchainData` and `daoData` and moved to a `src/data` or `src/context` directory?

Are you suggesting moving the entire `blockchainData` and `daoData` directories? (probably also include `web3` directory)? I like the idea of moving them all into a top level `src/contexts` directory, yes.

adamgall:
@Da-Colon what do you think of this?

Da-Colon (approved):
yessir. SHIP IT!

PR #230, opened 2022-05-06
BugFix: Token Allocation Validation

### LINKED ISSUE #212: Allocation Form Validation
## Problem
Token allocation form allows invalid addresses.

## Steps to recreate

In the token allocations:
1. Enter an invalid address as a token allocation. See that validation error shows.
2. Enter an amount for the invalid address. See that the validation error disappears (but should continue showing).
3. See that the "Next" button can be clicked (but should be disabled b/c there is an invalid address entered)
image

Da-Colon:
## Overview

Closes #212


adamgall (approved):
that was it, huh?

PR #240, opened 2022-05-06
Loading Proposal Twice

### LINKED ISSUE #227: Proposals Loading Bug
Problem:
When navigating away from proposals screen address validation is reseting causing not only the proposals to reload

 it may also be affecting the Browser's Back button and when on proposals it prevents use.
 
 The cause of the issue is on line 106 in `index.tsx`
 ```
   // if network changes remove address validation.
  useEffect(() => () => setValidatedAddress(undefined), [chainId]);
  ```
  
  This was added to reload the page when network has changed.

Da-Colon:
## Overview
Closes #227

### DIFF
--- a/src/components/DAOs/DAO/index.tsx
+++ b/src/components/DAOs/DAO/index.tsx
@@ -1,5 +1,5 @@
 import { useState, useEffect } from "react";
-import { Routes, Route, useParams, useLocation, useNavigate } from "react-router-dom";
+import { Routes, Route, useLocation, useNavigate } from "react-router-dom";
 import { toast } from "react-toastify";
 import { Helmet } from "react-helmet";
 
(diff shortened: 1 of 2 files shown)


adamgall (changes requested):
The back button behavior is still being weird. When on a proposal, need to click "back" two times to get back to DAO Summary.

The rest of the DAO Searching / Loading behavior seems to still work though, which is great!

Also, have confirmed that the "dao quickly reloading" bug when navigating between DAO summary and other dao pages is fixed 💪

Da-Colon:
@adamgall sorry for the delay, being casual today. This has been fixed. 
https://stackoverflow.com/questions/70237157/react-router-dom-go-back-twice-in-history-on-go-back-event

adamgall (approved):
Works great, thanks @Da-Colon

PR #248, opened 2022-05-06
Don't set timestamp with block.timestamp if block is null

### LINKED ISSUE #238: Bug: TypeError
When Connecting a wallet an error is shown in console
image

adamgall:
Closes #238

Fixes the console error.


Da-Colon (approved):
confirmed I no longer see console error

PR #251, opened 2022-05-07
File Structure refactor

Da-Colon:
## Overview

The goal with the this refactor was to make it easier navigating and finding files within the codebase

## Changes

At a high level, I've moved most entry points into the the main components into a `pages` directory. These files would contain the components that are passed into `Route` tags. 

For the components directory, I've flatten the file structure and created directories corresponding to the pages/directory. Components in the root of `/components` are 'shared' Components that don't quite fit for the `ui`  directory


src/types/proposal.ts
```diff
@@ -0,0 +1,8 @@
+import { BigNumber } from "ethers";
+
+export interface ProposalData {
+  targets: string[];
+  values: BigNumber[];
+  calldatas: string[];
+  description: string;
+};
```
Da-Colon, line 8: export interface/type/enums should be removed from components and pages and moved to `/types`. This not only makes easier to find but prevents having to worry about circular importing

src/pages/Daos/index.tsx
```diff
@@ -1,14 +1,15 @@
...
 
+// @todo refactor route files to own directory
 function DAOs() {
   return (
     <Routes>
       <Route index element={<DAOSearch />} />
-      <Route path="new" element={<New />} />
+      <Route path="new" element={<DaoCreate />} />
       <Route path=":address/*" element={<DAO />} />
     </Routes>
```
Da-Colon, line 14: We should think about the where these routing components live and find a way to make these more visible in the codebase

Da-Colon (commented):
I think there are 2 refactor tasks that come out of this refactor 
- see other comments

We could also take a PR to extract Prop Type definitions from within the React component to a "in-file" interface

```js
interface ComponentProps {
  someProps: string;
  }
  
  const Component = ({}: ComponentProps) => {
  // or 
  const Component: React.FC<ComponentProps> = ({}) => { 
  ```

adamgall (changes requested):
it don't build

Da-Colon:
@adamgall  yeah...ha. created draft as I about to go went to the movies. weird as locally it builds fine. gonna get situated and take a look at why.

Da-Colon:
:D didn't  push in the last commit. This might be a good place to stop for the scope of PR.

PR #254, opened 2022-05-09
removes extra margin

### LINKED ISSUE #225: Body container margins on DAO Summary are different than Proposal Detail
Notice how on DAO Summary, everything is left aligned on the same plane, from the Fractal logo, into Fractal name in breadcrumb, down to Proposals header, and finally to each proposal card.

But on the Proposal Detail page, the body is constrained by 8 or so pixels. That shouldn't be the case. The Proposal Details components should extend out to the same plane as the header margins.

Screen Shot 2022-05-06 at 12 58 14 AM
Screen Shot 2022-05-06 at 12 58 19 AM

Da-Colon:
closes #225 

Based off #251 

## Overview
Cleans up margin so DAO Summary and Proposal screen containers match

## Screenshot
image

### DIFF
--- a/src/components/Proposals/CastVote.tsx
+++ b/src/components/Proposals/CastVote.tsx
@@ -35,7 +35,7 @@ function CastVote({ proposal }: { proposal: ProposalData }) {
   const AbstainedSelected = () => (newVote === 2 || proposal.userVote === 2 ? <Check /> : null);
   return (
     <>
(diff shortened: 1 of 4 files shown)


adamgall:
@Da-Colon minor: to have github pick up the "closes" issue number, need to reference with a link to that issue (usually including the hash sign, then issue number)

adamgall (approved):
Looks Good Let's Go

PR #261, opened 2022-05-09
Disable action

christopherdancy:
PR details
- Disable Cast Vote button during transaction
* check transaction pending for cast vote button
* check pending for for, against, abstain button
 
image

- Disable queue button during transaction
* added pending state to check if transaction is being executed
image

- Disable execute button during transaction
* added pending state to check if transaction is being executed
image


adamgall (approved):
Looking good!

PR #262, opened 2022-05-10
Proposal state enum

### LINKED ISSUE #249: Turn the Proposal Data's `status` into an enum or an explicit list of strings
The proposal data's `status` property should be a Typescript enum, or just leave it as an integer.

Probably shouldn't even be exposing the `statusString` out of the `ProposalData` type. That makes it too risky that someone will use it to match on (like in the issue this issue was created from), causing bugs when data changes and no one knows to go update components downstream.

Create a hook that accepts a `status` integer and returns back a string for display.

_Originally posted by @adamgall in https://github.com/fractal-framework/fractal-interface/issues/213#issuecomment-1120095492_

adamgall:
Closes #249 

- Creates an enum type called `ProposalState` with the 8 states a proposal might be in.
- Updates the type of `state` in `ProposalData` from a `number` to `ProposalState`.
- Creates a hook for taking a `ProposalState` input and returning a string representation.
- Updates all UI components to no longer use `stateString`, but instead to logical comparisons using `state`, and visual display of state using the new hook.
- Removes `stateString` from `ProposalData`, instead exposing just `state`.

As always, the commits, when viewed in order, tell a story of the changes.

### DIFF
--- a/src/contexts/daoData/useProposals.ts
+++ b/src/contexts/daoData/useProposals.ts
@@ -30,7 +30,6 @@ type ProposalDataWithoutUserData = {
   calldatas: string[];
   description: string;
   state: ProposalState | undefined;
-  stateString: string | undefined;
   forVotesCount: BigNumber | undefined;
   againstVotesCount: BigNumber | undefined;
   abstainVotesCount: BigNumber | undefined;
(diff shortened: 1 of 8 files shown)


src/contexts/daoData/useProposals.ts
```diff
@@ -18,8 +29,7 @@ type ProposalDataWithoutUserData = {
   signatures: string[];
   calldatas: string[];
   description: string;
-  state: number | undefined;
-  stateString: string | undefined;
+  state: ProposalState | undefined;
```
christopherdancy, line 32: Hey @adamgall , is this ever undefined?

adamgall: It is on line 352 and 415 of this file, during the process in which proposals are being created through multiple on-chain data sources.

In the context of "our business logic", no, a Proposal's state should never be `undefined` (as we talked about last week, in addition to many other properties on the `ProposalData` type that are currently typed `something | undefined`).

However, "fixing" that is a much bigger task that I want @tbwebb22 to tackle.

christopherdancy: Understood - LFG

christopherdancy (approved):

PR #277, opened 2022-05-20
Show Dao Address in tooltip

### LINKED ISSUE #198: Show DAO address in UX
After creating a DAO, I had to copy the DAO address from the URL in order to find my way back to it.

We should display the DAO address somewhere prominent on the DAO page(s). 

It could potentially be truncated, but clicking it will copy the full address.

### LINKED ISSUE #268: Tooltip UI
We're looking for a reusable pattern for sticking tooltip popups wherever we want them

Da-Colon:
## Overview
closes #198 
closes #268 
Adds Tooltip library and set up a tool tip in DAO breadcrumbs that shows on DAO Dashboard, see screenshot.

## Screenshot

image


adamgall (changes requested):
Some general UX concerns with the current implementation:

- The breadcrumb bar used to show the DAO's address (in lieu of the DAO's name) when the DAOs name is loading (on first load).
  - But now, that address isn't shown, resulting in an "empty" breadcrumb bar for a second or two.
- When navigating to another page (like the Proposals page), the new hover component isn't being used. I'd think our intention would be to persist this new info component throughout any given sub-page.

Da-Colon:
@adamgall 

1:
- I thought it was a bug. my apologies I can revert the change that since it's intentional.
2: 
- I discussed this with @dannydecent and it was decided to have it only on the one page. This is easy to change as well if thats preferred.

adamgall:
@Da-Colon 

Cool, yes please revert the behavior from 1.

If you've already discussed 2 with @dannydecent and that behavior is intentional, then I have no problems with that!

adamgall:
@Da-Colon not to rush, but any updates on this PR?

Da-Colon:
@adamgall no updates, I was heads down in another branch. I'll switch to this one and knock this out real fast.

Da-Colon:
@adamgall updated

src/contexts/daoData/daoData.ts
```diff
@@ -48,9 +49,13 @@ const useDAODatas = () => {
   const governorModuleContract = useGovernorModuleContract(moduleAddresses);
   const tokenContract = useTokenContract(governorModuleContract);
   const tokenData = useTokenData(tokenContract);
-  const { currentBlockNumber } = useBlockchainData()
+  const { currentBlockNumber } = useBlockchainData();
   const proposals = useProposals(governorModuleContract, currentBlockNumber);
-
+  // variable to track when dao contracts have finished loading
+  const isDaoLoaded = useMemo(
```
adamgall, line 55: I'm not feeling strongly that this is good code to include here.

After your most recent commit, there's just one place that uses this property.

My main concern is that this `useMemo` will become stale/outdated as code changes around it, and we forget to copy the new properties in here.

Open to rebuttal, I just wonder if the intentions of this PR can be implemented without this new `isDaoLoaded` property.

adamgall (changes requested):

src/contexts/daoData/daoData.ts
Da-Colon: I don't have an good rebuttal other than I like having a boolean to check ensure that at least the contracts have loaded. 

But thinking about it you are right as this Context evolves this would change quite a bit and can be left out for now. ( should have no included the `TokenContract` and `GovernorModuleContract`. the original intent was to track that all the main DAO contract have loaded excluding the modules.

I don't have a problem with removing these (doing it now)

Da-Colon: @adamgall updated.

adamgall (approved):

PR #278, opened 2022-05-23
Not audited message

### LINKED ISSUE #275: Add non-audit disclaimer to webapp
There needs to be a message loud and clear on the webapp that the codebase (contracts and webapp) have not been audited.

adamgall:
Closes #275 

Show a "persistent" toast message to the user, indicating that the project is not audited.

On this toast, give the user a button indicating that they accept the risks involved.

When clicking "Accept", set a cookie in the browser's storage simply flagging that the user has accepted the risk and wants to dismiss the cookie.

When this cookie is set, the toast will go away and not be displayed again.

Screen Shot 2022-05-23 at 12 24 05 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,6 @@
   "packages": {
     "": {
       "name": "fractal-interface",
-      "version": "0.1.0",
       "dependencies": {
         "@headlessui/react": "^1.5.0",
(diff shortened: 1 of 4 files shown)


Da-Colon (approved):
looks good, works well. send it

PR #292, opened 2022-05-26
BugFix. Unmounted Component

Da-Colon:
## Overview
added a mounted variable to prevent unmounted setState


adamgall (approved):
Fix works for me, LFG

PR #295, opened 2022-05-27
resets state when url is manually updated on page

### LINKED ISSUE #293: Treasury not updating after loading new DAO
Load up a DAO with some values in the Treasury (ex: `0x0052654f8Eb94Ad56341dFCD1925186D4Ab0470B` on Rinkeby).

Then, "manually" load another dao by changing the DAO address in the url (ex: `0x8071912c42c73D8F32192443fE74f6Fc3D00A2c1`).

Notice how the Treasury page does not update; it still shows values from the first DAO.

This is because the app's state isn't properly resetting. Typically an easy fix in the patterns that we typically use. However, in `useTreasuryAssets`, this use of `Map` and lack of the state setter makes it weird. I don't like it.

Da-Colon:
## Overview

closes #293 

Adds logic to reset state when DAO address is changed in URL

### DIFF
--- a/src/contexts/daoData/treasury/useTreasuryAssets.ts
+++ b/src/contexts/daoData/treasury/useTreasuryAssets.ts
@@ -29,6 +29,7 @@ const useTreasuryAssets = (
    */
   useEffect(() => {
     if (!nativeDeposits || !nativeDeposits.length || !nativeWithdraws) {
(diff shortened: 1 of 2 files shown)


adamgall (approved):
Works for me, thank you sir

PR #298, opened 2022-05-28
Update types

adamgall:
bunch of plumbing shit happened today and this evening, which i'll describe here for the benefit of anyone who cares

today taylor deployed the full contract suite, which consisted of like 5 solidity repos, onto the major testnets (except ropsten). this was the culmination of work that chris did to separate modules out of the core-contracts repository.

in addition to simply moving code between repos, there was a new smart contract that was added, and some smaller changes to existing contracts.

because these contract ABIs have changed (slightly), we needed to identify and rectify those changes in our react codebase (the places where our frontend code was fetching data from the chain, or pushing data to the chain), else errors would occur because chain data would be malformed for what the react codebase was expecting.

unfortunately, this wasn't as easy as simply copying in a new `typechain-types` directory from "the" contracts repo, because now there are 5 contracts repos, each with their own `typechain-types` contract type definitions.

in addition, we created a new NPM module (`@fractal-framework/core-contracts`) which includes all of the core solidity contracts (needed for module developers to inherit the proper base contracts), _and_ the `typechain` definitions from those core contracts (our intention being that this would be good and helpful for anyone building client-side tooling that interacts with fractal DAOs, such as our webapp).

so, the plan was to remove the existing single big `typechain-types` directory from the frontend, then `npm install @fractal-framework/core-contracts` to get access to the core-contracts typechain definitions, then manually copy over all of the module `typechain-types` directories back into the frontend app. with all of these up-to-date typechain definitions (matching what was deployed on-chain earlier today), we could let our typescript compiler simply tell us what, if anything, was now broken in our frontend app.

for the most part this went well. it was good and easy to manually copy in all of the `typechain-types` directories from module repos into the frontend codebase.

i ran into a problem, though, with the npm published `core-contracts` typechain definitions. simply publishing the typechain defintions (which are pure typescript files, including both types, and some actual typescript implementation code) to NPM doesn't work. all NPM modules are expected to be javascript files, and optionally include typescript type files. by default a consuming typescript repo will not compile typescript in npm modules, those are expected to be compiled already.

so, i had to change our build + publish process in the `core-contracts` repo a bit, to actually compile those `typechain-types` files down into javascript+types files, and publish _those_, which is what our frontend could then consume (in addition to the actual solidity files, which is what module developers will need).

blah blah blah, i went through this whole process, got it working. at the end of the day, there was really only one small actual breaking change between the new contract ABIs and our frontend code, which was trivial to fix. so yay!

i also made sure to check each of the module repositories, which also have the `@fractal-framework/core-contracts` package as a dependency (to get access to the base solidity files needed for module inheritance), and there were no problems there thankfully.

we still need to actually deploy some new DAOs on these new contracts and make sure the frontend behaves as expected, but i'm feeling confident for our mainnet shenanigans next week.


Da-Colon (dismissed):
.

Da-Colon:
Getting an error message when trying to deploy a new DAO. I've tried locally and on netlify deployment site.

image
image

adamgall:
gonna need @christopherdancy or @tbwebb22 to chime in here... this is above my pay grade

adamgall:
As previously communicated with @tbwebb22:

My gut says that it’s related to those subtle changes in Governor, and the way that data is being formatted/passed into Metafactory.

I think the next step with testing is to copy over Governor and other supporting contracts over into Metafactory mocks, and hopefully the tests fail.

Fixing the tests should inform us of how to fix the frontend.

I was trying to debug it last night but it was hard. I finally realized that the best approach was to probably bring the contracts in the Metafactory “mocks” folder up to date with current (deployed) Governor and Treasury code. And then the tests in that codebase would hopefully fail, but it would be a lot easier to debug there, than across all these different repos. But by the time I realized that it was so late so I went to bed.

Da-Colon:
Can confirm DAO creation is successful, nicely done @tbwebb22

adamgall:
@Da-Colon @tbwebb22 this seems good now, thank you taylor! do you guys mind leaving approvals on here so we can merge?

Da-Colon (approved):
👍

PR #299, opened 2022-05-28
Remove the rpc error callback from useTransaction hook

adamgall:
This was getting in the way of toast errors being displayed to the user.

We should be striving to never even allow users to attempt to make transactions that will fail. We can do that with smart logic (looking at on-chain data and comparing it with who the user is and what they're about to do) to enable/disable buttons that trigger transaction.

Seems like the primary use case for this functionality was to inform users of business logic (ur not the owner, can't do this), but again, that shouldn't be caught at the time of making a transaction, it should be caught before then.

### DIFF
--- a/src/contexts/web3Data/transactions.ts
+++ b/src/contexts/web3Data/transactions.ts
@@ -16,7 +16,6 @@ interface ContractCallParams {
   failedCallback?: () => void;
   successCallback?: (txReceipt: ContractReceipt) => void;
   completedCallback?: () => void;
-  rpcErrorCallback?: (reason: string) => void;
 }
 
 const useTransaction = () => {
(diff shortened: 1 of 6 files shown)


Da-Colon (commented):
If you have a card created for "Form Error validation", this PR looks good to go. 

Also I'll take this opportunity to suggest the form library `Formik` and validation library `Yup`.

adamgall:
@Da-Colon i don't think we have a card for generic "form validation errors". 

For my own sake of making sure I understand your intention, why do you suggest form validation in the context of this PR?

Da-Colon:
>  We can do that with smart logic (looking at on-chain data and comparing it with who the user is and what they're about to do) to enable/disable buttons that trigger transaction.

@adamgall to take care of any more validation that needs to happen client side

Da-Colon (approved):

adamgall:
> > We can do that with smart logic (looking at on-chain data and comparing it with who the user is and what they're about to do) to enable/disable buttons that trigger transaction.
> 
> @adamgall noting to create one, to take care of any more validation that needs to happen client side

Ah ok. I guess I wasn't clear, but yeah we're already doing that in all the components where this PR touches.

for example: We only show the "queue transaction | execute transaction | create dao | cast vote | create proposal" button when it's actually possible to make the transaction. the transaction shouldn't fail for any reason, and if it does, a toast will pop up informing the user.

before this PR, if for some reason a transaction did fail (and i struggle to think of any situations where it would, given the use cases that this PR touches...) the "tx failed, sorry, check ur console" toast would get silenced... and the user would be left probably thinking that their transaction succeeded, when in fact it did not.

Issue #302, opened 2022-06-06
Error when attempting to propose Token Transfer with 18 digits


adamgall:
Broken with 18 digits

![Image](https://user-images.githubusercontent.com/706929/172230893-d890ed92-0090-41a4-a196-21302935a034.png)

Works with 15 digits

![Image](https://user-images.githubusercontent.com/706929/172230917-5d3afc7d-637e-4fbc-a646-7471c29016c3.png)

Is this an issue with the smart contracts or the frontend?

adamgall:
The solution is to wrap the big number in a string, then it works.

PR #323, opened 2022-06-08
 #302 error when adding BN without quotes 

christopherdancy:
[PR updates Changed example to include Bignumber w/ strings quotes.](https://github.com/decent-dao/fractal-interface/pull/322)


christopherdancy:
> Does passing a string still work for smaller numbers as well `"500"`.
> 
> is there something we can do in the behind the scenes, where a user doesn't have to add quotes?

Yes sir! it sure does work. If we add something behind the scenes we would need to update all the other strings as well. Also, given the fact that the address and number is passed as an array complicates things.

Da-Colon (approved):
@christopherdancy awesome. okay cool. I'm sure as we get plugins added this process will be refined.

adamgall (approved):

PR #333, opened 2022-06-10
Refactor/web3 provider

Da-Colon:
## Overview

A quick pass at some optimizations to the Web3 Wallet Provider we currently use.

- Instead of using `useState()` to manage a complex objects, it's recommended to use `useReducer()`. This manages state less reactively and allows for faster loading and updating of state.
- All of the places in the code that uses `useWeb3` has been updated to `useWeb3Provider`
- updated any typing that changed due to this update
- `get` fetches to the blockchain now use `provider`. leaving the Signer to be specifically use to `post` data.

## Notes 

The useWeb3Provider is now structure in a way that it can be deconstructed without worrying about undefined values

```js
const {
  state: { acount, signer, name, network, chainId, provider }
  dispatch,
  connect,
  disconnect,
} = useWeb3Provider();
```


Da-Colon:
Draft until more testing is performed. I'll @ when it's ready tomorrow.

src/contexts/web3Data/Web3Provider.tsx
```diff
@@ -0,0 +1,154 @@
...
+          payload: localProvider,
+        });
+        return;
+      })();
+    }
+    dispatch({
+      type: Web3ProviderActions.SET_FALLBACK_PROVIDER,
+      payload: getFallbackProvider(),
+    });
+  }, []);
```
Da-Colon, line 121: alot of the original useEffect logic in `providers` was moved into this method.

src/hooks/useDeployDAO.ts
```diff
@@ -35,7 +35,12 @@ const useDeployDAO = ({
   setPending: React.Dispatch<React.SetStateAction<boolean>>;
   clearState: () => void;
 }) => {
-  const [{ signerOrProvider, chainId }] = useWeb3();
+  const {
+    state: {
+      wallet: { signer },
+      connection: { chainId },
+    },
+  } = useWeb3Provider();
```
Da-Colon, line 43: Signer is used in places where a signer is necessary.

src/contexts/daoData/useAccessControlContract.ts
```diff
@@ -1,21 +1,21 @@
...
 import { AccessControlDAO, AccessControlDAO__factory } from '@fractal-framework/core-contracts';
-import { useWeb3 } from '../web3Data';
+import { useWeb3Provider } from '../web3Data/hooks/useWeb3Provider';
 
 const useAccessControlContract = (accessControlAddress: string | undefined) => {
   const [accessControlContract, setAccessControlContract] = useState<AccessControlDAO>();
-  const [{ signerOrProvider }] = useWeb3();
+  const {
+    state: { provider },
+  } = useWeb3Provider();
```
Da-Colon, line 10: Since Signer isn't needed for `get`ting data, just the provider is needed

adamgall (commented):
@Da-Colon honestly i'm not even sure how to begin reviewing this lmao

Da-Colon:
@adamgall Well overall not much has changed. 

I did just realized I need to update the `DISCONNECT` flow a bit. I went straight to FallbackProvider skiping the Local Provider on DIsconnect. I'll fix this in a moment.

The biggest changes other than useState -> useReducer is the structure of the return of the useWeb3Provider(). if you look through a few of the files that have been updated using it and your okay with that implementation I'd say everything else has just be restructured but the crust is the same.

src/contexts/web3Data/hooks/useLIsteners.ts
```diff
@@ -0,0 +1,114 @@
...
+      if (!supportedChains().includes(parseInt(_modalProvider.chainId))) {
+        toast(`Switch to a supported chain: ${supportedChains().join(', ')}`, {
+          toastId: 'switchChain',
+        });
+        // remove cached provider
+        web3Modal.clearCachedProvider();
+        // switch to a default provider
+        connectDefaultProvider();
+      } else {
+        setModalProvider(_modalProvider);
```
Da-Colon, line 40: Noticed the reason why the web3Modal listeners were working was that we were originally placing the event listeners on the ethers Provider method. this caused issues when not connected to correct network.

src/hooks/useDeployDAO.ts
Da-Colon: This object has changed.

```js
  const { state: { signer, chainId }  } = useWeb3Provider();    
```

src/contexts/daoData/useAccessControlContract.ts
Da-Colon: This object has changed.
```js
  const { state: { provider }  } = useWeb3Provider(); 
```

Da-Colon:
Went ahead and unnested the WebProvider state object. 

```ts
export interface InitialState {
  connectionType: string;
  network: string;
  chainId: number;
  provider: Providers | null;
  account: string | null;
  signer: ethers.Signer | null;
  isProviderLoading: boolean;
}
```

Da-Colon:
@adamgall i was thinking about it but maybe a live code review to get this thing reviewed and merged?

adamgall:
> @adamgall i was thinking about it but maybe a live code review to get this thing reviewed and merged?

This sounds like a great idea

adamgall (approved):
Approving with the caveat that I haven't pulled this down locally. before merging, please make sure it builds correctly :)

PR #334, opened 2022-06-10
Pause for 2 seconds before informing app that Transaction is completed

### LINKED ISSUE #306: BUG: UI lets me "Cast vote" twice
Narrowed this down to the slight delay between when a transaction is mined, and the completion of that transaction is relayed back to the app, and the block that the transaction made it into actually being emitted and played onto to the app's state.

to fix: add small delay in `useTransaction` before `setPending(false)`.

This fix will affect all transactions made from the app.

adamgall:
Closes #306 

Adds a 2 second delay after transaction is mined, before reporting back to the business logic layer that that the transaction is no longer pending.


Da-Colon (approved):
👍

PR #343, opened 2022-06-16
Create New Root DAO from within existing DAO

adamgall:
Ok, this is the first PR in what'll be a series of PRs, the sum of which will be a (hopefully) nice lil framework for this Plugin system.

This first PR here is very small and simple: replicate the current "Create a DAO" workflow inside of the context of an existing loaded DAO.

- Makes sure the "Create a DAO" workflow is entirely self-contained in one component.
- Creates new `{daoAddress}/transactions/new` route, which simply has the current existing "Create a DAO" workflow.
- Add a button to DAO Summary page to get to this "New Transaction" page.


Da-Colon (approved):
LGTM 👍

PR #346, opened 2022-06-18
updates listener event to unsubscribe properly when connected to wall…

Da-Colon:
## Overview
updates modal event listener to unsubscribe properly when connected using Wallet Connect.


adamgall (approved):
I trust that these changes are good and necessary. This web3 plumbing is thicc.

PR #352, opened 2022-06-23
Find modules

adamgall:
Build in support to "look for" the Treasury, Governor, and Timelock modules dynamically at runtime, instead of hardcoding their "index" in the `moduleAddresses` array.

### DIFF
--- a/src/components/Dao/Details.tsx
+++ b/src/components/Dao/Details.tsx
@@ -1,3 +1,4 @@
+import { useState, useEffect } from 'react';
 import { useDAOData } from '../../contexts/daoData';
 import ContentBox from '../ui/ContentBox';
 import ContentBoxTitle from '../ui/ContentBoxTitle';
@@ -8,13 +9,18 @@ import H1 from '../ui/H1';
 
 interface AddressDisplayProps {
   address?: string;
-  label: string;
+  label: string | Promise<string>;
 }
 
 function AddressDisplay({ address, label }: AddressDisplayProps) {
+  const [lbl, setLbl] = useState('');
+  useEffect(() => {
+    Promise.resolve(label).then(setLbl);
+  }, [label]);
+
   return (
     <div>
-      <div className="text-sm font-medium text-gray-50 pb-1">{label}</div>
+      <div className="text-sm font-medium text-gray-50 pb-1">{lbl}</div>
       <div className="flex items-center">
         <EtherscanLink address={address}>
           <div className="text-gold-500 hover:text-gold-300 font-semibold tracking-wider font-mono break-all">
@@ -32,12 +38,14 @@ function Details() {
     {
       name,
       accessControlAddress,
-      moduleAddresses,
       daoAddress,
       modules: {
         governor: {
+          governorModuleContract,
+          timelockModuleContract,
           votingToken: { votingTokenData },
         },
+        treasury: { treasuryModuleContract },
       },
     },
   ] = useDAOData();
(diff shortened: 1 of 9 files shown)


src/components/Dao/Details.tsx
```diff
@@ -8,13 +9,18 @@ import H1 from '../ui/H1';
...
 function AddressDisplay({ address, label }: AddressDisplayProps) {
+  const [lbl, setLbl] = useState('');
+  useEffect(() => {
+    Promise.resolve(label).then(setLbl);
+  }, [label]);
+
   return (
     <div>
-      <div className="text-sm font-medium text-gray-50 pb-1">{label}</div>
+      <div className="text-sm font-medium text-gray-50 pb-1">{lbl}</div>
```
adamgall, line 23: This is kind of hacky for now... Basically I wanted to capture the `name` of the contract (from `ModuleBase`) and display it in the UI, but there are some other items on this Details page that we still want to just manually hardcode the label as a string. So, made this `label` input take either a `string` or a `Promise<string>` and simply force-resolve it (it's totally fine to resolve a non-promise, it just resolves immediately with the same thing that it is).

The actual better solution would be to capture all of this module data (such as the `name` which is what i want here) up in the global dao data, but I simply didn't want to do that right now because of how much this is going to change in the near future anyway.

src/contexts/daoData/daoData.ts
```diff
@@ -85,14 +87,14 @@ const useDAODatas = () => {
     daoAddress,
     name,
     accessControlAddress,
-    moduleAddresses,
```
adamgall, line 88: Removed the `moduleAddresses` array (of just contract address strings) from global DAO state yay

src/contexts/daoData/daoData.ts
```diff
@@ -85,14 +87,14 @@ const useDAODatas = () => {
...
     modules: {
       treasury: {
         treasuryModuleContract,
         treasuryAssets,
       },
       governor: {
         governorModuleContract,
+        timelockModuleContract,
```
adamgall, line 97: Added the timelock contract into global DAO state.

src/contexts/daoData/treasury/useTreasuryModuleContract.ts
```diff
@@ -2,27 +2,42 @@ import { useEffect, useState } from 'react';
...
 
-const useTreasuryModuleContract = (moduleAddresses?: string[]) => {
+const useTreasuryModuleContract = (moduleAddresses: string[] | undefined) => {
+  const [treasuryModule, setTreasuryModule] = useState<TreasuryModule>();
   const {
-    state: { provider },
+    state: { signerOrProvider },
   } = useWeb3Provider();
-  const [treasuryModule, setTreasuryModule] = useState<TreasuryModule>();
+
```
adamgall, line 17: Gonna leave some comments throughout this file, which are effectively the same thing in the `useGovernorModuleContract` and `useTimelockModuleContract` hooks.

src/contexts/daoData/treasury/useTreasuryModuleContract.ts
```diff
@@ -2,27 +2,42 @@ import { useEffect, useState } from 'react';
...
+  const [contracts] = use165Contracts(moduleAddresses);
```
adamgall, line 18: Ok so first, pass in _all_ of the `moduleAddresses` (it's an array of strings) into `use165Contracts` hook, which just creates "165 contracts" out of each of them (adds typings to call `supportsInterface` basically).

src/contexts/daoData/treasury/useTreasuryModuleContract.ts
```diff
@@ -2,27 +2,42 @@ import { useEffect, useState } from 'react';
...
+  const [interfaces] = useState([
+    IModuleBase__factory.createInterface(),
+    ITreasuryModule__factory.createInterface(),
+  ]);
```
adamgall, line 22: Next create an array of the different interfaces that the module we're looking for needs to support. In all of these module cases, they need to be `IModuleBase` and whatever interface the actual module is.

src/contexts/daoData/treasury/useTreasuryModuleContract.ts
```diff
@@ -2,27 +2,42 @@ import { useEffect, useState } from 'react';
...
+
+  const [potentialTreasuries] = useSupportsInterfaces(contracts, interfaces);
```
adamgall, line 24: Now this one is kind of weird maybe, because of damn hooks rules. Can't really put hooks in a loop or a mapping or anything like that.

What we want to do is check each of the module addresses, to see if they support the interfaces that we've defined above. But again due to hooks rules, can't do that in a loop. So, I decided to just pass the whole damn array of contracts into `useSupportsInterfaces`, along with the array of interfaces that we're looking for support of.

It returns back an array of objects shaped like `{ address: string, match: boolean }`, informing us which of the contract addresses match (support) the interfaces that we passed in.

src/contexts/daoData/treasury/useTreasuryModuleContract.ts
```diff
@@ -2,27 +2,42 @@ import { useEffect, useState } from 'react';
...
-      !moduleAddresses.length ||
-      !moduleAddresses[0] ||
-      !provider
-    ) {
+    if (potentialTreasuries === undefined || !signerOrProvider) {
+      setTreasuryModule(undefined);
+      return;
+    }
+
+    const treasury = potentialTreasuries.find(t => t.match === true);
```
adamgall, line 32: At the current phase of this project, we are assuming that each module we care about is installed once and only once in a DAO. So here, we're just looking for the first address where `match` is true.

In later phases of this codebase, we will not make this assumption, because maybe no Treasury module exists. Or maybe multiple Treasury modules exists. But for now, just do a `.find` (which returns the first match).

src/hooks/use165Contracts.ts
```diff
@@ -3,28 +3,28 @@ import { useState, useEffect } from 'react';
 import { IERC165, IERC165__factory } from '@fractal-framework/core-contracts';
 import { useWeb3Provider } from '../contexts/web3Data/hooks/useWeb3Provider';
 
-const use165Contract = (address: string | undefined) => {
+const use165Contracts = (addresses: string[] | undefined) => {
```
adamgall, line 6: Simple changes in this hook -- just update it to support an array of address strings coming in, and spit out an array of `IERC156` contracts.

src/hooks/useSupportsInterfaces.ts
```diff
@@ -3,26 +3,37 @@ import { utils } from 'ethers';
 
 import { ERC165 } from '@fractal-framework/core-contracts';
 import useInterfaceSelectors from './useInterfaceSelectors';
+import { ContractMatch } from '../types/contractMatch';
 
-const useSupportsInterfaces = (contract: ERC165 | undefined, interfaces: utils.Interface[]) => {
-  const [supportsInterfaces, setSupportsInterfaces] = useState<boolean>();
+const useSupportsInterfaces = (contracts: ERC165[] | undefined, interfaces: utils.Interface[]) => {
```
adamgall, line 8: As mentioned above, this hook now takes an array of contracts instead of just one contract.

src/hooks/useSupportsInterfaces.ts
```diff
@@ -3,26 +3,37 @@ import { utils } from 'ethers';
...
+          interfaceSelectors.map(selector =>
+            contract.supportsInterface(selector).catch(() => false)
+          )
+        ).then(contractSupports => ({
+          address: contract.address,
+          match: contractSupports.reduce((p, c) => p && c),
+        }))
+      )
+    )
+      .then(allContractsSupports => setSupportsInterfaces(allContractsSupports))
```
adamgall, line 34: And I spent way too fucking long on this bullshit right here.

In "iterative terms": for each contract, call `supportsInterface` for each interface. If any of them are not supported, then the whole contract is a no-match. If all of them are supported, then the contract is a match. Do that for each contract, building up that `ContractMatch` object, put them all in an array, and return that array of `ContractMatch`es.

Da-Colon (approved):
This changes make sense, tested deployed and LGTM 👍 💯

PR #353, opened 2022-06-23
Reuse account display name

adamgall:
The issue is that each time the menu is opened, if the connected account has an ENS name, there's a quick flash as the address turns into the ens name, _every time the menu is opened_.

- @Da-Colon ur probably gonna yell at me, but first thing in this PR is that I changed some of the initial values of Web3Provider object from `null` to `undefined` (for better compatibility with how the rest of the app is set up)
- Next I added the "account display name" to global `useBlockchainData`, so that we don't have to fetch it each time we want to show it (for example, like _each time_ the menu is opened).
- Finally I used that thang in the Menu

### DIFF
--- a/src/components/ui/CopyToClipboard.tsx
+++ b/src/components/ui/CopyToClipboard.tsx
@@ -1,7 +1,7 @@
 import Clipboard from './svg/Clipboard';
 import { toast } from 'react-toastify';
 
-function CopyToClipboard({ textToCopy }: { textToCopy: string | undefined }) {
+function CopyToClipboard({ textToCopy }: { textToCopy: string | undefined | null }) {
   const copyTextToClipboard = () => {
     if (!textToCopy) return;
     toast('Copied to clipboard', {
--- a/src/components/ui/EtherscanLink.tsx
+++ b/src/components/ui/EtherscanLink.tsx
@@ -5,7 +5,7 @@ function EtherscanLink({
   address,
   children,
 }: {
-  address: string | undefined;
+  address: string | undefined | null;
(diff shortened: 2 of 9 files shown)


Da-Colon:
@adamgall you are right, here it comes 😠 lol. 

So first my reasoning for `null` versus `undefined`. 

- undefined for me is a value that after initialization, its unknown whether be set or not
- null being used for situation ways where the state of something is more intentional (set or not set).

> (for better compatibility with how the rest of the app is set up)
This package is meant to be more universal.

adamgall:
> @adamgall you are right, here it comes 😠 lol. 
> 
> 
> 
> So first my reasoning for `null` versus `undefined`. 
> 
> 
> 
> - undefined for me is a value that after initialization, its unknown whether be set or not
> 
> - null being used for situation ways where the state of something is more intentional (set or not set).
> 
> 
> 
> So for in this case, account is either connected or not connect. we will always be sure that is will be one or the other.
> 
> 
> 
> > (for better compatibility with how the rest of the app is set up)
> 
> 
> 
> This package is meant to be more universal. 

I agree completely with the points you're making, and argue that this PR puts things back into that pattern.

adamgall:
Eh, disregard my last comment. I need to think about this more.

For the sake of exercise, if you have time, pull this branch down and make those things `null` again, and see how much that complicates what I was trying to do with this PR.

adamgall:
@Da-Colon i've merged #354 into this branch. so it should pass your review now.

Da-Colon (approved):
@adamgall LGTM thanks for hearing me out.

PR #354, opened 2022-06-23
redefines provider states as null

Da-Colon:
@adamgall 

I think there are a few instances that does it fact take `undefined` from elsewhere in the app. But I think it's safe to update those types to include null as well so `string | undefined | null`.

### DIFF
--- a/src/components/ui/CopyToClipboard.tsx
+++ b/src/components/ui/CopyToClipboard.tsx
@@ -1,7 +1,7 @@
 import Clipboard from './svg/Clipboard';
 import { toast } from 'react-toastify';
 
-function CopyToClipboard({ textToCopy }: { textToCopy: string | undefined }) {
+function CopyToClipboard({ textToCopy }: { textToCopy: string | undefined | null }) {
   const copyTextToClipboard = () => {
     if (!textToCopy) return;
     toast('Copied to clipboard', {
--- a/src/components/ui/EtherscanLink.tsx
+++ b/src/components/ui/EtherscanLink.tsx
@@ -5,7 +5,7 @@ function EtherscanLink({
   address,
   children,
 }: {
-  address: string | undefined;
+  address: string | undefined | null;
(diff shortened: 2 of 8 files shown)


src/components/ui/CopyToClipboard.tsx
```diff
@@ -1,7 +1,7 @@
 import Clipboard from './svg/Clipboard';
 import { toast } from 'react-toastify';
 
-function CopyToClipboard({ textToCopy }: { textToCopy: string | undefined }) {
+function CopyToClipboard({ textToCopy }: { textToCopy: string | undefined | null }) {
   const copyTextToClipboard = () => {
     if (!textToCopy) return;
     toast('Copied to clipboard', {
```
Da-Colon, line 7: Here.

src/components/ui/EtherscanLink.tsx
```diff
@@ -5,7 +5,7 @@ function EtherscanLink({
   address,
   children,
 }: {
-  address: string | undefined;
```
Da-Colon, line 8: and here

src/components/ui/EtherscanLink.tsx
```diff
@@ -5,7 +5,7 @@ function EtherscanLink({
...
+  address: string | undefined | null;
```
Da-Colon, line 8: And here

adamgall:
This is the same route I went down, and blah I did not like having those types include both `undefined` and `null`.

Hmmmmmm

Da-Colon:
Yeah I see the conundrum but I don't know if it can be helped. because `daoAddress` for example follows my reasoning as well but it is passed to the component as undefined. I think to stay within the same patterns this component would need to accept both types for these specific components.

IMO it's not completely a bad thing considering the versatility of these components.

adamgall:
Ok you win this one

adamgall (commented):
Ok

Issue #300, opened 2022-05-28
Handle any outstanding TODOs that will affect mainnet/prod


adamgall:
Not sure if all of these need to be dealt with, but more than 0 of them do.

![Screen Shot 2022-05-28 at 2 14 57 PM](https://user-images.githubusercontent.com/706929/170838010-fd850264-b16d-4787-9864-f0abd3f1f182.png)

adamgall:
@christopherdancy any progress on this?

PR #364, opened 2022-06-29
#300 completed todo

christopherdancy:
Removed all todos and created issues for todos which should be handled at a separate time.

Todo handled in this PR
- Removed unused vars by exporting types
- Skipping native token and settlement layer

### DIFF
--- a/src/contexts/daoData/useProposals.ts
+++ b/src/contexts/daoData/useProposals.ts
@@ -227,7 +227,11 @@ const useUserVotes = (governorModule: GovernorModule | undefined) => {
 
     const filter = governorModule.filters.VoteCast(account);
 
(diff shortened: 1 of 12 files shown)


src/contexts/daoData/useProposals.ts
```diff
@@ -380,17 +375,16 @@ const useProposalsWithoutUserData = (
 
     const filter = governorModule.filters.ProposalCreated();
 
-    const listenerCallback = (
+    const listenerCallback: ProposalCreatedListener = (
```
adamgall, line 378: This "proposal created" listener isn't working.

Testing on the `develop` branch, it isn't working there either...

But regardless, rather than change the function signature here, just add an underscore to the `values` param to tell ESLint to ignore it.

adamgall (changes requested):

adamgall (approved):
Thank you!

PR #366, opened 2022-07-01
Erc721

christopherdancy:
This PR handles the following
1. Listen for ERC721 deposits / withdraws
2. Update treasury state based on ERC721 deposits / withdraws
3. Create a second UI box which displays treasury Assets / ERC721
4. Create an href for the tokenURI

Screen shots showing the different states a treasury can be in 
1. Token Assets 
image
2. NFT Assets
image
3. Both 
image

Steps to recreate
1. Create A DAO
2. Etherscan Treasury - Verify Proxy
3. Approve Treasury - via NFT / Token
4. Call Deposit(?TokenType) - watch magic happen


src/pages/Treasury/index.tsx
```diff
@@ -37,7 +37,9 @@ function Treasury() {
             </div>
             <div className="text-gray-50 text-xs font-medium">Amount</div>
           </div>
-          {!treasuryAssets.length && (
+          {!treasuryAssets.filter(
+            asset => asset.type.includes('erc20') || asset.type.includes('native')
+          ).length && (
```
adamgall, line 42: I think the UI layer here is not really a great place to "split" assets up between fungible/non-fungible. Probably better to have these assets split up back in the data layer, which is what I was trying to get at in my previous couple of comments.

adamgall (changes requested):
Main theme of my review here is to separate fungible and non-fungible logic back in the data layer, perhaps by renaming the `useTreasuryAssets` hook to `useTreasuryAssetsFungible` and creating a new `useTreasuryAssetsNonFungible` hook. Then just create a new parameter in the global treasury state to hold those "non fungible" datas.

christopherdancy:
@adamgall This is not completed. I will address before final product.

christopherdancy:
Leaving a comment here - @adamgall @Da-Colon - There seems to be an issue with assigning roles to actions. This is confirmed on etherscan. Will move this to a seperate issue.

src/contexts/daoData/daoData.ts
```diff
@@ -27,7 +28,8 @@ export interface DAOData {
   modules: {
     treasury: {
       treasuryModuleContract: TreasuryModule | undefined;
-      treasuryAssets: TreasuryAsset[];
+      treasuryAssetsFungible: TreasuryAssetFungible[];
+      treasuryAssetsNonFungible: TreasuryAssetNonFungible[];
```
adamgall, line 32: yes yes this is nice

src/pages/Treasury/index.tsx
```diff
@@ -104,30 +100,28 @@ function Treasury() {
...
+                <div className="text-gray-25 font-medium">{asset.name}</div>
+              </div>
+              <div className="text-gray-25 font-mono font-semibold tracking-wider">
+                <a
+                  href={`https://${asset.tokenURI}`}
+                  target="_blank"
+                  rel="noreferrer"
+                >
+                  {asset.tokenId.toString()}
+                </a>
```
adamgall, line 118: I question whether or not we should even have this part in here for now.

This snippet is nowhere near production ready. for example:

- why prefixing the `tokenURI` with a static `https://`? Don't `tokenURI`s typically include the http schema already?
- this doesn't work when the `tokenURI` is an IPFS hash
- this seems like it'll only link to the token's metadata object, not the art/image itself. Maybe that's intended.

We'd probably want to use a specific library for this that covers all the edge cases for "displaying NFTs", something like [useNFT](https://github.com/spectrexyz/use-nft).

In the meantime, might be safest to avoid all the edge cases entirely by not even trying to link to anything.

Or maybe!, for now, we just link to the NFT on Etherscan:

`https://[network.]etherscan.io/[tokenAddress]/[tokenId]`

e.g.: https://etherscan.io/nft/0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d/1512

(there's already a hook to help generate etherscan links depending on the connected network. perhaps build upon that to make this NFT link easy to generate)

adamgall (changes requested):
So close! Just one requested change at the place where NFT links are generated

src/pages/Treasury/index.tsx
christopherdancy: that is a dope idea

adamgall (dismissed):
Approving, but asking you to create an issue to clean up some tech debt that's being merged in this PR.

adamgall (changes requested):
These latest changes don't really solve the issue that we talked about yesterday, which is that callers can pass data into the component and receive an invalid link.

The base `EtherscanLink` component is still callable, and can still be called in a way that produces an invalid link.

adamgall (approved):
Looking good!

PR #374, opened 2022-07-18
Refactor/govenor module

Da-Colon:
## Overview

Part 2 of ?

This PR covers the extraction of the Governor Module hooks and data into its own provider. 

I see the `Module Provider` not wrapping the entire app but rather wrapping the route itself, being passed the moduleAddreses from the Fractal Data Hook. if the correct address is included then this module will be displayed to the user.

### DIFF
--- a/src/providers/govenor/GovenorModuleProvider.tsx
+++ b/src/providers/govenor/GovenorModuleProvider.tsx
@@ -1,0 +1,42 @@
+import { ReactNode, useMemo } from 'react';
+import useBlockchainDatas from '../../contexts/blockchainData/blockchainData';
+import { GovernorContext } from './hooks/useGovenorModule';
+import useGovernorModuleContract from './hooks/useGovernorModuleContract';
+import useProposals from './hooks/useProposals';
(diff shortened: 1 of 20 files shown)


src/providers/govenor/hooks/useProposals.ts
```diff
@@ -0,0 +1,684 @@
+import { useState, useEffect } from 'react';
```
Da-Colon, line 1: This hook needs to be have types extracted and looked at for any optimizations available

src/providers/govenor/GovenorModuleProvider.tsx
```diff
@@ -0,0 +1,42 @@
...
+      },
+    }),
+    [
+      governorModuleContract,
+      timelockModuleContract,
+      proposals,
+      votingTokenContract,
+      votingTokenData,
+    ]
+  );
```
Da-Colon, line 40: The value here hasn't changed from what is currently returned from the daoData provider

Da-Colon (commented):
This PR still has some changes to do as I look through these hooks more closely

Da-Colon:
This will be merged into the main refactor branch locally

PR #384, opened 2022-07-26
Change updateDelay function on timelock contract to be callable by DA…

tbwebb22:
This PR fixes a bug that was uncovered by the security audit. The DAO_ROLE should be able to call the updateDelay function on the Timelock contract. This allows a DAO to change the timelock delay by passing a proposal.

Previously, this function was configured to be callable by the GOVERNOR_ROLE, which is assigned to the Governor module. However, the Governor module has no function enabling it to call updateDelay on the Timelock contract.


adamgall:
@christopherdancy can you please review this before I do? Thank you

christopherdancy (approved):
This is correct. The governor contract did not call to the updateDelay method so it would be impossible to update the delay. Now the DAO is authorised to call this function.

PR #385, opened 2022-07-27
Parent Token Mirroring

christopherdancy:
**PR Explained**
This PR allows a parent daos token allocations to be mirrored to a child daos token allocation.

**Proposed Changes**
- When creating a subdao - a user has the ability to specify how many of the child tokens should be allocated to parent token holders.
- If there is an allocation present - the useCreateDAODataCreator.ts hook deploys a claimSubsidiary contract from the claimFactory, adds the parentAllocationAmount to the tokenAllocations array, send the allocated amount to the metafactory, and calls init on the claimSubsidiary contract by taking a snapshot of the current balance and sending the tokens to the claim contract.
- Create a use claimSubsidiary hook to display the contract address to the user in dao details.
- Allocation validation that should be <= totalSupply  

**Testing Purposes**
1. Should not show parentAllocationAmount when creating a rootDAO
image

2. Should fail with proposalAmount + allocations > total supply
image

3. Should be able to execute transaction
image

4. Shows Child DAO
image

5. Shows claimSub on Child DAO
image

6. Everyone has the correct amount of tokens - Claim Sub has correct tokens
image

7. Parent token holder can claim correct amount of tokens
image


src/components/Dao/ClaimToken.tsx
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  return (
+    <ContentBox isLightBackground>
+      <div className="flex justify-between items-center">
+        <div className="text-gray-25">
+          {`You have ${ethers.utils.formatUnits(
+            userClaimAmount.toString(),
+            decimals
+          )} ${symbol} available to claim!`}
```
Da-Colon, line 38: Can we move this formatting to outside the HTML?

```tsx
// change the name
const formattedValue = useMemo(() => ethers.utils.formatUnits(
            userClaimAmount.toString(),
            decimals
          ), [userClaimAmount, decimals])
```

or move it to the hook?

src/components/DaoCreator/TokenAllocations.tsx
```diff
@@ -46,14 +53,33 @@ function TokenAllocations({ tokenAllocations, supply, fieldUpdate }: TokenAlloca
...
     const totalAllocated = tokenAllocations.reduce((prev, cur) => Number(cur.amount) + prev, 0);
-    if (Number(supply) && totalAllocated) {
+    if (Number(supply) && Number(parentAllocationAmount)) {
+      setAmountError(Number(supply) < Number(parentAllocationAmount));
+      if (Number(supply) && totalAllocated) {
+        setAmountError(Number(supply) < totalAllocated + Number(parentAllocationAmount));
+      }
+    } else if (Number(supply) && totalAllocated) {
       setAmountError(Number(supply) < totalAllocated);
     }
```
Da-Colon, line 63:
```tsx
if (Number(supply)) {
  // no DAO token allocation with no parent allocations
  if (Number(totalAllocated) && !Number(parentAllocationAmount)) {
    setAmountError(Number(supply) < totalAllocated)
  // parent tokens allocated but no DAO token allocation
} else if (!Number(totalAllocated) && Number(parentAllocationAmount)) {
  setAmountError(Number(supply) < Number(parentAllocationAmount))
  // parent tokens allocated with DAO token allocation
  } else if (Number(totalAllocated) && Number(parentAllocationAmount)) {
    setAmountError(supply) < totalAllocated + Number(parentAllocationAmount))
  } else {
  // no allocation set amount error to false
    setAmountError(false)
  }
}
```

src/components/DaoCreator/provider/hooks/useNextDisabled.ts
```diff
@@ -23,17 +23,18 @@ export function useNextDisabled(state: CreatorState) {
         break;
       case CreatorSteps.TREASURY_GOV_TOKEN:
         if (state.govToken) {
-          if (
-            state.govToken.tokenAllocations === undefined ||
-            state.govToken.tokenSupply === undefined
```
Da-Colon, line 28: `state.govToken.tokenSupply === undefined` -> `!state.govToken.tokenSupply`
`state.govToken.tokenAllocations` -> `!state.govToken.tokenAllocations`

Da-Colon (changes requested):

src/components/Dao/ClaimToken.tsx
christopherdancy: Decided not to useMemo since it was being rendered conditionally. Also - decided not to render in hook since this is not the current pattern within daoData.ts.

christopherdancy:
Looks like everything is good to go? Should we merge or are we doing some more QA?

src/components/Dao/ClaimToken.tsx
```diff
@@ -0,0 +1,56 @@
...
+    () =>
+      userClaimAmount && decimals
+        ? ethers.utils.formatUnits(userClaimAmount.toString(), decimals)
+        : undefined,
+    [userClaimAmount, decimals]
+  );
+
+  if (!formattedValue || !Number(formattedValue)) {
+    return null;
+  }
```
Da-Colon, line 37: I did end up moving to the `useMemo` hook by updating the conditional to check for this value to be defined here.

Da-Colon:
So not quite ready yet.

image

image

So Context: I set the Parent DAO allocation to 50/100 tokens.

There is only 1 member of parent DAO (me). So all 50 of these should be set to me to claim yeah? 10 tokens are missing out in the either somewhere? It's not in either Treasury? @christopherdancy @tbwebb22

Da-Colon (approved):
The workflow is functioning as expected. A issue will be created to discuss and handle parent treasury claiming

tbwebb22:
Just completed some QA as well, looking good to me.

PR #386, opened 2022-07-27
Token claim

tbwebb22:
This PR:
 - Adds hooks and types for getting user token claim amounts
 - Adds a claim token component that enables a user to claim tokens for a new child DAO. This component renders on  the DAO summary page only when a user has more than zero tokens to claim

### DIFF
--- a/src/components/Dao/ClaimToken.tsx
+++ b/src/components/Dao/ClaimToken.tsx
@@ -1,0 +1,43 @@
+import { useState } from 'react';
+import ContentBox from '../ui/ContentBox';
+import { PrimaryButton } from '../ui/forms/Button';
+import { useDAOData } from '../../contexts/daoData';
+
+function ClaimToken() {
+  const [pending, setPending] = useState<boolean>(false);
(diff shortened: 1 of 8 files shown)


src/components/Dao/ClaimToken.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const claimToken = useClaimToken({
+    setPending,
+  });
+
+  if (userClaimAmount === undefined || userClaimAmount.eq(0)) {
+    return null;
+  }
+
+  return (
+    <ContentBox isLightBackground>
```
christopherdancy, line 32: Could we do a simple {Number(userClaimAmount) > 0 && </>}? Does the null add some extra features?

christopherdancy (approved):
Otherwise LGTM seen it work with my own eyes

src/components/Dao/ClaimToken.tsx
adamgall: Casting that `userClaimAmount` (which is a `BigNumber`, I assume), to a `Number` seems like a bad idea imo.

Issue #387, opened 2022-07-28
Proposal created without Voting powers fail


Da-Colon:
Proposal's created this way end up failing while queuing

adamgall:
what does this mean? what should the fix look like?

Da-Colon:
O yeah good call. This happens when I create a new DAO, do the token allocation to myself. But before 'Delegating' voting rights and I try to create a Proposal. The proposal will be created but will fail when attempting to `Queue Proposal`.  I don't have all the context on how queuing works to say that the solution here is.

adamgall:
I wonder if it's easy to check the "total delegated votes" for a given token? If so, we could disable the "create proposal" button if that number is 0.

@christopherdancy thoughts on this approach?

PR #409, opened 2022-08-10
338 create proposal tokens required

### LINKED ISSUE #338: Tokens required for creating a proposal
If a DAO requires users to have a token allocation before creating a proposal. We should check this on the frontend so the user does not revert a transaction.

Da-Colon:
## Overview

Updates the create proposal form validation to check for voting weight and proposal threshold. Displaying a toast to the user for information on this.

## Screenshots
image

image

closes #338


src/pages/ProposalCreate/index.tsx
```diff
@@ -92,18 +96,53 @@ function ProposalCreate() {
...
   }, [transactions, proposalDescription]);
 
-  const isValidProposalValid = useCallback(() => {
+  const canUserCreateProposal = useMemo(() => {
+    // disable while threshold and voting weight are loading
+    if (votingWeight === undefined || proposalTokenThreshold === undefined) {
+      return false;
+    }
+    // disable if the user's voting veight is 0
+    if (votingWeight.isZero()) {
```
adamgall, line 105: I'm not convinced that this is proper check in this scenario.

I'm under the impression that it's perfectly valid for a user with no voting weight to create a proposal. Isn't that the default setting when creating a new DAO?

I _believe_ (my hunch is) the issue you experienced in #387 is because there were zero TOTAL delegated voting power when the proposal was created.

Maybe I'm off base here. But something about this solution doesn't feel quite right to me.

Da-Colon: @adamgall I'll spin out with Chris to understand whats needed here better.

Da-Colon: Ah I see your comment there now. so instead of this check this would be that the TOTAL is not 0

Da-Colon: @adamgall @christopherdancy I read through the votes token and it's openzeplin extensions and from I can tell, the only way to check for this is by checking the DelegateChangedEvent and checking it's length > 0

```ts
    const isDelegatesSet = await (
      await tokenContract.queryFilter(tokenContract.filters.DelegateChanged())
    ).length;
```

adamgall: It would be good to isolate this as the actual problem though. As of now it is just a hunch I have. Can you reliably recreate the issue? Sounds like it would be a good unit test in the https://github.com/decent-dao/fractal-module-governor repo.

Da-Colon: can confirm that your hunch is correct.

DAO Threshold: 0
User Creating Proposal Weight: 0

This proposal can be executed.

So with that, is it safe to assume after the first person has delegated this is always true? can this weight ever be removed or just the delegation changed?

Da-Colon: Awesome thanks for your help earlier @adamgall. On top of updating the 'create proposal' page with these validation. I also went ahead and added this validation to the plugins with @xraystyle1980 help. (thanks matt). The toast that appear now show a clickable link to the delegate page within the governance module.

Da-Colon: image

Da-Colon: ignore the quotes in the toast, a commit to fix has been pushed.

adamgall (approved):

PR #438, opened 2022-08-22
updates github workflow secrets for dev release

Da-Colon:
## Overview
Updates IPFS secrets in github actions

## Note
Secrets in Fractal-interface settings have also been updated.


adamgall (approved):
LGTM

adamgall:
One thing to note, is that we'll have to remember to do this for the prod environment whenever we do our next production update.

Da-Colon:
I agree. it wouldn't hurt to go ahead and update now?

adamgall:
> I agree. it wouldn't hurt to go ahead and update now?

Some of the contracts simply don't exist on mainnet yet, and others are still pointing at old versions. So I don't think it's possible to update them now.

Da-Colon:
I was thinking just the prod workflow file, not actually adding the secrets to github.

adamgall:
> I was thinking just the prod workflow file, not actually adding the secrets to github.

Ah yes! This can be done now.

adamgall (commented):
@Da-Colon can you verbally confirm that all of the contract address env values in the dev yaml are captured in the prod yaml?

Da-Colon:
@adamgall I can confirm that both env files match....although......looking at the environmental varibles on github's prod. I don't see `REACT_APP_TIMELOCK_ADDRESSES` rather it's `REACT_APP_TIMELOCKUPGRADEABLE_ADDRESSES`

Da-Colon:
Which I guess is fine for now, but we will have to make sure we update this when the develop is merged and the secrets are updated.

adamgall:
> @adamgall I can confirm that both yaml files match....although......looking at the environmental varibles on github's prod. I don't see `REACT_APP_TIMELOCK_ADDRESSES` rather it's `REACT_APP_TIMELOCKUPGRADEABLE_ADDRESSES`

are you able to update that? i think you should be able to get the existing value from netlify

Da-Colon:
well the codebase deployed for prod, uses that variable. So I think it's safe to stay the same until master is updated?

adamgall (approved):

PR #440, opened 2022-08-22
fix missing types

Da-Colon:
## Overview

When updating the dependencies, this error popped. weird it didn't show it's face before. Anyway added the missing types

PR #441, opened 2022-08-22
update dependencies

Da-Colon:
## Overview

Updates the dependencies

## Note
- The PR will probably fail, until #440 is merged in.


package.json
```diff
@@ -2,33 +2,33 @@
...
   "dependencies": {
-    "@fractal-framework/core-contracts": "^0.1.8",
-    "@headlessui/react": "^1.5.0",
-    "@testing-library/jest-dom": "^5.16.2",
-    "@testing-library/react": "^12.1.3",
-    "@testing-library/user-event": "^13.5.0",
+    "@fractal-framework/core-contracts": "^0.1.9",
+    "@headlessui/react": "^1.6.6",
+    "@testing-library/jest-dom": "^5.16.5",
+    "@testing-library/react": "^13.3.0",
```
adamgall, line 8: What are the breaking changes in this major version upgrade, and do we need to update any of our code to account for those breaking changes?

package.json
```diff
@@ -2,33 +2,33 @@
...
+    "@testing-library/user-event": "^14.4.3",
```
adamgall, line 9: What are the breaking changes in this major version upgrade, and do we need to update any of our code to account for those breaking changes?

package.json
```diff
@@ -2,33 +2,33 @@
...
     "@tippyjs/react": "^4.2.6",
-    "@types/jest": "^27.4.1",
-    "@types/node": "^16.11.26",
-    "@types/react": "^17.0.39",
-    "@types/react-dom": "^17.0.13",
+    "@types/jest": "^28.1.7",
```
adamgall, line 11: What are the breaking changes in this major version upgrade, and do we need to update any of our code to account for those breaking changes?

package.json
```diff
@@ -2,33 +2,33 @@
...
+    "@types/node": "^18.7.9",
```
adamgall, line 12: What are the breaking changes in this major version upgrade, and do we need to update any of our code to account for those breaking changes?

package.json
```diff
@@ -2,33 +2,33 @@
...
+    "@types/react": "^18.0.17",
```
adamgall, line 13: What are the breaking changes in this major version upgrade, and do we need to update any of our code to account for those breaking changes?

package.json
```diff
@@ -2,33 +2,33 @@
...
+    "@types/react-dom": "^18.0.6",
```
adamgall, line 14: What are the breaking changes in this major version upgrade, and do we need to update any of our code to account for those breaking changes?

package.json
```diff
@@ -2,33 +2,33 @@
...
     "@ukstv/jazzicon-react": "^1.0.0",
-    "@walletconnect/ethereum-provider": "^1.7.3",
+    "@walletconnect/ethereum-provider": "^1.8.0",
     "classnames": "^2.3.1",
-    "ethers": "^5.5.4",
+    "ethers": "^5.7.0",
     "process": "^0.11.10",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "react": "^18.2.0",
```
adamgall, line 20: What are the breaking changes in this major version upgrade, and do we need to update any of our code to account for those breaking changes?

package.json
```diff
@@ -2,33 +2,33 @@
...
+    "react-dom": "^18.2.0",
```
adamgall, line 21: What are the breaking changes in this major version upgrade, and do we need to update any of our code to account for those breaking changes?

adamgall (commented):
This is a scary PR that makes me nervous when we don't have any automated tests lol

Da-Colon:
Yessir There is no rush to this PR. I wanted to show what npm outdated is showing. after viewing the data from `npm outdated` and doing a little digging. I use `npm-check-updates` https://www.npmjs.com/package/npm-check-updates. 

I'll double back and QA and check on the major version changes.

package.json
Da-Colon: https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html.

Easy update, only a warning. This has been updated.

package.json
Da-Colon: see below comment

package.json
Da-Colon: I've read through the docs and there doesn't appear to be any breaking changes that will effect our app

Da-Colon (commented):
Overall I think we are good to go. It appears that all the major version changes are related to React update to 18 (which reading through the docs seem like we are good to go.

I have tested creating a parent DAO of both types. and create a subDAO Token Governance. No other obviously errors with the files or workflow.

@SudoPumpkin can you give it a run through as well in the deploy preview?

adamgall (approved):
welp YOLO, as good of a time as any to do a bunch of package updates

SudoPumpkin:
I didn't see any glaring issues. I did a quick pass.

Da-Colon:
Here is one big plus for this update (Gonna wait till Monday to merge) https://blog.bitsrc.io/automatic-batching-in-react-18-what-you-should-know-d50141dc096e

adamgall:
Can you give a tldr lol, do we have to do anything special to receive those benefits?

Da-Colon:
TL:DR 

> as you web application grows, the number of nested components will increase. Therefore, if a parent component executes an un-batched state update, the entire component tree will be re-rendered per state update.

> This behavior can cause severe performance issues related to application speed

> React v18 ensures that state updates invoked from any location will be batched by default. This will batch state updates, including native event handlers, asynchronous operations, timeouts, and intervals.

There is a little more to this but basically say you have an a few different `useState()` 

```ts
const [loading, setLoading] = useState(true);
const [data, setData] = useState({});

useEffect(() => {
  (async () => {
    const response = await fetch(someURL);
   setLoading(false);
   setData(response);
   )();
```

So on a small local level you woudn't really tell much different but on a larger scale eash of the 'setState' here causes a re-render. This can cause unexpected side-effects and re-rerenders. Batching allows for both these to be updated in a call and only re-render once.

There is no updates that we need to do currently for the benefit. They do mention:

> Automatic batching is indeed an amazing feature. But, there can be situations where we need to prevent this from happening. For that, React provides a method named flushSync() in react-dom that allows us to trigger a re-render for a specific state update.

SudoPumpkin (approved):
This looks to be working as intended.

PR #453, opened 2022-08-24
BugFIx Token Supply/Amount Error

Da-Colon:
Thankfully this bug was not caused by any math error, rather it caused by the update that changed from using the `tokenAllocation` prop. The prediction was mapping the original prop array and not the updated array with the additional allocations

### DIFF
--- a/src/hooks/useCreateDAODataCreator.ts
+++ b/src/hooks/useCreateDAODataCreator.ts
@@ -173,8 +173,8 @@ const useCreateDAODataCreator = () => {
             [
               tokenName,
               tokenSymbol,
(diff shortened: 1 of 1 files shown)


adamgall (approved):

adamgall:
@Da-Colon before merging, let's let @SudoPumpkin run through her manual test and give her own approval.

christopherdancy (commented):
LFG

SudoPumpkin (approved):
This appears to be working as intended on https://deploy-preview-453--fractal-framework-interface-dev.netlify.app/#/daos/new

PR #454, opened 2022-08-24
Deploy a Gnosis Safe SubDAO with Token Voting Parent DAO

christopherdancy:
PR changes one file - TokenGovernanceInjector.

The intent of this PR is to provide the user with the ability to propose a gnosis / token voting sub DAO. This is accomplished by utilising cases to determine if the proposal should include the gnosis or token voting encoded data.

Things to note: The 1:1 token voting does not work due to a small bug which has been addressed in #453 - the bug will be fixed during the merge.

To test this PR - please try to create a gnosis Safe subdao

### DIFF
--- a/src/controller/Modules/TokenGovernanceInjector.tsx
+++ b/src/controller/Modules/TokenGovernanceInjector.tsx
@@ -72,7 +72,7 @@ export function GovernanceInjector({ children }: { children: JSX.Element }) {
 
       const data: ExecuteData = {
         targets: [metaFactoryContract.address],
-        values: [0],
+        values: [BigNumber.from('0')],
         calldatas: [
           metaFactoryContract.interface.encodeFunctionData('createDAOAndExecute', [
             newDAOData.calldata.daoFactory,
@@ -92,7 +92,7 @@ export function GovernanceInjector({ children }: { children: JSX.Element }) {
             // ERC20 transfer
             // Approve the new treasury to transfer tokens from the DAO
             data.targets.push(tokenToFund.asset.contractAddress);
-            data.values.push(0);
+            data.values.push(BigNumber.from('0'));
             data.calldatas.push(
               ERC20__factory.createInterface().encodeFunctionData('approve', [
                 newDAOData.predictedTreasuryAddress,
@@ -102,7 +102,7 @@ export function GovernanceInjector({ children }: { children: JSX.Element }) {
 
             // Withdraw tokens from the parent treasury into the parent DAO
             data.targets.push(treasuryModuleContract.address);
-            data.values.push(0);
+            data.values.push(BigNumber.from('0'));
             data.calldatas.push(
               TreasuryModule__factory.createInterface().encodeFunctionData('withdrawERC20Tokens', [
                 [tokenToFund.asset.contractAddress],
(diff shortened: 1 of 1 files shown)


Da-Colon (changes requested):
LFG. Successfully created a Gnosis Safe subDAO. Few minor 'clean up as you go comments' and we are on our way.

src/controller/Modules/TokenGovernanceInjector.tsx
```diff
@@ -34,146 +39,218 @@ export function GovernanceInjector({ children }: { children: JSX.Element }) {
...
-          data.values.push(0);
-          data.calldatas.push(
-            ERC20__factory.createInterface().encodeFunctionData('approve', [
-              newDAOData.predictedTreasuryAddress,
-              ethers.utils.parseUnits(tokenToFund.amount.toString(), tokenToFund.asset.decimals),
-            ])
-          );
+      const data: ExecuteData = {
+        targets: [metaFactoryContract.address],
+        values: [0],
```
Da-Colon, line 75: Can you BigNumber all the 0 values?

adamgall: @christopherdancy to hopefully provide more context to @Da-Colon 's request: functionally, it doesn't matter if `0` is a `BigNumber` or a `Number`. ethers.js will handle it just fine.

But from the point of view of our mental model, and the app's understanding, of types, it's best that we declare things as the "correct" type initially.

Since this value represents an amount of ether, `BigNumber` is the correct type.

If this is declared as a `Number` here, then at some point later down the execution stack, typescript will just inform us that the thing is a `Number`, and the programmer won't know if it's ok for them to cast/treat it as a `BigNumber`, if the situation arises where they might want/need to do so.

Declaring it as a `BigNumber` here means that down the execution stack, the programmer knows that it can be treated as a `BigNumber`.

Da-Colon: 🥇 exactly thanks for going into more detail @adamgall

christopherdancy: I see I see! perfect - I was wondering what the need for this was.

adamgall:
@christopherdancy fwiw, you can save a few keystrokes and instantiate a `BigNumber` from a number type, as well as a string 

For example

`BigNumber.from(0)`

Instead of

`BigNumber.from('0')`

adamgall (approved):

SudoPumpkin:
Are subDAO for Gnosis is not functioning yet?
Example DOA:
0x57232073Bba00C00db5a65920a1bEdF6D6FF7F08

image

Da-Colon:
> Are subDAO for Gnosis is not functioning yet?

@SudoPumpkin Is this on Rinkeby? This Branch doesn't contain `rinkeby` support just yet. Can you try on Goerli?

I just made this one fresh on Goerli. 0x489dB2adf2551c20b97515802A0289d0478c02e2

Da-Colon (approved):

SudoPumpkin:
> > Are subDAO for Gnosis is not functioning yet?
> 
> @SudoPumpkin Is this on Rinkeby? This Branch doesn't contain `rinkeby` support just yet. Can you try on Goerli?
> 
> I just made this one fresh on Goerli. 0x489dB2adf2551c20b97515802A0289d0478c02e2

Yeah, Rinkeby.

Da-Colon:
@SudoPumpkin Yeah this branch doesn't contain the Rinkeby Support Merg

SudoPumpkin (approved):
This looks to be working as intended on preview-454

PR #463, opened 2022-08-26
adds case for choose governance step

Da-Colon:
## Overview
Problem: Next button disable state remains when navigating back to `Choose Governance`. Update switch statement for case for this step.


SudoPumpkin (approved):
This appears to be working as intended on deploy-preview-463

Issue #459, opened 2022-08-25
Subsidary Funding


Da-Colon:
Getting a similar issue when trying to fund a subDAO from a parent DAO treasury 1:1 Token Voting Governance root a 1:1 Token Voting Governance subDAO. 

Repl:
- In a 1:1 Token Voting Governance Parent DAO. 
- Send funds to Treasury (0.05ETH)
- Click Create a subDAO
- Add ETH
- Set to 0.05
- Set other configurations (new subDAO token as normal
- Click Deploy......you will get an error

christopherdancy:
This error appears to occur before the actual proposal is created.

PR #465, opened 2022-08-29
update if statement to handle eth and erec20

christopherdancy:
This PR handles the #459 issue relating to the funding of sub daos. The issue - subdao proposals could not be created with subdaofunding. The reason for this error was created when checking if Ethereum was == address.zero - ethereum is not address.zero - it is an empty string.

Screenshot 2022-08-29 at 10 40 57


src/controller/Modules/TokenGovernanceInjector.tsx
```diff
@@ -86,7 +86,7 @@ export function GovernanceInjector({ children }: { children: JSX.Element }) {
 
     if (daoData.tokensToFund.length > 0) {
       daoData.tokensToFund.forEach(tokenToFund => {
-        if (tokenToFund.asset.contractAddress !== ethers.constants.AddressZero) {
+        if (tokenToFund.asset.contractAddress !== '') {
```
adamgall, line 89: Playing devil's advocate to this solution: should the value of `contractAddress` _actually_ be the zero address? Seems like that would make it more "explicit", vs the empty string value it currently has, which feels more "ambiguous" regarding "why it's an empty string"

adamgall (commented):
Left a comment

src/controller/Modules/TokenGovernanceInjector.tsx
adamgall: Is there anywhere else in the codebase that's expecting this `contractAddress` to be either an empty string or the Zero Address?

adamgall, line 89: Is there anywhere else in the codebase that's expecting this contractAddress to be either an empty string or the Zero Address?

adamgall (commented):
Left a question

adamgall:
@christopherdancy can you please address my comments/questions, and resolve the conflict, so we can get this PR merged? Thanks!

christopherdancy:
My apologies for the delay @adamgall. My day was packed yesterday - along with trying to get this other sprint completed. It is tough to go back. Reviewing now.

src/controller/Modules/TokenGovernanceInjector.tsx
christopherdancy: I am a tad confused by the question @adamgall, If the type is a string then it could be any arb string. Are you saying for the string to be casted to the zero address before it is passed down to this component?

christopherdancy: My next question - what is the concept of "contractAddress" when referring to ETH?

Da-Colon:
@christopherdancy can you fix the conflicts? (rebase on develop)

adamgall (approved):

PR #464, opened 2022-08-26
Gnosis Safe Provider Updates

Da-Colon:
## Overview
Updates the GnosisWrapper Provider to retrieve and store data from Gnosis Safe API

- Updates Provider with a reducer to handle state updating
- Retrieves and stores safe information in state.

PR #468, opened 2022-08-30
Update/create proposal for reusablity

Da-Colon:
## Overview

Extracted from #464 for easier review. This PR update the create proposal components (manually creating transactions) UI components to be used by Gnosis Governance Provider. 

## Note
- This is non-functioning update. 
- It does add a button on the Gnosis Module Page but this Button is removed in #464.


src/pages/ProposalCreate/index.tsx
```diff
@@ -19,20 +17,26 @@ const defaultTransaction = {
...
+
+function ProposalCreate({ submitProposal, pendingCreateTx, isUserAuthorized }: IProposalCreate) {
   const {
     dao: { daoAddress },
-    modules: { tokenVotingGovernanceModule },
   } = useFractal();
 
-  const {
-    createProposal: { submitProposal, pendingCreateTx },
-  } = useGovenorModule();
```
Da-Colon, line 30: Removes dependency on solely the Governor Module to function. needed props are passed to the components from parent

src/pages/GnosisWrapper/index.tsx
```diff
@@ -11,6 +13,9 @@ export function GnosisWrapper() {
         address={gnosisSafeAddress}
         label="Gnosis Safe"
       />
+      <NavLink to="proposals/new">
+        <PrimaryButton label="Create Proposal" />
+      </NavLink>
```
Da-Colon, line 18: This button is removed in #464

adamgall:
@Da-Colon when you say "this is a non-functioning update", what does that mean?

It looks like there are code changes that are utilized in the running app (changes to `ProposalCreate.tsx`, and `Proposals.tsx`, at a minimum.

Are you saying that there are no visible changes from the user's point of view?

Da-Colon:
Yes there is no visible changes and nothing new to interact with. I was going for no QA is needed. only a code review. But as I saw that someone else should run through create Proposal (1:1 Token Voting) to double check nothing it out of ordinary

src/controller/Modules/index.tsx
```diff
@@ -103,10 +76,10 @@ export function Modules() {
   }, [params.moduleAddress, state.moduleAddress, state.isLoading]);
 
   useEffect(() => {
-    if (!state.isLoading && !state.moduleType && dao.isLoading) {
+    if (!state.isLoading && !state.moduleType) {
       navigate(`/daos/${dao.daoAddress}`, { replace: true });
     }
-  }, [state.isLoading, state.moduleType, dao.daoAddress, navigate, dao.isLoading]);
+  }, [state.isLoading, state.moduleType, dao.daoAddress, navigate]);
```
adamgall, line 82: Why no more `dao.isLoading`?

src/pages/GnosisWrapper/routes/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+      />
+      <Route
+        index
+        element={
+          <Navigate
+            to="./.."
+            replace={true}
+          />
+        }
+      />
```
adamgall, line 22: Two index routes?

src/pages/GnosisWrapper/routes/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <Route
+        path="proposals/new"
+        element={
+          <ProposalCreate
+            submitProposal={createProposal}
+            pendingCreateTx={createPendingTx}
+            isUserAuthorized={true}
+          />
```
adamgall, line 30: I assume this is the same component that gets rendered on the traditional DAO's "new proposals" page? If so: nice

src/pages/ProposalCreate/index.tsx
```diff
@@ -117,12 +119,12 @@ function ProposalCreate() {
...
-    () => !canUserCreateProposal || !proposalDescription.trim().length,
-    [canUserCreateProposal, proposalDescription]
+    () => !isUserAuthorized || !proposalDescription.trim().length,
+    [isUserAuthorized, proposalDescription]
   );
   const isCreateDisabled = useMemo(
-    () => !canUserCreateProposal || !isValidProposal || pendingCreateTx,
-    [pendingCreateTx, isValidProposal, canUserCreateProposal]
+    () => !isUserAuthorized || !isValidProposal || pendingCreateTx,
+    [pendingCreateTx, isValidProposal, isUserAuthorized]
```
adamgall, line 127: This is great, thanks for passing this in as a dependency. Makes sense given that there will be different conditions for "can the connected user create a proposal" given the type of governance.

src/pages/Proposals/index.tsx
```diff
@@ -1,9 +1,15 @@
...
 
 import ProposalDetails from '../../components/Proposals/ProposalDetails';
+import { useGovenorModule } from '../../providers/govenor/hooks/useGovenorModule';
+import { useUserProposalValidation } from '../../providers/govenor/hooks/useUserProposalValidation';
 import ProposalCreate from '../ProposalCreate';
 
 function Proposals() {
+  const {
+    createProposal: { submitProposal, pendingCreateTx },
+  } = useGovenorModule();
```
adamgall, line 11: I assume that in the next PR is where more/different logic exists to allow the Gnosis module to be used here too?

adamgall (approved):
Left some comments and questions, but nothing blocking. Nice job!

adamgall:
@mudrila @christopherdancy @tbwebb22 @SudoPumpkin another PR Review from one of you would be appreciated!

src/controller/Modules/index.tsx
Da-Colon: uanessacry dependency. since the state.isLoading is `true` until `dao.IsLoading` is false anyway

src/pages/GnosisWrapper/routes/index.tsx
Da-Colon: yessir

src/pages/GnosisWrapper/routes/index.tsx
Da-Colon: good catch, copied over from proposals route without double checking. updated with a path to push users back to DAO home if the URL changes.

src/pages/GnosisWrapper/routes/index.tsx
```diff
@@ -12,7 +12,7 @@ export function GnosisRoutes() {
         element={<GnosisWrapper />}
       />
       <Route
-        index
+        path="proposals/*"
```
adamgall, line 15: Yes this looks better. The idea is likely that we want to support `/proposals/new`, but redirect back to `/` on `/proposals` and/or `/proposals/anything-else`.

(right? i'm pretty sure there's no content at `/proposals`. definitely want to redirect back when user navigates to `/proposals/anything-else`)

adamgall (approved):
Approving again, with the assumption that the behavior described in my comment is what we want and what is implemented.

src/pages/Proposals/index.tsx
Da-Colon: Yes a similar setup is in `src/pages/GnosisWrapper/routes/index.tsx` the create proposal method would come from the provider of the correct Governance module. 

Currently in the Gnosis Wrapper in #454 it still will be just a `() => {}` I think once the work is finished around deploying the subDAO. should be pretty non-trivial to create a the hook similar to `useCreateProposal` but for Gnosis

PR #475, opened 2022-09-01
Dao creator/token supply bn support

Da-Colon:
## Description

Updates to fully support BigNumbers smoothly In Token Supply Input.

## Notes

- For Now test with a single allocation and 1 allocated. The Token Allocation field will be updated in following PR.

## Issue (if applicable)
Will close Issue after final PR

## Testing
Create a DAO -> Esstentials -> Choose Governance (1:1 Token Voting) -> Treasury and Governance Token

- Token Supply: User should be able to type Large Number and decimals up to 18
- Token Supply: User should only type numbers
-> Governance Config -> Deploy

- User should be able to Deploy a new DAO with current Token Supply

## Screenshots (if applicable)
image

PR #477, opened 2022-09-01
Parent Token Allocation BigNumber Support

### LINKED ISSUE #456: Implement fractional numbers support with BigNumber
For some values that are held with BigNumber - we need to implement support for fractional values. 

Possible solution could be overriding BigNumber library in etherjs

### LINKED ISSUE #473: Create Token Voting DAO from GnosisSafe parent

Da-Colon:
## Description

Updates to fully support BigNumbers smoothly In Token Supply Input.

## Notes

- For Now test with a single allocation and 1 allocated. The Token Allocation field will be updated in following PR.

## Issue (if applicable)
closes #473 
closes #456 

## Testing
Create a subDAO -> Esstentials -> Choose Governance (1:1 Token Voting) -> Treasury and Governance Token

- Parent Token Allocation: User should be able to type Large Number and decimals up to 18
- Parent Token Allocation: User should only type numbers
-> Governance Config -> Create proposal

- User should be able to Deploy a new subDAO with current Token Supply and Allocations


Da-Colon (commented):
Similar Changes as were made in #475. 3 of 3

src/components/DaoCreator/provider/CreatorProvider.tsx
```diff
@@ -99,7 +99,7 @@ export function CreatorProvider({ deployDAO, pending, isSubDAO, children }: ICre
       ..._initialState,
       govToken: {
         ..._initialState.govToken,
-        parentAllocationAmount: isSubDAO ? BigNumber.from(0) : undefined,
+        parentAllocationAmount: isSubDAO ? { value: '', valueBN: null } : undefined,
```
christopherdancy, line 102: I remember we had a push to add bignumber.from(0) for big numbers.

adamgall: @christopherdancy please explain the relevancy of that statement

src/components/DaoCreator/provider/hooks/useNextDisabled.ts
```diff
@@ -41,7 +41,7 @@ export function useNextDisabled(state: CreatorState) {
           const isAllocationsValid = govToken.tokenAllocations
             .map(tokenAllocation => tokenAllocation.amount.bigNumberValue || BigNumber.from(0))
             .reduce((prev, curr) => prev.add(curr), BigNumber.from(0))
-            .add(govToken.parentAllocationAmount.bigNumberValue || 0)
+            .add(govToken.parentAllocationAmount?.bigNumberValue || BigNumber.from(0))
```
adamgall, line 44: FWIW leaving this at `0` is totally fine in this context, because the output (which is "used" in the next `lte` comparison) is still a `BigNumber` object.

You can pass `BigNumberish` into a `BigNumber` function operation (like `.add()` here), and it'll output a `BigNumber` always.

adamgall (approved):
Easy peasy, this gets a LGTM

src/components/DaoCreator/provider/CreatorProvider.tsx
```diff
@@ -99,7 +99,7 @@ export function CreatorProvider({ deployDAO, pending, isSubDAO, children }: ICre
       ..._initialState,
       govToken: {
         ..._initialState.govToken,
-        parentAllocationAmount: isSubDAO ? BigNumber.from(0) : undefined,
+        parentAllocationAmount: isSubDAO ? { value: '', bigNumberValue: null } : undefined,
```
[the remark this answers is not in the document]
Da-Colon: This `undefined` state currently controls whether that input field is rendered in the form. The UI layer of the app doesn't have the concept of DAO or subDAO. That is handled by this state thats set when it's mounted. 

if it it's set as an object (`value` (for inputs) and `bigNumberValue`) then field will be visible to a user and can be updated.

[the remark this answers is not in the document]
herbig: ahh, that makes sense, thanks for the clarification

PR #474, opened 2022-09-01
Deploy gnosis from gnosis

christopherdancy:
## Description

A gnosis safe may deploy a gnosis subDAO or a 1:1 token voting DAO.

## Notes

1. Currently, if a user is attempting to deploy a 1:1 token voting DAO from a Gnosis Safe, the Funding Details page is skipped because there is no way to know the tokens which exist within a gnosis safe since our current Interface only recognises a treasury balance through events. 
2. The parent Allocation form is skipped because the parent does not have a token which can be snapshotted and forked.

## Issue (if applicable)

closes #437 
closes #473

PR #482, opened 2022-09-06
Combine char restrictions

Da-Colon:
## Description
replaces the char restriction props into a single ENUM controlled prop

## Notes
Solves issue where the Input component can be implemented with `isWholeNumbers` and `isFloatNumbers`, This will cause the two methods to conflict with each other.

## Issue (if applicable)

discussed in #474

## Testing

no Visual QA is needed. More just a technical review

n/a

### DIFF
--- a/src/components/ui/forms/Input.tsx
+++ b/src/components/ui/forms/Input.tsx
@@ -145,12 +145,18 @@ function Input({
       if (decimals) {
         limitDecimals(event);
       }
-      if (restrictChar === RestrictCharTypes.WHOLE_NUMBERS_ONLY) {
-        wholeNumbersOnly(event);
+
+      switch (restrictChar) {
+        case RestrictCharTypes.WHOLE_NUMBERS_ONLY: {
+          wholeNumbersOnly(event);
+          break;
+        }
+        case RestrictCharTypes.FLOAT_NUMBERS: {
+          floatNumbersOnly(event);
+          break;
+        }
       }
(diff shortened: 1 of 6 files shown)


src/components/ui/forms/Input.tsx
```diff
@@ -139,17 +142,18 @@ function Input({
...
-      }
       if (decimals) {
         limitDecimals(event);
       }
+      if (restrictChar === RestrictCharTypes.WHOLE_NUMBERS_ONLY) {
+        wholeNumbersOnly(event);
+      }
+      if (restrictChar === RestrictCharTypes.FLOAT_NUMBERS) {
+        floatNumbersOnly(event);
+      }
```
adamgall, line 153: nitpick

The interface suggests that only one of these types can be set on an instance of `Input`, but this implementation is written in a way that, at first glance, suggests that both of those `if` blocks can be entered.

Can the implementation here be changed to a `switch`, which clearly shows that only ONE of the types will be enforced?

adamgall (approved):
Approving this, but also asking for a small change which is very nitpicky

src/components/ui/forms/Input.tsx
Da-Colon: Ha i had thought about it but didnt. For sure i will get added before merging

adamgall: I did this in commit https://github.com/decent-dao/fractal-interface/pull/482/commits/f7e8cc954e0bec787d0147aacc66eb5f50ed5947

PR #489, opened 2022-09-12
subsidary funding to support big numbers

Da-Colon:
## Description

This small update adds subsidiary funding bignumber support

## Notes
n/a

## Testing

- Create a 1:1 Token Voting subDAO from a 1:1 Token Voting Parent and fund using a large decimal number.

n/a

### DIFF
--- a/src/controller/Modules/TreasuryController.tsx
+++ b/src/controller/Modules/TreasuryController.tsx
@@ -37,7 +37,7 @@ export function TreasuryController({ children }: { children: JSX.Element }) {
       );
     }
     default: {
-      return children;
+      return null;
     }
   }
 }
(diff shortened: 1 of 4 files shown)


adamgall:
@Da-Colon what needs to happen for this to go from Draft -> Open?

Da-Colon:
@adamgall This one is dependent on #488 as the changes are made after the file is moved

src/controller/Modules/TreasuryController.tsx
```diff
@@ -37,7 +37,7 @@ export function TreasuryController({ children }: { children: JSX.Element }) {
       );
     }
     default: {
-      return children;
+      return null;
```
Da-Colon, line 40: This fixes a bug when navigating to the plugins. When this mounts it defaults to children. Because of the governance injector using the context hook this causes an error. The solution here is to return null so that the component isn't mounted until the modules are loaded

adamgall (approved):
No comments, looks good, lesgo

herbig (approved):
ship it 🚢

PR #492, opened 2022-09-12
Fiat exchange rate

brianfryer:
## Description

This PR adds two (2) new hooks to the Treasury module that fetch/format prices from Coingecko, and calculate/format fiat amounts for native and ERC20 tokens owned by the DAO.

1. `src/providers/treasury/hooks/useTreasuryAssetsFungiblePrices.ts`
2. `src/providers/treasury/hooks/useTreasuryAssetsFungibleFiatAmounts.ts`

## Notes

- For performance reasons, I've opted to keep the fiat amount of each asset separate from the `Web3Token` class (`src/providers/treasury/models/Web3NFT.ts`). Doing so enables us to perform a single API request for all ERC20 tokens instead of one request per token.
- An option for `selectedCurrency` has been stubbed out. As part of a future iteration, we could implement a "currency selector" to allow folks to see prices in other, non-USD currencies.
- Lastly, CoinGecko returns a single, unformatted value for the currency amount. Our display requires this value to be formatted in specific ways. As such, the incoming API response is transformed like in the two examples below:

<table>
<thead>
<th>Example 1</th>
<th>Example 2</th>
</thead>
<tbody>
<tr>
<td>

```js
// api response
{
  usd: 1234.567
}

// transformed amount
{
  usd: {
    amount: 1234.567,
    currency: '1,234.56',
    formattedAmount: '1,234.567'
  }
}
```
</td>
<td>

```js
// api response
{
  usd: 9876
}

// transformed amount
{
  usd: {
    amount: 9876,
    currency: '9,876.00',
    formattedAmount: '9,876'
  }
}
```

</td>
</tr>
</tbody>
</table>

## Testing

`npm run test`

image


xraystyle1980:
Looks good! My only thoughts are the verbiage inside the tooltip. Can we do something like the attached screenshot? (whichever is preferred)

CleanShot 2022-09-13 at 11 27 53@2x

src/providers/treasury/types/index.ts
```diff
@@ -29,9 +30,24 @@ export interface TreasuryAssetFungible {
...
+  };
+}
+
+export interface TreasuryAssetFungibleFiatAmounts {
+  [address: string]: TreasuryAssetFungibleFiatAmount;
+};
+
+export interface TreasuryAssetFungiblePrice extends TreasuryAssetFungibleFiatAmount {};
+export interface TreasuryAssetFungiblePrices extends TreasuryAssetFungibleFiatAmounts {};
+
```
adamgall, line 50: We're talking about this on the live call, but my first reaction was that I don't understand the need for this abstraction.

src/providers/treasury/types/coingecko.ts
```diff
@@ -0,0 +1,5 @@
+export type CoingeckoApiResponse = {
```
adamgall, line 1: Nitpick: Capitalize the "G" in CoinGecko

src/utils/countDecimals.ts
```diff
@@ -0,0 +1,7 @@
+const countDecimals = (value: number) => {
+  if (!value || value % 1 === 0) return 0;
```
adamgall, line 2: Can `!value` be written as `value === 0`, to be more explicit about what is being checked for?

adamgall: Also suggest to leave a comment here, explaining what `value % 1 === 0` means (it's a whole number)

src/utils/countDecimals.ts
```diff
@@ -0,0 +1,7 @@
...
+  const count = value?.toString().split('.').pop();
```
adamgall, line 3: Will value ever be not defined? Doesn't look like. Can probably remove the `?` from `value?`.

src/utils/countDecimals.ts
```diff
@@ -0,0 +1,7 @@
...
+  return count?.length || 0;
```
adamgall, line 4: Will `count` ever be undefined here? My gut says no, due to previous assertions.

src/providers/treasury/utils/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  // 1234.567 -> 1,234.567, 1234 -> 1,234
+  const formattedAmount = numberWithCommas(amount, maxDecimals);
+
+  return {
+    [currencyKey]: {
+      amount,
+      currency,
+      formattedAmount,
+    },
+  };
```
adamgall, line 27: This isn't really needed to be dealt with, but, I know I'm going back on previous statements, but would importing and using `Numeral.js` make this code easier to understand and extend in the future? 

Thinking about internationalization support in the future.

adamgall (changes requested):
Looks great, but I haven't gotten to the beefy files yet. Requested a couple of changes, mostly small things. Will try to leave another review at some point today, but if other engineers can comb through the bigger files first, that'd be great!

src/index.css
```diff
@@ -111,7 +111,7 @@
 }
 
 #tooltip[data-placement^='top'] > #arrow {
-  top: -12px;
+  bottom: -4px;
```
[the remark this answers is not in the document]
brianfryer: This does not affect any existing tooltip, as no other tooltip uses `top` placement 🥳

src/utils/countDecimals.ts
brianfryer: While it appears that `count` would never be defined, according to typescript its value is possibly 'undefined' due to the `split/pop` methods being used.

We'll keep this in place to keep typescript from throwing an error.

src/utils/countDecimals.ts
brianfryer: While it appears that `count` would never be defined, according to typescript its value is possibly 'undefined' due to the split/pop methods being used.

We'll keep this in place to keep typescript from throwing an error.

src/providers/treasury/types/index.ts
brianfryer: The "usefulness" of having distinct types here is for clarity when viewing the return values from the new hooks:

1. The return value from `useTreasuryAssetsFungibleFiatAmounts` ([link](https://github.com/decent-dao/fractal-interface/pull/492/files#diff-de7a4f88b6e2f1095d07825e6bdd34a190b1cec864dac0f540169791c773bca3R24)) is type `TreasuryAssetsFungibleFiatAmounts`.
2. The return value from `useTreasuryAssetsFungiblePrices` ([link](https://github.com/decent-dao/fractal-interface/pull/492/files#diff-f2405e6904887508084f19625c04d4635e3e5a9f60c7ec3aee74dc74b988d95fR21)) is type `TreasuryAssestFungiblePrices`.

Even though these types are identical, the utility here is that one can see that the hook's name matches the type of its return value.

adamgall (changes requested):
There are various linter warnings and errors, which can be found by running `npm run lint`.

(Currently, linting does not work due to a circular dependency in `src/providers/treasury/types/treasury.ts`. Comment the line out to successfully lint. An issue has been created for this circular dependency in #494).

Please fix the linter warnings and errors until there are none left 🙏

src/providers/treasury/hooks/useTreasuryAssetsFungiblePrices.ts
```diff
@@ -0,0 +1,125 @@
...
+    return data;
+  }, [erc20TokensAddresses, selectedCurrency]);
+
+  useQuery(queryKeyErc20Tokens, fetchPricesErc20Tokens, queryConfig);
+
+  useEffect(() => {
+    const data = queryClient.getQueryData(queryKeyErc20Tokens);
+
+    if (data) {
+      formatPricesData(data);
```
adamgall, line 118: Same comment as previous regarding type issues.

src/providers/treasury/hooks/useTreasuryAssetsFungiblePrices.ts
```diff
@@ -0,0 +1,125 @@
...
+    const { data } = await axios.get<CoinGeckoApiResponse>(url);
+
+    // this coingecko endpoint returns an object with a
+    // key named "ethereum" like so: `{ ethereum: 1000 }`.
+    // however, the value of this object is displayed in
+    // the UI based on the key matching `contractAddress`.
+    // here, the value is deconstructed, and then
+    // reconstructed with the appropriate address.
+    const [value] = Object.values(data);
+    return { [constants.AddressZero]: value };
```
adamgall, line 74: What happens when the coin gecko API is down, or returns an error, or returns data in an unexpected format?

src/providers/treasury/helpers/api.ts
```diff
@@ -0,0 +1,40 @@
+/**
```
herbig, line 1: One thing they mention in the CoinGecko docs is that attribution is required for their free plan.  Is that something we've already discussed with design?

https://www.coingecko.com/en/api/pricing

src/providers/treasury/helpers/api.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param currencies string or array of currencies
+ * @url https://api.coingecko.com/api/v3/simple/price
+ * @returns
+ */
+export const buildCoinGeckoApiUrlForNativeToken = (currencies: string | string[]) => {
+  const COIN_GECKO_URL = 'https://api.coingecko.com/api/v3/simple/price';
+  const vsCurrencies = typeof currencies === 'string' ? currencies : currencies.join(',');
+
+  const query = new URLSearchParams({
+    ids: 'ethereum',
```
herbig, line 12: if this assumes `etheruem`, how do we handle displaying price on e.g. Polygon/Matic, or is that not something we're currently worried about?

src/providers/treasury/hooks/useTreasuryAssetsFungible.ts
```diff
@@ -114,20 +108,27 @@ const useTreasuryAssets = (
         const name = await token.tokenData.name();
         const symbol = await token.tokenData.symbol();
         const decimals = await token.tokenData.decimals();
-        const tokenAddress = token.tokenData.tokenAddress;
+        // const tokenAddress = token.tokenData.tokenAddress;
+
+        // TODO: remove this bit of logic before pushing to production
```
herbig, line 113: I remember we discussed this during live review, but don't recall what was decided for this? This is just to spoof mainnet prices on testnet right?

src/providers/treasury/hooks/useTreasuryAssetsFungibleFiatAmounts.ts
```diff
@@ -0,0 +1,70 @@
...
+) => {
+  const [amounts, setAmounts] = useState<TreasuryAssetsFungibleFiatAmounts | {}>({});
+
+  const calculateFiatAmount = useCallback(
+    (price: TreasuryAssetFungiblePrice, asset: TreasuryAssetFungible) => {
+      // this Number will usually come in as a float
+      const { amount: priceAmount } = price[selectedCurrency];
+
+      // BigNumber does not support floats, so a
+      // multiplier is computed to perform the calculation.
```
herbig, line 32: Maybe we could add this conversion to some utility class if we have one? Seems like it would be used again sometime, not required but just a thought

herbig (commented):
A few comments, mostly for my own understanding!

src/providers/treasury/helpers/api.ts
adamgall: that's a good question. our multi-chain strategy will definitely need to account for this.

src/providers/treasury/helpers/api.ts
adamgall: i did not realize this! thanks for pointing it out.

the fractal webapp will need to provide this attribution somewhere (we also are supposed to have one for netlify...)

@herbig can you please create an issue so we can track our need for this page (or, a footer)

src/providers/treasury/TreasuryModuleProvider.tsx
```diff
@@ -22,24 +25,54 @@ export function TreasuryModuleProvider({
...
     nativeDeposits,
     nativeWithdraws,
     erc20TokenDeposits,
     erc20TokenWithdraws
   );
 
+  // TODO: where should this live?
+  // it might be useful to present a "currencies"
+  // option to select a currency in a future iteration.
+  const selectedCurrency = 'usd';
```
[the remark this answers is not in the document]
adamgall: i like how https://beaconcha.in/ does it.

there's a "currency selector" dropdown in their header, and switching it will change all "value display" units across the whole app into the selected currency.

cc: @xraystyle1980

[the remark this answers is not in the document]
xraystyle1980: To be clear, `Currency Selector` is a feature we need to see a design for, correct? And it sounds like `Currency Selector` could be used across the app, correct?

src/providers/treasury/helpers/api.ts
xraystyle1980: Roger that. We can add some footer type in... or can attribution happen in comments in the code?

herbig: @xraystyle1980 I believe it's required visually in the app, but doesn't need to be too large.

Issue: https://github.com/decent-dao/fractal-interface/issues/500

adamgall:
@brianfryer any updates on this PR?

herbig:
> Looks good! My only thoughts are the verbiage inside the tooltip. Can we do something like the attached screenshot? (whichever is preferred)
> 
> CleanShot 2022-09-13 at 11 27 53@2x

I like the first option makes the most sense, since it explicitly says _equals_. The second one is potentially confusing to me that it's saying you have 1 eth.

Also @brianfryer is there a DAO address you already have loaded with these tokens I can look at without having to send tokens to a DAO to see the tooltips?

adamgall:
@herbig goerli http://localhost:3000/#/daos/0xB4ea9Ad14195ebd60F1d7F3D8aEf0c4F1bC4BAa3/modules/0x6B185de798A0F34013E731D6B688E1A904B4C4fc

src/providers/treasury/hooks/useTreasuryAssetsFungiblePrices.ts
```diff
@@ -0,0 +1,125 @@
...
+  }, [hasNativeDeposits, selectedCurrency]);
+
+  useQuery(queryKeyNativeToken, fetchPriceNativeToken, queryConfig);
+
+  useEffect(() => {
+    const data = queryClient.getQueryData(queryKeyNativeToken);
+
+    if (data) {
+      formatPricesData(data);
```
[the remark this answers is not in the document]
brianfryer: Good catch! I'm curious how you got this error to show up (it wasn't popping up for me locally).

src/providers/treasury/hooks/useTreasuryAssetsFungiblePrices.ts
brianfryer: > What happens when the coin gecko API is down, or returns an error

Fortunately, since we're using `axios`, any errors (including timeouts) simply prevent the promise from being resolved—thus ensuring the code in the rest of the function doesn't run.

Basically, the current implementation makes the following assumptions:
* If price is available, show it
* If there is no price, don't show (for any reason, including errors)
* We don't make visible any errors that may arise as it's not useful to the end-user.

Perhaps when we implement error logging, we can do something like this:

```js
const { data } =
  await axios
    .get<CoinGeckoApiResponse>(URL)
    .catch((error) => /* do something with `error` */);
```

> or returns data in an unexpected format?

An update to gracefully exit early for incorrectly formatted API responses is forthcoming...

src/utils/countDecimals.ts
```diff
@@ -0,0 +1,9 @@
+const countDecimals = (value: number) => {
+  const isWholeNumber = value % 1 === 0;
+  if (isWholeNumber) return 0;
+
+  const count = value.toString().split('.').pop();
```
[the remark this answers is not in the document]
brianfryer: <3

src/providers/treasury/TreasuryModuleProvider.tsx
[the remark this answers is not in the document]
brianfryer: Yeah, we'll need a design for a `Currency Selector`.

I'll leave this as is for the time being.

src/providers/treasury/helpers/api.ts
brianfryer: 🙌

src/providers/treasury/helpers/api.ts
brianfryer: A very good question. The answer to which will likely need to include updates to `useTreasuryAssetsFungible.ts` too:

https://github.com/decent-dao/fractal-interface/blob/b49ee13c895c37a672a65cf4da53bd40d67af713/src/providers/treasury/hooks/useTreasuryAssetsFungible.ts#L57-L64

As you can see ^ here, all native tokens are set to `'Ethereum'`.

@adamgall @herbig — is our "multi-chain strategy" something we're actively working on right now?

src/providers/treasury/hooks/useTreasuryAssetsFungible.ts
brianfryer: Correct!

src/providers/treasury/hooks/useTreasuryAssetsFungibleFiatAmounts.ts
brianfryer: Good idea! Let's revisit this when it comes up again 💪

adamgall (approved):
Thanks @brianfryer , looks great!

src/providers/treasury/helpers/api.ts
herbig: I'm not sure what @adamgall has in mind but I think we might want to keep it chain agnostic in general, unless that adds a ton of complexity

herbig:
Hey @brianfryer code and attribution looks good, I was just wondering, how long should this page take to update?

I sent wrapped Eth to the this treasury and it's been about 20 minutes and it doesn't appear yet (it should have both eth and wrapped eth in different amounts):

https://deploy-preview-492--fractal-framework-interface-dev.netlify.app/#/daos/0x68dB43B90E6284019c0513168b514d103F4e1092/modules/0xe438C694ac785d912101e515498f9cD98Dc857f7

https://goerli.etherscan.io/address/0xe438C694ac785d912101e515498f9cD98Dc857f7

src/providers/treasury/helpers/api.ts
brianfryer: @herbig love that.

Correct me if I'm wrong, but it sounds like this might be something that should be part of a separate PR. If so, may we leave this PR as-is?

brianfryer:
@herbig I see a fiat amount for the ETH (see screenshot below). Maybe Goerli was experiencing some sort of slowdown?

image

src/providers/treasury/helpers/api.ts
herbig: yeah, definitely a larger refactor I suppose, I don't think we're handling multi chain well in general

adamgall:
@herbig this behavior is expected.

looking at the transaction which resulted in WETH being sent to the Treasury contract

https://goerli.etherscan.io/tx/0x21a097404f81192a08c717ca82cebcb4d23bf8c0650bef50f20477900797625f

we can see that it was a direct `transfer` call to the ERC20 contract itself.

When ERC20 tokens are transferred in this way, the destination contract (in our case, the Treasury contract) has no idea about that transfer. It doesn't know that it owns those WETH tokens.

The way the UI works, is that it is looking for events that were emitted from the Treasury contract, in order to build its own understanding of its token balances. Tokens must be sent to the Treasury contract "through" the Treasury contract's `[depositERC20Tokens](https://github.com/decent-dao/fractal-module-treasury/blob/main/contracts/TreasuryModule.sol#L56)` function, which emits an Event that the frontend listens for.

This workaround is strictly because our architecture is just stateless frontend app, and smart contracts. We _could_ hardcode some list of like, top 100 tokens, into the frontend app and "check" each Treasury for the existence of those tokens, but we don't do that currently.

Does this make sense?

herbig:
> @adamgall Does this make sense?

Ahhhhhh yeah totally. I forget we're not using straightforward token balance APIs and can't listen to any arbitrary token.

Yeah love the stateless architecture though, just these nuances that I'm catching up on.

I wish ERC20 had something like ERC721Receiver has with NFTs.

herbig (approved):
Looks good!

adamgall:
Thankfully, as we kill our Treasury contract and go pure Gnosis, we _will_ be able to use the Gnosis APIs to get that balance information.

src/providers/treasury/hooks/useTreasuryAssetsFungible.ts
Da-Colon, line 113: Looks like the conditional below is good to go do we need this TODO and commented out line here anymore

Da-Colon (approved):
LGTM pending a single comment

adamgall:
Gonna merge this :)

src/providers/treasury/hooks/useTreasuryAssetsFungible.ts
adamgall: I have done this

Da-Colon:
do it 🚀

adamgall:
actually I'm doing one last thing first

Issue #449, opened 2022-08-23
Usability | UI - Null token allocation address field should disable button selection


SudoPumpkin:
If the user leaves the Token Allocation address field blank the deploy button at the end is still selectable. 

_STR_
1. Log on to https://app.dev.fractalframework.xyz/#/daos/new
2. Input a Fractal Name
3. Click 'Next'
4. Select the default Governance token of '1:1 Token Voting'
5. Click 'Next'
6. Fill out all fields on this page except 'Address'
7. Click 'Next' (_Suggestion - make the address field highlighted with a red boarder here and not allow 'Next' button selection_)
8. Notice how the 'Deploy' button allows selection on this page

![image](https://user-images.githubusercontent.com/86701852/186246828-7338f2ab-389f-4cba-a708-5baa32201916.png)

herbig:
I'm going to take a look at improving input validation on this form, as it seems there's a few issues I've seen!

herbig:
I don't have admin access but if someone could assign me please do so

PR #501, opened 2022-09-15
Governance Token form validation improvements

herbig:
## Description

Fixes a number of issues on the _Treasury and Governance Token_ page related for form validation:

- ~~implements a `maxLength` parameter on `Input`, to constrain Token Symbol to 5 characters~~ (reverted after discussion)
- fixes misalignment on the two paired fields in `TokenAllocationInput` by making both fields "errored" if one of them is.
- fixes valid wallet addresses being marked as invalid if your wallet isn't yet connected
- adds additional checks to disable Next if the user has not input a `tokenName` or `tokenSymbol`
- fix https://github.com/decent-dao/fractal-interface/issues/449 using @SudoPumpkin suggestion to disable Next if allocation addresses are missing

## Notes

Each change was discretely committed, so it makes sense to review each commit individually, currently.

## Issue 

closes https://github.com/decent-dao/fractal-interface/issues/449, and fixes other issues we did not have tickets for (to my knowledge)

## Testing

Navigate from new Fractal page -> Create a Fractal -> _enter name_ -> _1:1 Token Voting_

All visual changes appear on this page.

- [ ] If one or the other fields on a `TokenAllocationInput` component has an error, but appear with red circles, although onl one should show error text
- [ ] valid wallet addresses (eg 0x1234..) no longer appear invalid when your wallet is not yet connected
- [ ] _Next_ is now disabled if Token Name or Token Symbol are not provided
- [ ] _Next_ is now disabled if any wallet address is provided for a `TokenAllocationInput`
- [ ] Please spot check other input fields in other sections of the app to ensure no unintended effects

Screen Shot 2022-09-15 at 2 27 25 PM


src/components/ui/forms/Input.tsx
```diff
@@ -162,6 +167,21 @@ function Input({
...
 
+  const handleChange = (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+
+    // handle any externally set change listener
+    if (onChange) onChange(event);
+
+    // handle max length (typing or pasting)
+    const newValue = event.currentTarget.value;
+    if (maxLength && newValue.length > maxLength) {
+      const setter = Object.getOwnPropertyDescriptor(window.HTMLInputElement.prototype, 'value')?.set;
```
herbig, line 178: This sets the current input value programmatically, but I feel like there has to be a cleaner way to do this in React?

src/components/ui/forms/Input.tsx
```diff
@@ -53,14 +55,17 @@ function Input({
   onChange,
   onClickMax,
   decimals,
+  maxLength,
 }: InputProps) {
+
+  const id = useId();
```
herbig, line 61: this is a React method I found which creates stable ids, which I think is probably just an incrementing counter or something.  But essentially in order to get the current component's id without grabbing other inputs above it, each input needed a unique id other than "form-field"

src/components/ui/forms/Input.tsx
```diff
@@ -190,7 +210,7 @@ function Input({
           min={min}
           max={max}
           onKeyDown={handleKeyDown}
-          onChange={onChange}
+          onChange={handleChange}
```
herbig, line 213: internal `handleChange` swapped for the externally provided `onChange`, and that's called there instead

src/types/tokenAllocation.ts
```diff
@@ -2,6 +2,7 @@ import { BigNumberInput } from './../components/DaoCreator/provider/types/index'
 
 export type TokenAllocation = {
   address: string;
+  isValidAddress: boolean;
```
herbig, line 5: this gets propagated to [useNextDisabled.ts](https://github.com/decent-dao/fractal-interface/pull/501/files#diff-e06cb30686b344ca48a605a57304e5f5d6c65ed902efe4d96f5284c5f7b0997c), in order to enable/disable the Next button based on whether al the allocation fields have valid addresses.

I made this required rather than optional, since I didn't want future uses to be confused as to why it's there, if you use `TokenAllocation` you should know whether you're using a valid address

src/components/DaoCreator/tokenVotingGovernance/TokenDetails.tsx
```diff
@@ -64,7 +64,6 @@ function TokenDetails() {
           onChange={e => onSupplyChange(e.target.value)}
           label="Token Supply"
           helperText="Max: 18 decimals"
-          disabled={false}
```
herbig, line 67: these are default enabled, so these lines were superfluous

src/components/DaoCreator/tokenVotingGovernance/TokenAllocationInput.tsx
```diff
@@ -52,14 +62,14 @@ function TokenAllocationInput({
         type="text"
         value={tokenAllocation.address}
         onChange={event => updateAddress(event.target.value)}
-        errorMessage={tokenAllocation.addressError}
+        errorMessage={tokenAllocation.addressError ? tokenAllocation.addressError : hasAmountError ? '‎' : undefined}
```
herbig, line 65: you can't see it rendered on Github, but this `''` is actually an empty box `U+200e` character, which serves to just take space so that the error boxes aren't misaligned

src/components/DaoCreator/tokenVotingGovernance/TokenAllocationInput.tsx
```diff
@@ -30,16 +30,26 @@ function TokenAllocationInput({
     if (address.trim()) {
       isValidAddress = await checkAddress(provider, address);
     }
+    let errorMsg = undefined;
+    if (!isValidAddress) {
+      if (!provider && address.includes('.')) { // simple check if this might be an ENS address
```
herbig, line 35: I thought about using a regex here, but ENS names aren't as simple as "xxx.eth", since they can be any domain, so just simply checking if they might be trying to input an ENS before they've connected should suffice.

Side note, our field currently says ENS names with emojis are invalid, which should not be the case, although I didn't investigate that in this PR.

herbig:
> ### ❌ Deploy Preview for _fractal-framework-interface-dev_ failed.
> Name 	Link
> 🔨 Latest commit 	[671c1be](https://github.com/decent-dao/fractal-interface/commit/671c1be6d4f333c99bddfe81ee2212601a89b1ab)
> 🔍 Latest deploy log 	https://app.netlify.com/sites/fractal-framework-interface-dev/deploys/6323766fe0020d0008a38b1a

@adamgall it's not clear to me why these are failing exactly, it suggests some formatting improvements or something?

adamgall:
> > ### ❌ Deploy Preview for _fractal-framework-interface-dev_ failed.
> > Name 	Link
> > 🔨 Latest commit 	[671c1be](https://github.com/decent-dao/fractal-interface/commit/671c1be6d4f333c99bddfe81ee2212601a89b1ab)
> > 🔍 Latest deploy log 	https://app.netlify.com/sites/fractal-framework-interface-dev/deploys/6323766fe0020d0008a38b1a
> 
> @adamgall it's not clear to me why these are failing exactly, it suggests some formatting improvements or something?

Oh yeah buddy, seems like lots of linter errors.

Da-Colon (approved):
Nicely done, changes to the code make sense 👍. Thanks for identifying and covering these other issues as well

src/components/DaoCreator/tokenVotingGovernance/TokenAllocationInput.tsx
adamgall: Is it possible to use the hook at `src/hooks/useAddress.ts` here? It takes an input string and returns back data regarding if the address is valid or not, and has ENS support built in.

src/components/ui/forms/Input.tsx
```diff
@@ -162,6 +165,23 @@ function Input({
...
+    const newValue = event.currentTarget.value;
+    if (maxLength && newValue.length > maxLength) {
+      const setter = Object.getOwnPropertyDescriptor(
+        window.HTMLInputElement.prototype,
+        'value'
+      )?.set;
+      const input = document.getElementById(id);
+      setter?.call(input, newValue.substring(0, maxLength));
+      input?.dispatchEvent(new Event('input', { bubbles: true }));
+    }
```
adamgall, line 182: This seems... idk, not very React-y to me. What is the full scope of the intention of this code?

adamgall (approved):
Left some comments, but also giving this an approval.

src/components/DaoCreator/tokenVotingGovernance/TokenAllocationInput.tsx
```diff
@@ -30,16 +30,27 @@ function TokenAllocationInput({
...
     }
+    let errorMsg = undefined;
+    if (!isValidAddress) {
+      if (!provider && address.includes('.')) {
+        // simple check if this might be an ENS address
+        errorMsg = 'Connect wallet to validate';
+      } else if (address.trim()) {
+        errorMsg = 'Invalid address';
+      }
+    }
```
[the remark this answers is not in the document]
herbig: `checkAddress` will return false for an ENS address if a wallet isn't connected, because you need to be connected to check the ENS name, so it could be valid in that case

We would need to require the user be connected in order to actually validate this field, the intention here is to give them a better message that says to connect, rather than saying "example.eth" isn't valid, which isn't true.

src/components/ui/forms/Input.tsx
herbig: Yeah, I had this comment before but it got outdated: https://github.com/decent-dao/fractal-interface/pull/501#discussion_r972319207

The intention is to take what is put into the field and cut it down to the maxLength, then trigger an input event on the field, so it will revalidate.

herbig: @adamgall the latest commit makes more sense I think

src/hooks/useAddress.ts
```diff
@@ -63,21 +63,21 @@ const useAddress = (addressInput: string | undefined) => {
...
-  if (!provider || addressInput.trim() === '') {
+  if (addressInput === undefined || addressInput.trim() === '') {
     return false;
   }
 
   if (addressInput === constants.AddressZero) {
     return false;
   }
 
+  // check if it's a standard wallet address (e.g. 0xasdf)
```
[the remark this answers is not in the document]
herbig: That's fair, though mainly I had this one because of the comment about ENS below, which is a bit less obvious what's happening.

src/hooks/useAddress.ts
```diff
@@ -63,21 +63,21 @@ const useAddress = (addressInput: string | undefined) => {
 };
 
 export const checkAddress = async (provider: any, addressInput?: string): Promise<boolean> => {
-  if (addressInput === undefined) {
-    return false;
-  }
-
-  if (!provider || addressInput.trim() === '') {
+  if (addressInput === undefined || addressInput.trim() === '') {
```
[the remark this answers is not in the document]
herbig: ah yeah I was just moving the above check here instead, without changing the logic

src/components/ui/forms/Input.tsx
```diff
@@ -53,7 +54,9 @@ function Input({
   onChange,
   onClickMax,
   decimals,
+  maxLength,
 }: InputProps) {
+  const [valInternal, setValInternal] = useState(value);
```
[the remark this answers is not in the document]
herbig: ah, right, I was thinking value _was_ the default value, rather than considering that you would pass a stored value into here

Hmm, that does complicate things.  The parent component level validation makes more sense to me, but is way less clean...

I need to rethink this

herbig:
>I'm all good to transitioning to `yarn`, btw. But I think it would also make sense to do that in separate PR

Ah yeah that's my mistake and not being familiar with this FE dev, I didn't even realize that was there 🤦

src/components/ui/forms/Input.tsx
herbig: Scratch that, getting feedback from @mudrila so disregard..

src/components/ui/forms/Input.tsx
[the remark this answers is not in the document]
herbig: ah, yeah that makes sense, I appreciate that.

I'm actually going to implement this locally and see how it works, but revert the whole `maxWidth` from this PR, and we can discuss form libraries as a better solution later on (though not a priority).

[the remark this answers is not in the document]
Da-Colon: I want to toss in I'm fond of the Formik (form library) and Yup (validation library).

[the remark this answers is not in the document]
herbig: I've updated the PR and removed this altogther.. thanks!

src/components/DaoCreator/tokenVotingGovernance/TokenAllocationInput.tsx
```diff
@@ -52,14 +63,26 @@ function TokenAllocationInput({
         type="text"
         value={tokenAllocation.address}
         onChange={event => updateAddress(event.target.value)}
-        errorMessage={tokenAllocation.addressError}
+        errorMessage={
+          tokenAllocation.addressError
+            ? tokenAllocation.addressError
+            : hasAmountError
+            ? '‎'
```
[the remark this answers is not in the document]
herbig: ah yeah that's actually not rendered properly on Github, but it's an empty space character, which serves to make both fields errored if one of them is, in order to align them both

Screen Shot 2022-09-21 at 6 30 25 PM

Before:
Screen Shot 2022-09-21 at 6 28 54 PM

After:
Screen Shot 2022-09-21 at 6 29 27 PM

Da-Colon (approved):
The code looks good to me 👍

src/components/DaoCreator/tokenVotingGovernance/TokenAllocationInput.tsx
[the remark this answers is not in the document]
Da-Colon: Good catch. I would think `'  '` works too but having the character code works too as it shows an intentional space is added. Though probably want to do a browser check to make sure this behavior is repeatable in other browsers.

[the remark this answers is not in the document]
herbig: yeah good call.  I had tried `' '` but it didn't work, which is why I added the unicode instead.  Likely because it's trimmed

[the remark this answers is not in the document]
Da-Colon: o doi yeah that makes sense.

PR #503, opened 2022-09-16
Web3 Wallet Provider Update: Connect to local node as signer (test environment)

Da-Colon:
## Description

Update to allow for test environment to connect to 'local' blockchain node as Signer

## Notes

These updates have been made to [`wallet-provider` ](https://github.com/decent-dao/wallet-provider). 

## Testing

On the preview site everything should continue working as usual.

See comment in code for how to test locally.

n/a


.env
```diff
@@ -1,5 +1,8 @@
-REACT_APP_LOCAL_CHAIN_ID="31337"
-REACT_APP_LOCAL_PROVIDER_URL="http://127.0.0.1:8545"
+# HARDHAT DEFAULT PROVIDER CHAIN_ID: "31337"
+REACT_APP_LOCAL_CHAIN_ID=""
+
+# HARDHAT DEFAULT PROVIDER URL: "http://127.0.0.1:8545"
+REACT_APP_LOCAL_PROVIDER_URL=""
```
Da-Colon, line 6: Unset these in the `.env` so these should be set by in a `local.env` file

src/contexts/web3Data/Web3Provider.tsx
```diff
@@ -70,17 +72,20 @@ export function Web3Provider({ children }: { children: React.ReactNode }) {
...
-      dispatch({
-        type: Web3ProviderActions.SET_LOCAL_PROVIDER,
-        payload: await getLocalProvider(),
-      });
-    } else {
-      dispatch({
-        type: Web3ProviderActions.SET_FALLBACK_PROVIDER,
-        payload: getFallbackProvider(),
-      });
+    if (process.env.REACT_APP_LOCAL_PROVIDER_URL && process.env.NODE_ENV !== 'production') {
```
Da-Colon, line 75: This allow for a `read-only` fallback provider to still connect if the `REACT_APP_LOCAL_PROVIDER_URL` is set

src/contexts/web3Data/helpers.ts
```diff
@@ -35,24 +40,33 @@ export const getInjectedProvider = (
...
-  });
+
+  try {
+    const network = await localProvider.detectNetwork();
+
+    // sets the account and signer automatically in a test environment.
+    const signerOrProvider = isTestEnvironment ? localProvider.getSigner() : localProvider;
+    const account = isTestEnvironment
+      ? await (signerOrProvider as providers.JsonRpcSigner).getAddress()
+      : null;
```
Da-Colon, line 57: Set up this way to only connect as a signer via the local node in a test environment. otherwise this returns a read-only provider

Da-Colon (commented):
Noting here: There are several optimizations that have been made in the `@decent-org/wallet-provider` to where the setup is a little different but I was able adapt those changes for this version. In short order I'll work on getting the library installed and the front-end updated after of course some testing and eyes on that repo

src/contexts/web3Data/helpers.ts
```diff
@@ -35,24 +40,33 @@ export const getInjectedProvider = (
   });
 };
 
-export const getLocalProvider = (): Promise<BaseProviderInfo> => {
+export const getLocalProvider = async (): Promise<LocalInjectedProviderInfo | undefined> => {
+  const isTestEnvironment = process.env.NODE_ENV === 'test';
```
Da-Colon, line 44: To Test Locally

- Start a local blockchain node
- (if you don't already) add a `env.local` and set the `REACT_APP_LOCAL_PROVIDER_URL` and `REACT_APP_LOCAL_CHAIN_ID` to connect to local your local node (defaults are commented in `.env`
- Set `isTestEnvironment` to true in the local codebase
- start development server


There is some errors that appear (localaddress are not set) as you probably won't have all the contract deployed to that local node. but the first account of that locally ran node should connect and appear in the top right

.env
adamgall: Just curious, why did you unset these in `.env`?

Da-Colon: short answer: so that this code doesn't try to execute by default. feel like these should be intially set.

Longer Thought:

This part of the code feels eh to me. I need to spend a little more time here. this is technically a fallback provider. I'm going to switch this PR to a draft temporary. Currently in a test environment this will automatically connect to the local node in the test env and the url is provider. 

What needs to be done:
- Add a local provider connection when 'connect wallet' is clicked
- remove the auto connection on test environment
- revert change on env (your probably right here that this change is unneeded since these are the defaults anyway

Da-Colon: I'll have this ready by in the morning.

src/contexts/web3Data/helpers.ts
adamgall: Could we simplify this and always just assume that if we're connected to a local node, we'll have a signer account on that local node? I don't see why not.

.env
adamgall: Sorry, I deleted my question mere seconds before you posted your answer.

Original question was: "why did you unset these values in `.env`?"

src/contexts/web3Data/helpers.ts
Da-Colon: yeah I thought about asking that question too. do we need the ability to have a local fallback provider. (this would let you connect via metamask as well?). I could just set this up for if it's a test env expect and connect to a local node

adamgall: Hm, yeah there's an order of operations thing going on here. If someone wanted to connect to a local node via Metamask, that connection would happen entirely within Metamask (add your local node as a network in MM). But, if that local node is running, in what order does the webapp try to connect to it? I think that happens _before_ checking for an Injected Provider. And so if we're connected to the Local node directly, and it has an unlocked Signer that we use, then the UI probably won't allow us to override that signer and jump into the Injected Provider.

Ha, what a shitshow. This could use a flowchart lol

Da-Colon: I went with a modal that allows you to choose from the wallet menu. This option will only show when the local provider url is provided as a `env`

image

This allows for the tests to choose this option and connect as a signer to the local node.

The image there is a filler. since it's dev only it doesn't REALLY matter what image goes there. I was thinking of just putting the Fractal F there.

.env
```diff
@@ -1,20 +1,5 @@
-REACT_APP_LOCAL_CHAIN_ID="31337"
-REACT_APP_LOCAL_PROVIDER_URL="http://127.0.0.1:8545"
-
```
Da-Colon, line 3: The changes in this file is to merge these to different patterns into one. So now the local address is set just like any other addresses. currently the defaults for `31337` are set to goerli for our specific use within docker ran node.

src/contexts/web3Data/Web3Provider.tsx
```diff
@@ -70,22 +72,26 @@ export function Web3Provider({ children }: { children: React.ReactNode }) {
...
+      const localProviderInfo = await getLocalProvider();
+      if (!!localProviderInfo) {
+        dispatch({
+          type: Web3ProviderActions.SET_LOCAL_PROVIDER,
+          payload: localProviderInfo,
+        });
+        return;
+      }
+      web3Modal.clearCachedProvider();
     }
```
Da-Colon, line 85: Fallback provider fallback to local provider if url is set

src/contexts/web3Data/web3Modal.config.ts
```diff
@@ -9,6 +11,24 @@ export const WEB3_MODAL_CONFIG = {
         infuraId: process.env.REACT_APP_INFURA_API_KEY,
       },
     },
+    'custom-localprovider': {
+      display: {
+        logo: glow,
+        name: 'Local Node',
+        description: 'Connects as Signer to local provider',
+      },
+      package: !!process.env.REACT_APP_LOCAL_PROVIDER_URL,
```
Da-Colon, line 20: This determines whether the option is shown or not

src/contexts/web3Data/web3Modal.config.ts
```diff
@@ -9,6 +11,24 @@ export const WEB3_MODAL_CONFIG = {
...
+      connector: async () => {
+        const localProvider = new ethers.providers.JsonRpcProvider(
+          process.env.REACT_APP_LOCAL_PROVIDER_URL
+        );
+        const network = await localProvider.detectNetwork().catch();
+        if (!network) {
+          return { chainId: '' };
+        }
```
Da-Colon, line 28: I'm returning an empty string here for the sake of the web3modal 'connect' listener in `useListeners` hook. It expects a chainId property. This still fails the conditonal statement and will throw a toast for invalid network.

herbig: That happens only if REACT_APP_LOCAL_PROVIDER_URL is set, and there's a problem connecting to a network right?

Would it help to provide a different error, or console log so devs can pinpoint what might be happening a lot quicker than only seeing the invalid network toast?

src/contexts/web3Data/Web3Provider.tsx
```diff
@@ -70,22 +72,26 @@ export function Web3Provider({ children }: { children: React.ReactNode }) {
...
-      dispatch({
-        type: Web3ProviderActions.SET_LOCAL_PROVIDER,
-        payload: await getLocalProvider(),
-      });
-    } else {
-      dispatch({
-        type: Web3ProviderActions.SET_FALLBACK_PROVIDER,
-        payload: getFallbackProvider(),
-      });
+    if (process.env.REACT_APP_LOCAL_PROVIDER_URL && process.env.NODE_ENV !== 'production') {
```
herbig, line 75: we don't need to change this for the PR, but just for super optimization sake it's useful to keep in mind that ordering production / development checks makes sense to put it first, so the second condition never needs to be checked

src/contexts/web3Data/web3Modal.config.ts
herbig: cool cool, makes sense

herbig (approved):
Awesome, looks good to go!

herbig (approved):
🚢🚢🚢

PR #511, opened 2022-09-22
Playwright setup cleanup

adamgall:
## Description

Slight updates to the Playwright config setup including:

- adding `playwright` as a dev dependency, which makes it so that we don't have to separately do `npx playwright install` after `npm install` to get the correct browsers installed.
- updates the run script to target both chromium and firefox, and doesn't explicitly say to use the default config file (it just uses it by default)

## Notes

n/a

## Issue (if applicable)

n/a

## Testing

n/a

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     ]
   },
   "devDependencies": {
-    "@playwright/test": "^1.25.2",
+    "@playwright/test": "^1.26.0",
     "@typescript-eslint/eslint-plugin": "^5.34.0",
     "@typescript-eslint/parser": "^5.34.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -58,7 +58,7 @@
     ]
   },
   "devDependencies": {
-    "@playwright/test": "^1.25.2",
+    "@playwright/test": "^1.26.0",
```
herbig, line 61: Pump em up

herbig (approved):

SudoPumpkin (approved):
The minor changes in this PR look fine to me. 👍

PR #517, opened 2022-09-28
Replace typing on Treasury Module Event Filtering

adamgall:
## Description

Remove `getPastEvents` function in `useTreasuryEvents`, which loses typing on event filtering. Adds typing back into code that gets data from those events.

## Notes

n/a

## Issue (if applicable)

n/a

## Testing

n/a

n/a

### DIFF
--- a/src/providers/treasury/hooks/useTreasuryEvents.ts
+++ b/src/providers/treasury/hooks/useTreasuryEvents.ts
@@ -1,5 +1,5 @@
 import { BigNumber } from 'ethers';
-import { useCallback, useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { TreasuryModule } from '../../../assets/typechain-types/module-treasury';
 import { TypedEvent } from '../../../assets/typechain-types/module-treasury/common';
 import {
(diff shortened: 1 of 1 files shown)


src/providers/treasury/hooks/useTreasuryEvents.ts
```diff
@@ -18,17 +18,6 @@ const useTreasuryEvents = (treasuryModuleContract?: TreasuryModule) => {
...
-   * retreives past events and sets it to state
-   *
-   * @param treasuryModuleContract
-   * @param filter contract query filter for specific event
-   */
-  const getPastEvents = useCallback(async (treasuryContract: TreasuryModule, filter: any) => {
-    const events = await treasuryContract.queryFilter(filter);
-    return events;
-  }, []);
-
```
adamgall, line 31: The offending function. Seems to be a result of over-abstraction, which had a negative consequence of losing types when doing event filtering.

src/providers/treasury/hooks/useTreasuryEvents.ts
```diff
@@ -209,8 +198,7 @@ const useTreasuryEvents = (treasuryModuleContract?: TreasuryModule) => {
 
     const getData = async () => {
       // capture past native deposit events
-      const events: TypedEvent<any, any>[] = await getPastEvents(
-        treasuryModuleContract,
```
adamgall, line 213: We actually save a line of code for each fix (resulting from the offending function being removed).

These two lines can be replace with...

src/providers/treasury/hooks/useTreasuryEvents.ts
```diff
@@ -209,8 +198,7 @@ const useTreasuryEvents = (treasuryModuleContract?: TreasuryModule) => {
...
+      const events = await treasuryModuleContract.queryFilter(
```
adamgall, line 201: ... this one line!

src/providers/treasury/hooks/useTreasuryEvents.ts
```diff
@@ -223,8 +211,8 @@ const useTreasuryEvents = (treasuryModuleContract?: TreasuryModule) => {
         events.map(async event => {
           const block = await event.getBlock();
           return {
-            address: event.args[0],
-            amount: event.args[1],
+            address: event.args.sender,
+            amount: event.args.amount,
```
adamgall, line 215: And now we've got type-checked human readable names for event parameters, as coming from typechain.

Da-Colon (approved):
good catch with this

PR #522, opened 2022-10-02
Explicit MVD state

adamgall:
## Description

First in a series of PRs, with the overall goal being to allow the webapp to load up either an "MVD"-style Node, or a "Gnosis Safe"-style Node.

## Notes

This First PR makes absolutely no functional changes, but it does 4 things (as indicated by the 4 commits):

- Fixes a typo in a README lol
- Removes an unused file
- Wraps the top level state returned from `useFractal` in an object called `mvd`, to make room for more pieces of top level state
- Renames some things up in that state management layer to differentiate between a "DAO" and an "MVD"-style DAO

In the next couple of PRs, reviewers will see the other "top level state" objects that are going to be added, namely:

- `node`, which will be very light, simply holding a property that describes what "kind" of DAO Node this is, either `MVD` or `GNOSIS`.
- `gnosis`, which will include state/information about the Gnosis Safe itself.

## Issue (if applicable)

n/a

## Testing

n/a, no functional changes here. Things should work exactly the same as before this PR.

n/a


herbig (approved):
looks pretty straightforward, no notes!

Da-Colon (approved):
Code changes lgtm. One question though. Is the ultimate goal still removing MVD, its contracts and its support from the codebase?

herbig:
> Code changes lgtm. One question though. Is the ultimate goal still removing MVD, its contracts and its support from the codebase?

Yeah, that's the ultimate endpoint, removing MVD altogether.  The thought is that adding Gnosis DAO support rather than refactoring the whole thing will make changes easier to review.

adamgall:
@Da-Colon Yes, ultimately.

But the approach I'm taking with these PRs is to "add Gnosis Safe support", rather than "replace MVD support with Gnosis support".

Reason is mostly so that code reviews can go smoother. Also kinda so that if adding Gnosis support takes longer than expected, the whole app isn't broken in the meantime.

PR #526, opened 2022-10-05
Gnosis Addresses

adamgall:
## Description

Use Gnosis Proxy Factory and Safe Singleton addresses from the official Gnosis `safe-deployments` NPM package, instead of being set manually by us in environment variables.

## Notes

After this PR is merged, @adamgall should delete these now-unused environment variables from Github Secrets and Netlify.

## Issue (if applicable)

n/a

## Testing

The easiest way for a reviewer to manually confirm that this code works, is to:

- Delete any record of `REACT_APP_GNOSISSAFE_ADDRESSES` and `REACT_APP_GNOSISSAFEFACTORY_ADDRESSES` environment variables from their `.env.local` files (if applicable).
  - But! Take note of what those addresses are before you delete them. Or, look at the change-set of this PR to see what those deleted addresses were from in `.env`.
-  Somewhere in the app, for example right at the top in `App.tsx`, import `useWeb3Provider` and `useAddresses`, call it at the top of the `App` component, console.log the output

```tsx
  const {
    state: { chainId },
  } = useWeb3Provider();
  console.log(useAddresses(chainId));
```

- Notice how the address records for `gnosisSafeFactory` and `gnosisSafe` (names unchanged -- so all downstream code works the same) include the same addresses that are no longer in environment variables.

n/a


src/hooks/useAddresses.ts
```diff
@@ -65,7 +78,10 @@ export function useAddresses(chainId: number | undefined) {
     const gnosisWrapperFactoryNetworksAddresses: { [chaindId: number]: { address: string } } =
       JSON.parse(process.env.REACT_APP_GNOSISWRAPPERFACTORY_ADDRESSES);
     const gnosisSafeFactoryNetworksAddresses: { [chaindId: number]: { address: string } } =
-      JSON.parse(process.env.REACT_APP_GNOSISSAFEFACTORY_ADDRESSES);
+      Object.keys(gnosisProxyFactoryDeployment.networkAddresses).reduce(
+        (p, c) => ({ ...p, [c]: { address: gnosisProxyFactoryDeployment.networkAddresses[c] } }),
+        {}
+      );
```
adamgall, line 84: This was fun.

The format of data in `gnosisProxyFactoryDeployment` was different than the format of data in the rest of our address data coming from environment variables.

In order to make the rest of this code stay unchanged, the only thing we had to do here was:

- Not feed the environment variable data into `JSON.parse` (because these environment variables are gone now)
- Instead, just make sure to get the data from the Gnosis function into the exact same format as what was previously being set into the `gnosisSafeFactoryNetworksAddresses` variable.

So that's what this snippet does.

herbig (approved):
🚢🚢🚢

PR #531, opened 2022-10-06
Language switcher component

herbig:
## Description

Adds a language switcher to the dropdown menu, allowing users to manually select their display language from supported languages.

## Notes

Only English / Spanish appear currently, and only the dropdown menu itself has translations.

Also, given that our UI is changing, I didn't fully flesh out a few UI issues with the switcher, this is mostly to have in place for a redesign / discussion.

## Issue (if applicable)

N/A

## Testing

You can switch between English / Spanish, and see that the menu language changes.

Screen Shot 2022-10-06 at 2 50 48 AM


src/i18n/LanguageSwitcher.tsx
```diff
@@ -0,0 +1,20 @@
+import { useTranslation } from 'react-i18next';
+import Translate from '../components/ui/svg/Translate';
+
+function LanguageSwitcher() {
+  const { i18n } = useTranslation();
+  return (
+    <div className="flex items-center justify-start gap-4 p-4 hover:bg-slate-200 hover:text-black">
+      <Translate />
+      <select
+        className="bg-gray-500 focus:outline-none text-white hover:bg-slate-200 hover:text-black"
```
herbig, line 10: I took styles from the menu items and existing switchers.  Ideally I would like the <selec /> hover style to match that of the <div /> parent, but I didn't want to spend too much time on it when these menu items are going away with the redesign, so there's a bit of visual weirdness on hover.

src/i18n/index.ts
```diff
@@ -58,9 +58,6 @@ i18n
   .use(initReactI18next)
   .init({
     resources: resources,
-    detection: {
-      caches: [], // don't cache detected browser language
```
herbig, line 62: without a switcher, we don't cache the selected language, that way every time you go to the app it checks the current browser language.

Caching it now means the first time they come to the app their browser language is cached as the current language, and they can then switch via the switcher instead.

src/components/ui/svg/Translate.tsx
```diff
@@ -0,0 +1,18 @@
...
+    >
+      <path
+        d="m11.9 22 4.55-12h2.1l4.55 12H21l-1.05-3.05H15.1L14 22Zm3.8-4.8h3.6l-1.75-4.95h-.1ZM4 19l-1.4-1.4 5.05-5.05q-.95-1.05-1.662-2.175Q5.275 9.25 4.75 8h2.1q.45.9.963 1.625.512.725 1.237 1.525 1.1-1.2 1.825-2.462Q11.6 7.425 12.1 6H1V4h7V2h2v2h7v2h-2.9q-.525 1.775-1.425 3.45-.9 1.675-2.225 3.15l2.4 2.45-.75 2.05L9 14Z"
+        fill="currentColor"
+      />
+    </svg>
+  );
+}
+
+export default Translate;
```
herbig, line 18: This icon just comes from Google's material design iconset.

src/i18n/LanguageSwitcher.tsx
```diff
@@ -0,0 +1,20 @@
...
+        value={i18n.language}
+        onChange={e => i18n.changeLanguage(e.target.value)}
+      >
+        <option value="en">English</option>
+        <option value="es">Español</option>
```
herbig, line 15: Should these strings be translated for every language as well??? I have no idea.

adamgall (approved):
LGTM, no comments, thanks @herbig!

src/i18n/LanguageSwitcher.tsx
herbig: The current one meaning the list of supported languages is static? Or should we translate it when you switch languages?

Now that I'm thinking about it we probably shouldn't translate them all each time, since then the alphabetical order would change.

src/i18n/LanguageSwitcher.tsx
Da-Colon: Yeah for sure. Probably need to ask design where this will go

src/i18n/LanguageSwitcher.tsx
Da-Colon: probably could sort them by international symbol 'en', 'esp' etc versus the translated name? (in the future)

Da-Colon (approved):
LGTM

PR #533, opened 2022-10-08
Component for token gating via POAP

herbig:
## Description

This is a simple drop-in component for potentially "token gating" features within the app behind a POAP NFT (https://poap.xyz/), similar to `ConnectWalletToast`

The idea is that we can have certain features on the development server that are available early for engaged community members, to increase community engagement.

Working with the community / marketing folks, we'll have our own POAP that can be awarded during community calls, on site conferences, etc.

The purpose of this PR is to open discussion about this idea.  If anyone has other comments, questions, or concerns with this approach, feel free to bring them up here!

## Notes

- Please test with a wallet that both has the POAP and does not have the POAP.
- The `Sign up` button will link to a typeform or Google form to allow users to input an email or wallet address to get into the list of folks we award the POAP.

## Issue (if applicable)

N/A

## Testing

The POAP used in this PR my own POAP. Single use mint links here:

- http://POAP.xyz/claim/3ym2a4
- http://POAP.xyz/claim/k1n0az
- http://POAP.xyz/claim/pdwgzb
- http://POAP.xyz/claim/0ofu9c
- http://POAP.xyz/claim/v67sx6

If we decide to go ahead with implementing this in some way, we will work with design to get a proper POAP image and copy.

Screen Shot 2022-10-10 at 4 48 06 PM


src/components/TokenGate.tsx
```diff
@@ -0,0 +1,112 @@
...
+    </div>
+  );
+}
+
+/**
+ * @param featureName a string describing the feature being gated
+ * @param alwaysRequireConnected whether this feature should continue to require a connected wallet, even on production
+ * @returns null the component can display a Toast notification, but doesn't itself render anything
+ */
+function TokenGate({
```
herbig, line 43: This is similar to our `ConnectWalletToast`, with the added feature that we can also gate certain features based on possession of a specific POAP

src/components/ConnectWalletToast.tsx
```diff
@@ -32,6 +32,7 @@ function ConnectWalletToast({ label }: { label: string }) {
       closeOnClick: false,
       draggable: false,
       progress: 1,
+      toastId: 'connectWalletToast:' + label,
```
herbig, line 35: fixes a bug with `ConnectWalletToast` that displays duplicate toasts

.env
```diff
@@ -7,7 +7,12 @@ REACT_APP_SUPPORTED_CHAIN_IDS="4,5,11155111"
 REACT_APP_INFURA_API_KEY=""
 REACT_APP_ALCHEMY_API_KEY=""
 REACT_APP_ETHERSCAN_API_KEY=""
+REACT_APP_POAP_API_KEY=""
```
herbig, line 10: I have a poap api key locally, but it doesn't actually seem to need one 🤷‍♂️

A successful request looks like: 
https://api.poap.tech/actions/scan/0x8E2695650D09FD940516d6e050D0Ba87d8deF032/62673

.env
```diff
@@ -7,7 +7,12 @@ REACT_APP_SUPPORTED_CHAIN_IDS="4,5,11155111"
...
 
+// TODO create a custom POAP to gate with
+REACT_APP_POAP_GATE_ID="62673"
+// TODO form link
+REACT_APP_POAP_GATE_SIGN_UP_URL="https://www.google.com"
```
herbig, line 15: These will be the actual Fractal early access POAP id, and a typeform to sign up for our mailing list

src/components/DaoCreator/provider/CreatorProvider.tsx
```diff
@@ -116,7 +116,6 @@ export function CreatorProvider({ deployDAO, pending, isSubDAO, children }: ICre
   return (
     <CreatorContext.Provider value={value}>
       <div className="pb-16">
-        <ConnectWalletToast label={t('labelConnectWallet')} />
```
herbig, line 119: removed for testing, but this will be brought back before merging, since we probably don't want to gate DAO creation

src/components/TokenGate.tsx
```diff
@@ -0,0 +1,113 @@
...
+        label={actionLabel}
+        onClick={action}
+      />
+    </div>
+  );
+}
+
+/**
+ * @param featureName a string describing the feature being gated
+ * @param alwaysRequireConnected whether this feature should continue to require a connected wallet, even on production
```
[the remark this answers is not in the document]
christopherdancy: do you mean if it requires "conected wallet" while on development?

src/components/TokenGate.tsx
```diff
@@ -0,0 +1,113 @@
...
+      message = t('tokenGateNoPOAPMessage', { feature: featureName });
+      actionLabel = t('tokenGateActionNoPOAP');
+      action = () => window.open(process.env.REACT_APP_POAP_GATE_SIGNUP_URL, '_blank');
+    } else {
+      // we are both connected and have the POAP (or are on production)
+      return;
+    }
+
+    // redirect out of here
+    navigate('/');
```
christopherdancy, line 95: Could we stop the page from loading here instead of doing a page reload? not sure what this would look like.

src/pages/DaoCreate/index.tsx
```diff
@@ -17,10 +18,17 @@ function DaoCreate() {
   };
 
   return (
-    <DaoCreator
-      pending={pending}
-      deployDAO={deployDAO}
-    />
+    <>
+      {/* TODO just gating this for testing, we don't necessarily want to gate DAO creation */}
+      <TokenGate
```
christopherdancy, line 23: It would also be nice to move this down to the component which will be using it - I.e moving it to the DAOCreator compoent

src/pages/DaoCreate/index.tsx
```diff
@@ -17,10 +18,17 @@ function DaoCreate() {
...
+        featureName="DAO creation"
+        alwaysRequireConnected={false}
+      />
+      <DaoCreator
```
christopherdancy, line 27: Here is an idea - perhaps the button should be disabled and if the user clicks on it - a toast or some helper text could pop to explain to the user what to do next.

src/components/TokenGate.tsx
herbig: Yeah not sure how that would work exactly, and might end up causing weird state issues.

I do think the redirect might be a little janky though.

src/pages/DaoCreate/index.tsx
herbig: yeah rather than a component which redirects, perhaps a component to enclose and block a feature you want to token gate, that then shows info on how to get access

src/pages/DaoCreate/index.tsx
herbig: true, it would probably work there as well

herbig:
Closing this PR and will discuss new features with product / design / marketing going forward.

PR #536, opened 2022-10-10
Playwright tests - Parent DAO Creation

SudoPumpkin:
## Description

- Several selectors have been updated in the Fractal app to aid in solid test results. 
- A new test to create a 1:1 Gnosis Parent DAO has been created. 
- Test 002 has been updated with a delay. 

## Notes

Playwright needs to be listed in the dependencies of the package.json due to how Playwright references imports. That's why it's been added. 

## Issue (if applicable)

A manual delay was added to allow the app and local wallet to be loaded with multiple workers running at the same time.

## Testing

`npm run tests
`


package.json
```diff
@@ -23,6 +23,7 @@
     "ethers": "^5.7.0",
     "i18next": "^21.9.2",
     "i18next-browser-languagedetector": "^6.1.5",
+    "@playwright/test": "^1.26.0",
```
adamgall, line 26: Please put this in alphabetical order in the dependency list

adamgall (changes requested):

package.json
SudoPumpkin: Done. =)

src/components/ui/Header/MenuItems.tsx
```diff
@@ -108,14 +116,22 @@ function ItemContent({
...
+  testId,
+}: ActionMenuItem) {
   const { t } = useTranslation('menu');
   if (!isVisible) {
     return null;
   }
   return (
     <Menu.Item>
       <button
+        data-testid={testId}
```
adamgall, line 134: Is there a reason why this attribute isn't just called `id`?

src/components/ui/Header/MenuItems.tsx
```diff
@@ -147,11 +164,14 @@ function LinkItem({ titleKey: title, link, Icon }: LinkMenuItem) {
...
 
-function LinkItemInternal({ titleKey: title, link, Icon }: LinkMenuItem) {
+function LinkItemInternal({ titleKey: title, link, Icon, testId }: LinkMenuItem) {
   const { t } = useTranslation('menu');
   return (
     <Menu.Item>
-      <Link to={link}>
+      <Link
+        to={link}
+        data-testid={testId}
```
adamgall, line 173: Is there a reason why this attribute isn't just called `id`?

adamgall (commented):
Left some comments

src/components/ui/Header/MenuItems.tsx
adamgall: i got an answer -- because sometimes React overwrites the `id` attribute

src/components/ui/Header/MenuItems.tsx
```diff
@@ -129,11 +145,12 @@ function ActionItem({ titleKey: title, className, action, Icon, isVisible }: Act
   );
 }
 
-function LinkItem({ titleKey: title, link, Icon }: LinkMenuItem) {
+function LinkItem({ titleKey: title, link, Icon, testId }: LinkMenuItem) {
   const { t } = useTranslation('menu');
   return (
     <Menu.Item>
       <a
+        data-testid={testId}
```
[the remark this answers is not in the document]
adamgall: i got an answer -- because sometimes React overwrites the `id` attribute

src/components/ui/Header/MenuItems.tsx
adamgall: i got an answer -- because sometimes React overwrites the `id` attribute

playwright.config.ts
```diff
@@ -13,7 +13,7 @@ const config: PlaywrightTestConfig = {
      * Maximum time expect() should wait for the condition to be met.
      * For example in `await expect(locator).toHaveText();`
      */
-    timeout: 10000
+    timeout: 15000
```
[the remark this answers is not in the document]
adamgall: I got an answer, it's because stuff sucks and sometimes you just gotta wait a lil longer

tests/002confirmWalletConnect.spec.ts
```diff
@@ -1,24 +1,31 @@
 import { expect, test } from '@playwright/test';
```
herbig, line 1: not a change from this PR, but it might be good to rename this `002confirmWalletConnected`, just so it's not confused with being a test for WalletConnect the wallet brand.

page-objects/components/Navbuttons.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.localWallet = page.locator(
+      '#WEB3_CONNECT_MODAL_ID div.web3modal-provider-name:has-text("Local Node")'
+    );
+    this.next = page.locator('button:has-text("* * *Next")');
+    this.deploy = page.locator('button:has-text("* * *Deploy")');
+  }
+
+  async clickOnButton(buttonName: any) {
+    switch (buttonName) {
+      case 'Return to Homepage':
```
herbig, line 59: Is there a reason for having these designated with string descriptions rather than having a separate method for each clickable item on screen?

I feel like having `clickReturnToHomePage();`, `clickHeaderConnectWallet();`, etc, would be cleaner than a large switch statement with new strings you need to remember when writing the tests.   You also get autocomplete this way.

tests/003createParentDao.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  let notifications: Notifications;
+
+  test.beforeEach(async ({ page }) => {
+    daoSafe = new DaoSafe(page);
+    homePage = new HomePage(page);
+    navButtons = new Navbuttons(page);
+    notifications = new Notifications(page);
+    inputFields = new InputFields(page);
+    homePage.visit();
+    await delay(3000);
```
herbig, line 23: can we put this delay in the `visit();` methods of each page instead?

Da-Colon (approved):
Approved pending Mike's comments

tests/002confirmWalletConnect.spec.ts
SudoPumpkin: I can rename this to ```002connectToWallet```, does this work?

tests/003createParentDao.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    notifications.closeButton('Close Audit Message');
+    await page.waitForLoadState();
+    await delay(1500);
```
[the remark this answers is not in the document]
SudoPumpkin: If the current page has already reached the required state, it resolves the wait immediately. So here no particular state is defined so it's just waiting until load is completed to continue. I can remove this and increase the delay time to '2000'. I had this here to negate a minor amount of ms in wait time compared to a hard increased delay wait. I can totally just add another 500ms(0.5secs) to the present '1500'(1.5secs) to make it '2000'(2secs) and yoink line 25.

page-objects/components/Navbuttons.ts
SudoPumpkin: With similar action components this is a common method for e2e test scripting. It's easier for code readability.  I can change it however to be similar to HomePage.ts.

tests/003createParentDao.spec.ts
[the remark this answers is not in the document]
SudoPumpkin: Actually I am leaving line 25 in there. I removed line 25 locally, and when the tests ran in parallel multiple times they become flaky. A test will random fail in different browser.

tests/002confirmWalletConnect.spec.ts
herbig: yeah that sounds good, I like the distinction

page-objects/components/Navbuttons.ts
herbig: cool cool, sounds good, I defer to you on that, just curious what's common practice

tests/003createParentDao.spec.ts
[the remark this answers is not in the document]
herbig: sounds good.  I'd just like to better understand why the hardcoded waits everywhere are necessary, and what the difference is when running locally

tests/003createParentDao.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  test('Create 1:1 Gnosis Parent DAO', async ({ page }) => {
+    navButtons.clickOnButton('Create a Fractal - Button');
+
+    /* Check URL to make sure navigation is correct. */
+    await expect(page).toHaveURL('http://localhost:3000/#/daos/new');
+
+    /* Input a Fractal name */
+    inputFields.fillField('Insert Fractal Name');
```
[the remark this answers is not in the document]
herbig: that makes sense!

tests/003createParentDao.spec.ts
SudoPumpkin: No idea, but it can't find the menu 'connect' option without it.

tests/003createParentDao.spec.ts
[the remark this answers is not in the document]
SudoPumpkin: Locally it runs with up to 8 workers in parallel while in CI it runs 1 worker. These 8 workers are all running at the same time. Each browser behaves a little bit differently from one another with load times of content with the Fractal app. Now, let's add the local wallet connecting to the app and local blockchain on top of that varied browser behavior. 
The local provider checks to see if the network is active before it connects. When the app loads, it loads this twice(because react in dev). In the middle of this async call the test goes and clicks the connect button. The promise is then returned and interrupts the other state setting. The other 'fallback' provider is not async, there is are promises.

Long story short, sometimes the wallet isn't ready for Playwright to click 'connect' and it 1/2 connects. These hard delays force Playwright to wait for all the pieces to be loaded.

PR #544, opened 2022-10-11
Revert gnosis safe deployments import

### LINKED ISSUE #538: BUG: Local Addresses are not set
## Description
When connected to local hardhat node (blockchain running in docker thats forked from goerli). We are unable to deploy a gnosis safe. Looking at the console I see that an address is not being set for local chain id

image

adamgall:
## Description

This PR reverts PR #526.

The bug exists because, in the code for #526, we pull the static Gnosis addresses out of the objects returned from `getProxyFactoryDeployment` and `getSafeSingletonDeployment`.

When running the app in e2e test mode, the webapp is connected to a local blockchain with `chainId` 31337, which is obviously not a chain id that Gnosis has deployed to. In _our_ reality, that local blockchain is actually a fork of Goerli, so _we know_ that this local blockchain does contain the static Gnosis addresses.

It would have been pretty hacky to "inject" the localhost (Goerli fork) addresses into objects returned from `getProxyFactoryDeployment` and `getSafeSingletonDeployment` in the workflow in `useAddresses`, especially considering we didn't have a good way to programmatically determine which "real" network is being forked in the tests.

So, my approach here was simply to undo all of that, and go back to the original pattern of just setting these addresses from .env, where we can easily control the address value for our localhost blockchain (31337).

## Notes

I have already added these environment variables back into Github Secrets and Netlify Build Settings

## Issue (if applicable)

Closes #538 

## Testing

The easiest way for a reviewer to manually confirm that this code works, is to:

- create a `./docker/.env.tests.local` file, add a `TESTING_ALCHEMY_API_KEY` variable with a valid API key.
- start the local goerli fork test blockchain with `docker compose up blockchain`
-  Somewhere in the app, for example right at the top in `App.tsx`, import `useWeb3Provider` and `useAddresses`, call it at the top of the `App` component, console.log the output

```tsx
  const {
    state: { chainId },
  } = useWeb3Provider();
  console.log(useAddresses(chainId));
```

- start the app with `npm run start`
- in the console output, notice how the address records for `gnosisSafeFactory` and `gnosisSafe` include the correct addresses.


herbig (approved):
ship it

SudoPumpkin (commented):
I currently get a MetaMask popup asking me to login again after I am already logged in and connected to the app. Perhaps related or perhaps another ticket issue. Please advise. 

image



https://user-images.githubusercontent.com/86701852/195125288-baf60612-da1b-4359-9767-adc6df47c064.mp4

Da-Colon (approved):

PR #551, opened 2022-10-13
Basic Github issue templates

herbig:
## Description

@Da-Colon pointed me to Shapeshift's beautiful issue templates, so I investigated those and found instructions for creating our own:

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

These are a paired down version of Shapeshift's, from which we can improve upon based on what works for the team.

## Notes

Please add suggestions about fields I might be missing.

## Testing

There doesn't seem to be a way to test these until they're merged into master 🤷‍♂️.  If there are issues once this gets merged I will create a PR to fix...

### DIFF
--- a/.github/bugs.yml
+++ b/.github/bugs.yml
@@ -1,0 +1,56 @@
+name: Bug
+description: Report a bug
+title: "[Bug]: "
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Please give your bug report a short and descriptive title.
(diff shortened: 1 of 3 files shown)


herbig:
Ah shoot, I built this PR off of dev, I need to do so off of master.

herbig:
Closing this, as it needs to be based off master, will open another PR

PR #552, opened 2022-10-13
Github issue templates

herbig:
## Description

@Da-Colon pointed me to Shapeshift's beautiful issue templates, so I investigated those and found instructions for creating our own:

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

These are a paired down version of Shapeshift's, from which we can improve upon based on what works for the team.

## Notes

Please add suggestions about fields I might be missing.

## Testing

There doesn't seem to be a way to test these until they're merged into master 🤷‍♂️.  If there are issues once this gets merged I will create a PR to fix...

You can see Shapeshift's Issue flow by creating a new issue there:
https://github.com/shapeshift/lib/issues/new/choose

And view their templates:
https://github.com/shapeshift/lib/tree/main/.github/ISSUE_TEMPLATE


.github/config.yml
```diff
@@ -0,0 +1,11 @@
+blank_issues_enabled: true
```
herbig, line 1: there should still be an option to create a regular 'ol issue, without a template

.github/bugs.yml
```diff
@@ -0,0 +1,56 @@
+name: Bug
+description: Report a bug
+title: "[Bug]: "
+labels: ["bug"]
```
herbig, line 4: We can auto preface the title and assign labels to issues created with these templates.

.github/bugs.yml
```diff
@@ -0,0 +1,56 @@
...
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Please give your bug report a short and descriptive title.
+  - type: textarea
```
herbig, line 10: @SudoPumpkin since you're the one most reporting bugs, are there any fields you'd suggest?

I have Overview, Steps to Reproduce, Screenshots, Version, and Browser, which seems sufficient to start

SudoPumpkin: Wallet Type maybe? MetaMask, WalletConnect(Then w/e wallet) The rest of the fields look accurate.

Da-Colon (commented):
Does this need to be based off master? The other template I made for the pull requests is only in develop?

herbig:
> Does this need to be based off master? The other template I made for the pull requests is only in develop?

Yeah it makes sense that PR templates don't need to be, since you could have different templates based on which branch is merging into what.

For Github to recognize and add new functionality to the Issue creation, which is only repository specific, it needs to be on master I believe, the setup instructions said "Commit your file to the default branch of your repository."

.github/bugs.yml
herbig: added those, thanks!

adamgall:
> > Does this need to be based off master? The other template I made for the pull requests is only in develop?
> 
> 
> 
> Yeah it makes sense that PR templates don't need to be, since you could have different templates based on which branch is merging into what.
> 
> 
> 
> For Github to recognize and add new functionality to the Issue creation, which is only repository specific, it needs to be on master I believe, the setup instructions said "Commit your file to the default branch of your repository."

The default branch of this repo is `develop`

herbig:
> > > Does this need to be based off master? The other template I made for the pull requests is only in develop?
> > 
> > 
> > Yeah it makes sense that PR templates don't need to be, since you could have different templates based on which branch is merging into what.
> > For Github to recognize and add new functionality to the Issue creation, which is only repository specific, it needs to be on master I believe, the setup instructions said "Commit your file to the default branch of your repository."
> 
> The default branch of this repo is `develop`

ahhhh ok

Da-Colon (approved):
LFG

herbig:
@Da-Colon whelp, this is merged and the `New issue` button has no changes....

Da-Colon:
It loos like it still running actions on develop. I'd give it a bit

herbig:
> It loos like it still running actions on develop. I'd give it a bit

yeah I also didn't have them in the proper directory...
https://github.com/decent-dao/fractal-interface/pull/567

PR #553, opened 2022-10-14
V1 Design | Wallet Provider Update

Da-Colon:
## Description

When working on the redesign, a opportunity to optimize the wallet provider came up. When trying to auto correct after connecting and refreshing the page, 'Connected' toast would show but the wallet would be connect to the fallback provider. On closer look I noticed that a similar thing that were found in the tests was happening where state was being updated out of order causing the async methods not returning (best I can do to explain what I observed)

## Notes

- On the initial load, instead of checking for a cachedProvider, the fallback provider is connected to (local or specified fallback). Once this is set only then is the cachedProvider is checked and connected to. This creates a single flow and eliminates side effects

## Issue (if applicable)

- No issue created

## Testing

I've done some testing locally but won't be available until the header PR is posted

n/a


src/contexts/web3Data/Web3Provider.tsx
```diff
@@ -3,106 +3,109 @@ import Web3Modal from 'web3modal';
...
+      return { ...initialState, isProviderLoading: false, ...localProviderInfo };
+    }
+  }
+  return { ...initialState, isProviderLoading: false, ...getFallbackProvider() };
+};
+
 export function Web3Provider({ children }: { children: React.ReactNode }) {
-  const [state, dispatch] = useReducer(reducer, getInitialState());
+  const [state, dispatch] = useReducer(reducer, initialState, getInitialState);
+  const web3Modal = useMemo(() => new Web3Modal(WEB3_MODAL_CONFIG), []);
```
Da-Colon, line 62: move this into the function, I originally moved this as I thought maybe this wasn't being set correctly. I don't think it was causing anything but felt it should probably go here anyway

src/contexts/web3Data/Web3Provider.tsx
```diff
@@ -3,106 +3,109 @@ import Web3Modal from 'web3modal';
...
+  }, [web3Modal]);
 
+  /**
+   * Connects using web3Modal
+   * If the connect network is not supported the fallback provider is set
+   */
   const connect: ConnectFn = useCallback(async () => {
     const userInjectedProvider = await getInjectedProvider(web3Modal);
     if (!!userInjectedProvider && getSupportedChains().includes(userInjectedProvider.chainId)) {
+      toast(t('toastConnected'), { toastId: 'connected' });
```
Da-Colon, line 96: moved the connect toast here as it was firing when not connecting in the listener. This makes sure state is setting when displaying toast

src/contexts/web3Data/Web3Provider.tsx
```diff
@@ -111,15 +114,13 @@ export function Web3Provider({ children }: { children: React.ReactNode }) {
...
+    if (web3Modal.cachedProvider && !state.isProviderLoading) {
+      (async () => {
+        await connect();
+      })();
     }
-    connectDefaultProvider();
-  }, [connect, connectDefaultProvider]);
-
-  useEffect(() => load(), [load]);
+  }, [connect, state.isProviderLoading, web3Modal]);
```
Da-Colon, line 123: The big fix was here. This now waits until the initialState is set (fallbackProvider) then attempts to connect if a cachedProvider exists

src/contexts/web3Data/types.ts
```diff
@@ -51,3 +51,41 @@ export type ProviderApiKeys = {
...
 };
+
+export interface Web3ModalProvider {
+  _isProvider: boolean;
+  _events: [];
+  _emitted: {
+    block: number;
+  };
+  disableCcipRead: false;
+  formatter: {
```
Da-Colon, line 62: This is the returned type from the Web3Modal. `const provider = await web3Modal.connect()`
I wanted to type it since it was returning 'any', more of a reference versus a need

src/contexts/web3Data/utilities/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './fallbackProvider';
+export * from './injectedProvider';
+export * from './localProvider';
```
Da-Colon, line 3: I like separating concerns

herbig (approved):
🎉

PR #554, opened 2022-10-15
CONTRIBUTING.md and translation helper scripts

herbig:
## Description

Adding a `CONTRIBUTING.md`, inspired by ShapeShift's language, along with scripts to facilitate community translations.

The purpose of the two scripts is to translate from a series of `json` files into a single spreadsheet, to easily see what translations are missing for a given language.

You can then take the updated spreadsheet and output the `json` files back again, with any updates that have been made.

Keeping translation files in sync can be a headache, and having community members sift through multiple json files and pick out which strings are missing is a bit much to ask, so making that process a bit smoother will increase the likelihood of community contributions.

## Notes

I used python for these scripts just because it was the most efficient for command line scripting.  I added a lot of comments, in case folks aren't as familiar with python, but it's pretty readable.

## Testing

You can run these scripts right from VS Code, or navigate to the script directory and run with `python3 {name}.py`

Converting our current Spanish translations from .json to .csv and back again should incur no changes.


src/i18n/locales/es/common.json
```diff
@@ -1,4 +1,4 @@
-{   
+{
```
herbig, line 1: this was the only change when going from json to csv and back again, which was to remove whitespace, so I kept it so it wouldn't appear when we go ahead with translations in the future

CONTRIBUTING.md
```diff
@@ -0,0 +1,42 @@
...
+- All of your components should be composed of [Chakra UI](https://chakra-ui.com/) base components, using our custom theme.
+
+- A member of the Fractal engineering team will respond with any further discussion or requested changes and once approved
+  you are free to merge.
+  
+- We appreciate the collaboration!
+
+### Submitting Language Translations
+
+- Languages we are actively seeking translations for will appear as a Github Issue, though we always appreciate other
```
herbig, line 31: @Da-Colon how do we do bounties? We should probably describe that process a bit here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,42 @@
...
+
+- Some issues may already be reported, please search existing issues before submitting a new one.
+
+- Otherwise, simply create a new issue and the Fractal development team will respond in a timely manner. Please be as descriptive of the issue as possible.
+
+- If looking to contribute a bug fix or feature contribution, please discuss the change via the bug ticket or a new Github issue before submitting a pull request for the greatest likelihood of being accepted.
+
+### Submitting a Pull Request
+
+- Follow our [README](https://github.com/decent-dao/fractal-interface/blob/develop/README.md) for instructions on setting up your local environment.
```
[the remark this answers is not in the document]
herbig: yeah good catch, I just copied it from where I found it, ha

utils/i18n/2_csv_to_json.py
```diff
@@ -0,0 +1,52 @@
+from pathlib import Path
```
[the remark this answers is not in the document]
herbig: yeah, I had the numbers just to make the steps even clearer.  Especially for more nontechnical folks who might be able to translate but less familiar with running scripts, I wanted to make it as clear as possible what to do

CONTRIBUTING.md
Da-Colon: I would refer to @adamgall Do we have the process defined yet on how we go from idea of bounty to dework to github issues?

CONTRIBUTING.md
```diff
@@ -0,0 +1,48 @@
...
+  ```console
+  python3 1_json_to_csv.py
+  ```
+
+- Add your translations in the last column of the output CSV file.  Feel free to improve existing translations!
+
+- When finished translating, run `2_csv_to_json.py` to generate the required `.json` files for the Fractal app:
+  ```console
+  python3 2_csv_to_json.py
+  ```
```
Da-Colon, line 44: its obviously with the command but might be a good idea to mention you need Python 3 and a link to 'getting started' instructions or home page (Python 3)

herbig: good call, updated with a link to download / install if they don't have it already

PR #566, opened 2022-10-17
V1 design | Header Cleanup

Da-Colon:
## Description
This PR covers a few smaller things to clean up from the header refactor. Removing no longer needed library, removing unused components and updating Test ids and tests

## Notes
The large number of file changes is due to adding test ids

## Testing
So I have updated tests 001 (page renders) and 002 (users can connect) to pass now. 003 (create gnosis workflow) I have added the `skip` method to skip this test for now until the workflow has been added and the tests can be updated. Yay passing tests!. 

When it comes time to merge the feature branch into develop I will work with @SudoPumpkin to updated any skip tests and update the helper file with new test ids.


src/App.tsx
```diff
@@ -12,7 +12,7 @@ function App() {
 
   useActionToast({
     toastId: 'audit:toast',
-    testId: 'audit:accept',
```
Da-Colon, line 15: One thing I noticed in the tests that `:` isn't allowed as a valid Selector. so moving forward moving to camelCase with a intent to keep it `pageComponent-descriptor`

src/i18n/index.ts
```diff
@@ -44,12 +46,14 @@ const resources = {
...
     transaction: TRANSACTION_EN,
     treasury: TREASURY_EN,
+    walletMenu: WALLET_MENU_EN,
   },
   es: {
     common: COMMON_ES,
     daoCreate: DAOCREATE_ES,
     menu: MENU_ES,
     treasury: TREASURY_ES,
+    walletMenu: WALLET_MENU_ES,
```
Da-Colon, line 56: I went ahead and created new json files specifically for the new wallet menu. wasn't wanting to touch the current setup, until we merge feature into develop. 

@herbig thoughts here?

tests/001navigateToFractal.spec.ts
```diff
@@ -5,6 +5,6 @@ test('Go to the Fractal App', async ({ page }) => {
   await page.goto('http://localhost:3000');
 
   /* Assert Fractal landing page is present */
-  const welcome = page.locator('h1');
-  await expect(welcome).toContainText('Welcome');
+  const headerWalletMenu = page.locator('[data-testid=header-accountMenu]');
+  await expect(headerWalletMenu).toBeVisible();
```
Da-Colon, line 9: @SudoPumpkin I moved the selector to the menu button itself. The pervious library made this more difficult than it needed to be.

tests/002confirmWalletConnect.spec.ts
```diff
@@ -1,16 +1,14 @@
...
   await page.goto('http://localhost:3000');
-  await delay(5000);
 
   /* Click("Connect Wallet") */
-  await page.locator("//div[@id='menu:down-arrow']//*[name()='svg']").click();
+  await page.locator('[data-testid=header-accountMenu]').click();
 
   /* Dropdown menu of "Connect Wallet" */
-  await page.locator('//span[normalize-space()="Connect"]').click();
+  await page.locator('button[data-testid=accountMenu-connect]').click();
```
Da-Colon, line 11: Let's default to using `data-testid` moving forward to select things. Unless its a 3rd party thing we need to lock on to. but selecting something should start with a testid then find the needed button as needed (unless the selector is on a button)

tests/002confirmWalletConnect.spec.ts
```diff
@@ -23,11 +21,13 @@ test('Confirm Wallet is Connected to Fractal', async ({ page }) => {
...
 
   /* Assert defined wallet address is present ("0xf39F...2266") */
-  await page.locator('//div[@id="menu:down-arrow"]//*[name()="svg"]').click();
-  const walletAddress = page.locator('//a[normalize-space()="0xf39F...2266"]');
-  await expect(walletAddress).toContainText('0xf39F...2266');
+  await page.locator('[data-testid=header-accountMenu]').click();
+  const accountDisplay = page.locator(
+    '[data-testid=accountMenu-wallet] [data-testid=walletMenu-accountDisplay]'
+  );
+  await expect(accountDisplay).toContainText('0xf39F...2266');
```
Da-Colon, line 28: Might want to think about updating this test to just check that the Account menu test is no longer 'Connect Wallet' as we don't want this test to depend on text that we don't fully control appearing.

tests/003createParentDao.spec.ts
```diff
@@ -6,7 +6,8 @@ import { InputFields } from '../page-objects/InputFields';
 import { Navbuttons } from '../page-objects/components/Navbuttons';
 import { Notifications } from '../page-objects/Notifications';
 
-test.describe('DAO Creation', () => {
+// @todo update when components are added
+test.describe.skip('DAO Creation', () => {
```
Da-Colon, line 10: skip

tests/002confirmWalletConnect.spec.ts
```diff
@@ -1,16 +1,14 @@
 import { expect, test } from '@playwright/test';
-import { delay } from '../page-objects/Helpers/helpers';
 
 test('Confirm Wallet is Connected to Fractal', async ({ page }) => {
   /* Go to local host http link */
   await page.goto('http://localhost:3000');
-  await delay(5000);
```
Da-Colon, line 7: I think I may have solved this issue with the lastest update to the wallet-provider that caused the need for this delay. We will know for sure when more tests are added/enabled.

SudoPumpkin: Awww yeah. 🤞

tests/002confirmWalletConnect.spec.ts
herbig: @SudoPumpkin we'll have to look at tests once we start pushing these changes, since they're all going to break in different ways..

tests/002confirmWalletConnect.spec.ts
herbig: @SudoPumpkin

src/i18n/locales/es/walletMenu.json
```diff
@@ -0,0 +1,4 @@
+{
+  "network": "la Red",
+  "wallet": "Cartera"
```
herbig, line 3: Unless we're pretty sure about translations (like @mudrila with Ukrainian) we should probably not include our own or direct google translations, since context can change things.

In this example, `cartera` directly translates to `wallet`, but `crypto wallet` uses the word `billetera`, so I'm not really sure which word makes more sense to native speakers.

I added Google translations for the menu just as a way to test, but we should probably also remove those as well.

Once the design is finalized and we have a copy review with design, we should be able to find community translators who would love to help us out.

src/i18n/index.ts
herbig: I think it's fine to change the current files, and add these to maybe menu.json.

We should only ever be adding to the English (en) json file, and let translations be done via community translators using the scripts we have now, which will help keep them organized in the same order and keep us from missing translations.

herbig (changes requested):
Looks good, just a few more strings that can be moved to json files.

src/i18n/locales/es/walletMenu.json
Da-Colon: for sure, I've removed this file

herbig (approved):
🚢

PR #575, opened 2022-10-19
V1 Design | Account Menu Update

Da-Colon:
## Description

Noticed that my implementation was a little off from the designs. 

- The background colors were not odd/even rather the wallet menu was a different color alone with a rounded bottom
- Added a border to the top of the button
- Got rid of the over engineered component (MenuItem.tsx)

image

image


herbig (approved):
Nice, starting to come togetherrrr

src/components/menus/AccountDisplay/MenuItemButton.tsx
```diff
@@ -17,18 +17,27 @@ export function MenuItemButton({
...
   return (
-    <MenuItem
-      testId={testId}
-      onClick={onClick}
-    >
-      <Flex
-        alignItems="center"
-        justifyContent="space-between"
-        w="full"
+    <Fragment>
```
christopherdancy, line 20: Hey @Da-Colon what exactly is happening with this fragment?

christopherdancy:
Also I am not sure if I missed something but is the web page only supposed to show the wallet connection?

image

src/components/menus/AccountDisplay/MenuItemButton.tsx
Da-Colon: It is basically the same as doing `<></>` https://reactjs.org/docs/fragments.html

TL:DR its bad pattern to have multiple elements passed to the return

```jsx
return (
  <div>...</div>
  <div>...<div>
)
```

So instead it should be
```jsx
return (
  <div>
    <div>...</div>
    <div>...</div>
  </div>
)
```

Fragments lets you return the multiple elements without having adding an extra element to wrap them render. (Fragments are not rendered in the DOM) so you just see the elements tab on the dev tools when you wrap them in the <Fragment> tag

```jsx
<div>...</div>
<div>...</div>
```

Sorry if this is an over explanation lol

Da-Colon:
> Also I am not sure if I missed something but is the web page only supposed to show the wallet connection?
> 
> image

what do you mean? see https://www.figma.com/file/uRV3Nfowdzt5CknW1FgUkQ/%F0%9F%93%9A-Fractal-Design-System-%7C-Base-Components-%7C-October-2022?node-id=1802%3A12132

herbig:
> > Also I am not sure if I missed something but is the web page only supposed to show the wallet connection?
> > image
> 
> what do you mean? see https://www.figma.com/file/uRV3Nfowdzt5CknW1FgUkQ/%F0%9F%93%9A-Fractal-Design-System-%7C-Base-Components-%7C-October-2022?node-id=1802%3A12132

I think @christopherdancy means why is the page blank, ha.  Yeah these PRs are just smaller component specific implementation PRs, merged into the `feature/v1-design branch`, not develop

Lots of stuff will be broken along the way because of the UI redesign.

src/components/menus/AccountDisplay/MenuItemButton.tsx
christopherdancy: Thanks for the through discussion! This is great. Sorry just want to make sure I understand.

christopherdancy:
Yes lol. approving now

src/components/menus/AccountDisplay/MenuItemButton.tsx
herbig: And `<>` is just shorthand for `<Fragment>`, they do the same thing, but we discussed in a different PR that for better readability we'd like to write it out.  More of a coding style preference.

christopherdancy (approved):
LFG

PR #587, opened 2022-10-23
V1 Design | Header Depth and Wallet Menu

Da-Colon:
## Description
Quick bug fix. Noticed that the menu was being rendered behind some content on the page. Added 'sticky' zIndex. which is a chakra-ui key for 1100.

## Notes

I also went ahead and updated the styling of the wallet menu. noticed the width and padding was quite off as I was creating a screenshot for the quick fix.

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -47,6 +47,7 @@ function App() {
           bg="chocolate.900"
           h="4rem"
           position="fixed"
+          zIndex="sticky"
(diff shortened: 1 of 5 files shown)


herbig (approved):
Nice, yeah this is an issue with the old UI as well, so good to fix that.

Da-Colon:
Tests are broken but I'll update as needed based on #589 new setup

PR #598, opened 2022-10-26
V1 Design | header favorites

Da-Colon:
## Description
Implements Favorites Menu

## Notes
-  useDisplayName hook will need to be updated when the DAO name register contract is implemented

image


src/components/menus/FavoritesMenu/FavoritesList.tsx
```diff
@@ -0,0 +1,85 @@
...
+                  border: 'none',
+                  boxShadow: 'none',
+                  background: 'grayscale.500',
+                  borderRadius: '0.5rem',
+                  minHeight: '2.5rem',
+                },
+                '&::-webkit-scrollbar-thumb:hover': {
+                  backgroundColor: 'grayscale.300',
+                },
+              }}
```
Da-Colon, line 71: default scroll looked really bad here. I will look into adding a class to the theme that can be used instead

herbig:
@Da-Colon is there a way I can set a DAO as a favorite in this branch?

src/components/menus/FavoritesMenu/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+          <Flex
+            gap="2"
+            alignItems="center"
+          >
+            <StarOutline color="gold.500" />
+            <Text
+              textStyle="text-sm-mono-medium"
+              color="gold.500"
+            >
+              Favorites
```
herbig, line 20: should use `titleFavorites` here

herbig (changes requested):

Da-Colon:
> @Da-Colon is there a way I can set a DAO as a favorite in this branch?

Not yet. I believe this will be added on the Dashboard when thats complete

src/components/menus/FavoritesMenu/index.tsx
Da-Colon: done.

PR #625, opened 2022-11-02
Refactor dApp Structure 02 | Delete Components

Da-Colon:
## Description

This one covers several no longer needed components.


src/components/DaoCreator/provider/hooks/useSteps.ts
```diff
@@ -9,22 +8,14 @@ import { useFractal } from '../../../../providers/fractal/hooks/useFractal';
...
-    const isGnosisWithUsul = state.governance === GovernanceTypes.GNOSIS_SAFE_USUL;
     const isPureGnosis = state.governance === GovernanceTypes.GNOSIS_SAFE;
 
     // True when:
     // 1) The DAO being created is a subDAO,
-    // 2) The DAO being created is a token voting DAO
-    // 3) The parent DAO is not a Gnosis Safe DAO
-    const showFunding = isSubDAO && isTokenGovernance && !gnosisWrapperModule;
+    // @note setting this to false for now, this should be set to the isSubDAO prop
+    const showFunding = false;
```
herbig, line 18: subfunding we may actually hold off on sub funding altogether for v1, so this is fine to keep disabled for now

src/providers/govenor/hooks/useUserProposalValidation.tsx
```diff
@@ -45,7 +39,8 @@ export function useUserProposalValidation() {
             title={t('errorNoDelegates')}
             label={t('buttonDelegate')}
             action={() => {
-              navigate(`/daos/${daoAddress}/modules/${governorModuleContract!.address}/delegate`);
+              // @todo update route
+              // navigate(`/daos/${daoAddress}/modules/${governorModuleContract!.address}/delegate`);
```
herbig, line 43: @Da-Colon where is this action used? In the proposal creation? @tbwebb22 might be good to keep in mind this will need to be reenabled once delegation is re implemented

herbig (approved):
rm -rf

src/providers/govenor/hooks/useUserProposalValidation.tsx
Da-Colon: yessir, I left the governance components and hooks untouched (except for removing what I needed too for the app not to break on us). so we have a reference before we have another delete party

tbwebb22: Cool, will do

tbwebb22 (approved):

PR #627, opened 2022-11-02
Refactor dApp Structure 04 | Removes/Updates Page Components

Da-Colon:
## Description

There is probably some left over hooks but this takes care of any other screens that needed to be updated to use the new FractalProvider or be removed.


src/pages/Treasury/index.tsx
```diff
@@ -21,13 +20,15 @@ function TableRowWrapper({ children }: { children?: ReactNode }) {
...
+  const {
+    gnosis: { safe },
+    treasury: { assetsFungible, assetsNonFungible },
+  } = useFractal();
 
   const { t } = useTranslation('treasury');
   return (
     <div>
-      <H1>{t('titleTreasury', { daoName: state.name })}</H1>
+      <H1>{t('titleTreasury', { daoName: '' })}</H1>
```
herbig, line 31: @Da-Colon I realize we don't have name actually loading yet, but does the `gnosis` `safe` have a name field to reference here?

src/pages/Treasury/index.tsx
```diff
@@ -180,9 +181,9 @@ function Treasury() {
           ))}
         </div>
       </div>
-      {!gnosisAssetsFungible!.length && !gnosisAssetsNonFungible.length && (
+      {!assetsFungible!.length && !assetsNonFungible.length && (
         <div className="px-1">
-          <ContentBanner description={t('descTreasury', { daoName: state.name })} />
+          <ContentBanner description={t('descTreasury', { daoName: '' })} />
```
herbig, line 186: same with this, it seems like we should have a name field in state somewhere even if it's not there yet

herbig (commented):
just questions about the name field! Looks good otherwise

src/pages/Treasury/index.tsx
Da-Colon: not at the moment nope.

herbig: @mudrila could you make note to put the name in here once your work on the name contract is good to go

tbwebb22 (approved):
LGTM

PR #637, opened 2022-11-03
speeds up development server launch

Da-Colon:
## Description
Reduces typescript version. Was reading that eslint and typescript can effect sever deploying speeds. can definately tell this speeds it up a little.


package.json
```diff
@@ -81,6 +81,6 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "playwright": "^1.26.0",
     "prettier": "^2.7.1",
-    "typescript": "^4.8.4"
+    "typescript": "^4.8.0"
```
herbig, line 84: does the speed increase justify downgrading versions? I guess I don't know enough about typescript versions to know if this matters

Da-Colon: This is the warning I was solving for. Sorry probably should have added more context to this change.

image

This step was slowing down the launching the development server

herbig: cool cool, sounds good then

herbig (approved):

Da-Colon:
@mudrila basically as I understand it. eslint and typescript communicating with each other can cause problems if versioning is off. been noticing for a while that the difference between a fresh project (even with a few things installed) `npm run start` doesn't take as long as it does for fractal. Looking into it, this is a part of what we can do to tighten it up.

PR #638, opened 2022-11-03
centers splash screen

Da-Colon:
## Description
Noticed that the splash screen was scrolling. added `calc(100vh - 4rem) to account for header

image


src/pages/Home/index.tsx
```diff
@@ -93,7 +93,7 @@ function Home() {
     navigate(DAO_ROUTES.new.relative);
   };
   return (
-    <Center h="100vh">
+    <Center h="calc(100vh - 4rem)">
```
herbig, line 96: can we add this to a constant somewhere? In case we need to reference it again elsewhere it'd be better to define it.

herbig: HEADER_HEIGHT = 4rem or something like that I mean

Da-Colon: good idea, done.

herbig (approved):

PR #642, opened 2022-11-05
Treasury redesign

herbig:
## Description

Implements a redesign of the Treasury page.

## Notes

- This PR doesn't implement the Send Assets button modal, that will come in a later PR.
- This also doesn't implement historical price, as the designs show in the Transactions list.  This is a bit more complex and will come in a later PR as well.
- Coin icons and NFT images aren't returned via the Safe API on **Georli**.  I've checked the **mainnet** API and they return as expected, but for **Georli** this page is going to look janky, even when I have default images.

Still waiting on three assets:

- Ether token icon
- fallback NFT image
- fallback coin icon

I've stubbed all three just with plain colors for now...

## Issue / Notion doc (if applicable)

Figma design:
https://www.figma.com/file/nKTAQ6bivtHMwc3O5FG47l/%F0%9F%93%9A-Fractal-Design-Handoff-%7C-Approved-Pages-%7C-October-2022?node-id=2252%3A19350

Closes https://github.com/decent-dao/fractal-interface/issues/616

## Testing

Please test with DAOs that have a number of assets.  I've been using:
http://localhost:3000/#/daos/0xB87AD11E40F66137BD4cbB5f9d9438451000251B/treasury

## Screenshots (if applicable)
Screen Shot 2022-11-07 at 5 01 48 PM

PR #643, opened 2022-11-06
V1 Design | Dashboard Activities

Da-Colon:
## Description
Updates the DAO Dashboard with an Activity Feed. This update only shows Treasury Events currently. Because of the nature of the data returned from the endpoint.

Once we are pulling in Proposal Data from both Usul and Multisig we can update this page to show that data.

image


src/pages/DaoDashboard/Activities/hooks/useActivities.ts
```diff
@@ -0,0 +1,187 @@
...
+                  : transfer.to
+              );
+              return addressOrName;
+            })
+          );
+
+          return {
+            transaction,
+            eventDate: getTimestampString(new Date(transaction.executionDate)),
+            eventType: 'Treasury',
```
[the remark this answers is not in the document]
Da-Colon: Currently this isn't being used, I was going to use this in the future when we add in Proposal Data. At that time will update into a ENUM for sure

src/pages/DaoDashboard/Activities/hooks/useActivities.ts
```diff
@@ -0,0 +1,187 @@
...
+          const transferAmountTotalsArr = Array.from(transferAmountTotalsMap.values()).map(
+            token => {
+              const totalAmount = formatWeiToValue(token.bn, token.decimals);
+              const symbol = token.symbol;
+              return `${totalAmount} ${symbol}`;
+            }
+          );
+          const transferAddressesDisplayArr = await Promise.all(
+            transaction.transfers.map(async transfer => {
+              const addressOrName = await getAddressDisplay(
```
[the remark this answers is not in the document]
Da-Colon: The hook wouldn't be able to used for this. As the transfers addresses are mapped. multiple to one transaction.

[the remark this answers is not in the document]
Da-Colon: because we aren't displaying the tooltip I will add that it won't be using this for every address just if the length of addresses is less than 3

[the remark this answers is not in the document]
Da-Colon: ugh fine you may be right. let me look at this.

Da-Colon:
@mudrila All comments have been addressed sir. good idea taking that address out. I do see an update that needs to be made to the `useDisplayName` hook. it shows to subString address for a spilt second before it shows ENS name. we probably need this thing to wait before it spits out a result until the lookup is finished

src/components/ui/EtherscanLinkAddress.tsx
```diff
@@ -1,14 +1,16 @@
 import useSubDomain from '../../hooks/useSubDomain';
 
 function EtherscanLinkAddress({
+  path = 'address',
   address,
   children,
 }: {
+  path?: string;
```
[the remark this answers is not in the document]
herbig: There's a separate `EtherscanTransactionLink` component for that as well.

We can merge the two into a single component, but would be good to change the name to something more generic, or keep the two separate.

src/components/ui/containers/InfoBox.tsx
```diff
@@ -0,0 +1,25 @@
+import { Box } from '@chakra-ui/react';
+import { ReactNode } from 'react';
+
+export function InfoBox({
```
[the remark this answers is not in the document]
herbig: hahahaha, oh no I renamed my own component `InfoBox` 🤦

src/components/ui/loaders/BarLoader.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { motion } from 'framer-motion';
+
+const animationKeyframes = keyframes`
+  0%   { height: 48px} 
+  100% { height: 4px}
+`;
+
+const animation = `${animationKeyframes} 0.75s 1s linear infinite alternate`;
+
+export function BarLoader() {
```
herbig, line 11: this looks pretty sick in the app, I might want to add this to the Treasury

src/components/ui/containers/AcitivityBox.tsx
```diff
@@ -0,0 +1,21 @@
+import { Box, SlideFade } from '@chakra-ui/react';
+import { ReactNode } from 'react';
+
+export function ActivityBox({ children }: { children?: ReactNode }) {
+  return (
+    <SlideFade
```
herbig, line 6: Oh nice, might need to think about standardizing component animations in some way as well.

src/i18n/locales/en/common.json
```diff
@@ -53,5 +53,9 @@
     "labelDateYear_other": "{{count}} years",
     "toastClipboardCopy": "Copied to clipboard",
     "newest": "Newest",
-    "oldest": "Oldest"
+    "oldest": "Oldest",
+    "received": "Recieved",
+    "send": "Send",
```
herbig, line 58: got a typo in `Recieved`, and I assume this should be `"sent": "Sent"`?

src/utils/currency.ts
```diff
@@ -0,0 +1,9 @@
+import { utils } from 'ethers';
+
+export const formatWeiToValue = (amountInWei: string, decimals: number) => {
```
herbig, line 3: I have a couple similar formatting functions in the Treasury PR:

https://github.com/decent-dao/fractal-interface/pull/642/files#r1016150767

src/types/activity.ts
```diff
@@ -0,0 +1,32 @@
...
+  eventDate: string;
+  eventType: ActivityEventType;
+  isDeposit: boolean;
+};
+
+export enum ActivityEventType {
+  Treasury,
+}
+
+export enum GnosisTransferType {
```
herbig, line 28: this is the same as the Transfers API request, so we'll have to merge some of these models at some point

src/providers/fractal/types/gnosis.ts
```diff
@@ -42,3 +42,48 @@ export interface GnosisTransactionAPI {
...
+export interface TokenInfoResponse {
+  type: string;
+  address: string;
+  name: string;
+  symbol: string;
+  decimals: number;
+  logoUri: string;
+}
+
+export interface GnosisTransfer {
```
herbig, line 55: Same with these models, I have the same stuff in the Treasury PR, blast

src/pages/DaoDashboard/Activities/index.tsx
```diff
@@ -1,27 +1,42 @@
...
           sortBy={sortBy}
           setSortBy={setSortBy}
         />
       </Flex>
+      {isActivitiesLoading ? (
+        <InfoBoxLoader />
+      ) : (
+        <Flex
+          flexDirection="column"
+          gap="1rem"
```
herbig, line 30: Is this going to leave a gap somewhere if there are no activities in `sortedActivities` ?

src/pages/DaoDashboard/Activities/hooks/useActivities.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  const parsedActivities = useMemo(() => {
+    if (![...transactions.results].length || !safe) {
+      return [];
+    }
+    return [...transactions.results]
+      .filter(t => !!t.transfers.length)
+      .map(transaction => {
+        const isDeposit = transaction.transfers.every(
+          t => t.to.toLowerCase() === safe.address!.toLowerCase()
```
herbig, line 34: you probably can't assume `safe.address!` here because of the or `||` condition above, but ? should do the same here I think?

src/pages/DaoDashboard/Activities/hooks/useActivities.ts
```diff
@@ -0,0 +1,157 @@
...
+          return prev;
+        }, new Map());
+
+        const transferAmountTotalsArr = Array.from(transferAmountTotalsMap.values()).map(token => {
+          const totalAmount = formatWeiToValue(token.bn, token.decimals);
+          const symbol = token.symbol;
+          return `${totalAmount} ${symbol}`;
+        });
+        const transferAddresses = transaction.transfers.map(transfer =>
+          transfer.to.toLowerCase() === safe.address!.toLowerCase() ? transfer.from : transfer.to
```
herbig, line 89: similarly, I'm not sure what the React result would be here if we don't have a safe.address, but I think doing the safe thing of `safe.address?.toLowerCase()` would do the same thing here?

src/pages/DaoDashboard/Activities/ActivityTreasury.tsx
```diff
@@ -0,0 +1,104 @@
...
+  transferAddresses: string[];
+}) {
+  return (
+    <Flex
+      color="grayscale.100"
+      textStyle="text-lg-mono-semibold"
+      gap="0.5rem"
+      flexWrap="wrap"
+    >
+      <Text>{transferTypeStr}</Text>
```
herbig, line 38: do these text components need to be styled at all, or is the default capturing what it should be?

src/pages/DaoDashboard/Activities/ActivityTreasury.tsx
```diff
@@ -0,0 +1,104 @@
...
+      <Text>{totalAmounts.join(', ')}</Text>
+      <Text>{transferDirStr}</Text>
+      {transferAddresses.length > 2 ? (
+        <Text>{transferAddresses.length} addresses</Text>
```
herbig, line 42: `{{num}} addresses` can be in our json strings

src/pages/DaoDashboard/Activities/ActivityCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+interface IAcitivityCard {
+  eventDate: string;
+  description: ReactNode;
+  RightElement?: ReactNode;
+  Badge?: ReactNode;
+}
+
+export function AcitivityCard({
+  Badge,
+  eventDate = '2022-10-03',
```
herbig, line 15: don't think you need the value here anymore

herbig (changes requested):
A few comments!

Also some things from testing:

1. the activity feed needs the same default left/right padding that the upper cards have (see screenshot)
2. treasury breaks a bit weird at one point and doesn't fill the screen, also without a lot of content in there maybe it's better to break it a different way, but I'm not really sure
3. The Newest/Oldest filter opens SUPER slow on Firefox, but not on Chrome, for some reason.  I actually think maybe it should just toggle between the two without needing to open the dialog to select one.
Screen Shot 2022-11-08 at 1 14 52 AM
Screen Shot 2022-11-08 at 1 15 33 AM

src/pages/DaoDashboard/Activities/index.tsx
Da-Colon: no sir? wdym?

src/pages/DaoDashboard/Activities/hooks/useActivities.ts
Da-Colon: same as above, this is just to prevent typescript from yelling. but its safe to assume that safe is defined here. I'm not sure why the address is typed to be undefined

src/pages/DaoDashboard/Activities/ActivityTreasury.tsx
Da-Colon: Its inheriting the styling from the parent

Da-Colon:
> the activity feed needs the same default left/right padding that the upper cards have (see screenshot)
There isn't padding here this is flex containers. I'll look into making this smooth together

Da-Colon:
> The Newest/Oldest filter opens SUPER slow on Firefox, but not on Chrome, for some reason. I actually think maybe it should just toggle between the two without needing to open the dialog to select one.

Don't use firefox....also toggle seems non intuitive IMO, plus we may want to add more sorts in the future?. I'll look at see what the problem on Fox is

src/pages/DaoDashboard/Activities/index.tsx
herbig: ah I was thinking this was padding of some sort, but it's the gap between components.

herbig:
> @herbig I'm not getting 'slow' behavior on firefox?

Hmmmm, working fine for me now as well. I'll keep an eye on it.

herbig (approved):
🛳️🛳️🛳️

src/pages/DaoDashboard/Activities/hooks/useActivities.ts
```diff
@@ -0,0 +1,157 @@
...
+      }
+      return dataB - dataA;
+    });
+  }, [parsedActivities, sortBy]);
+
+  /**
+   * When data is ready, set loading to false
+   */
+  useEffect(() => {
+    if (sortedActivities.length) {
```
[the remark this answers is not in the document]
Da-Colon: gosh good point. I'll get this updated

src/pages/DaoDashboard/Activities/ActivityTreasury.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { useTranslation } from 'react-i18next';
+import EtherscanLinkAddress from '../../../components/ui/EtherscanLinkAddress';
+import useDisplayName from '../../../hooks/useDisplayName';
+import { Activity } from '../../../types';
+import { AcitivityCard } from './ActivityCard';
+
+function ActivityAddress({ address, addComma }: { address: string; addComma: boolean }) {
+  const { displayName } = useDisplayName(address);
+  return (
+    <Text>
```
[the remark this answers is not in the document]
Da-Colon: yeah the newer design doesn't call for it.

src/components/ui/loaders/BarLoader.tsx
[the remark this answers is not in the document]
Da-Colon: totally preference. I think there is an argument for either way

PR #644, opened 2022-11-07
Token data

tbwebb22:
## Description

This PR:
1) Adds module contract objects to the modules array
2) Adds governance token data to the Gnosis Governance data. This also includes the token voting contract object and the Usul voting strategy object
3) Adds a temporary delegate page & route for testing purposes
4) Fixes bug where the predicted token address wasn't correct

## Notes

This adds a temporary route to the delegate page at /<dao address>/delegate
This route is for testing purposes so that we can delegate votes to enable votes casting and such.
This delegate component will likely be moved into a modal in a future PR, and won't actually have its own route
This PR does not include styling yet, this can be added in a future PR

## Testing
1) Create a Usul DAO, and allocate some of the governance tokens to your address
2) Once on the DAO home page, add `/delegate` to the end of the URL route
3) Try allocating tokens to your address and other addresses, and verify that the `delegatee` and `voting weight` update accordingly

image


src/providers/fractal/hooks/useGovernanceTokenData.ts
```diff
@@ -234,35 +303,23 @@ const useTokenData = (
       _previousBalance: BigNumber,
       currentBalance: BigNumber
     ) => {
-      dispatch({ type: TokenActions.UPDATE_VOTING_WEIGHT, payload: currentBalance });
+      dispatch({
+        type: TokenActions.UPDATE_VOTING_WEIGHTS,
+        payload: {
+          votingWeight: currentBalance,
+          votingWeightString: ethers.utils.formatUnits(currentBalance, decimals),
```
herbig, line 310: I added a function in the Treasury page to better format token strings with commas and cut non significant decimals, will probably be useful here https://github.com/decent-dao/fractal-interface/pull/642/files#r1016150767

src/providers/fractal/hooks/useGovernanceTokenData.ts
```diff
@@ -71,37 +78,124 @@ const reducer = (state: IGoveranceTokenData, action: TokenAction) => {
...
+    if (!usulModule || !signerOrProvider) {
+      return;
+    }
+
+    (async () => {
+      // todo: This assumes that only one strategy has been enabled, but Usul supports more than one
+      const votingContractAddress = await usulModule
+        .queryFilter(usulModule.filters.EnabledStrategy())
+        .then(strategiesEnabled => {
+          return strategiesEnabled[0].args.strategy;
```
herbig, line 140: this gets the first strategy set on the Usul mod? Is there any potential for abuse with us assuming to always use the first strategy?

herbig (approved):
Looks good.  @mudrila should check this out, as he's doing proposal work

src/providers/fractal/hooks/useGovernanceTokenData.ts
tbwebb22: Ah, nice. Just added that util and applied it to those formatted strings in latest commit

src/providers/fractal/hooks/useGovernanceTokenData.ts
tbwebb22: Just pushed a commit that updates this logic. This will now check that the voting contract was deployed with the master copy contract address that we expect, otherwise it sets that contract to undefined.

src/utils/numberFormats.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const coinFormatter = new Intl.NumberFormat(undefined, { maximumSignificantDigits: 18 });
+
+export const formatCoin = (rawBalance: BigNumber | string, decimals?: number, symbol?: string) => {
+  return symbol
+    ? coinFormatter.format(parseFloat(ethers.utils.formatUnits(rawBalance, decimals))) +
+        ' ' +
+        symbol
+    : coinFormatter.format(parseFloat(ethers.utils.formatEther(rawBalance))) + ' ETH';
+};
```
Da-Colon, line 24: should this be strictly ETH or any symbol?

src/utils/numberFormats.ts
```diff
@@ -0,0 +1,24 @@
...
+  style: 'currency',
+  currency: 'USD',
+});
+
+export const formatUSD = (rawUSD: number) => {
+  return usdFormatter.format(rawUSD);
+};
+
+const coinFormatter = new Intl.NumberFormat(undefined, { maximumSignificantDigits: 18 });
+
```
Da-Colon, line 17: Correctly me if I'm wrong but this shows the values as `10.0` without the `minimumFractionDigits` option set to `0`

Da-Colon (commented):
small comments, otherwise looks good

src/utils/numberFormats.ts
herbig: @tbwebb22 imported this from my Treasury PR, but yeah it will show `10` as `10`.

src/utils/numberFormats.ts
herbig: this is another one imported from the Treasury PR.

This checks if a symbol is provided, and if so formats it using the given decimals as well.

If a symbol is not provided, it formats it as ETH.

src/utils/numberFormats.ts
tbwebb22: Is this ok @Da-Colon or do you think a change is needed here?

Da-Colon: yeah I think its fine for now if its showing correctly in the UI. WE have 2 methods doing almost the same thing. we will get these all using the same method once this stuff is all merged

Da-Colon (approved):

PR #652, opened 2022-11-09
V1 | DAO Favoriting and Local Storage Update

### LINKED ISSUE #651: DAO Favoriting

Da-Colon:
## Description
Reenables DAO Favorites, while also doing an overall to local storage hooks and how this data is stored and used

### Fractal Provider
- Update Fractal Provider with new ConnectedAccount reducer and methods to store 'connected' user information. As we move forward this would be a good location to store user settings and other things we might want to cache for a user.
### Local Storage
- Speaking of caching. While updating the re-hooking up the favorites, I noticed that when favoriting it wasn't adding/removing from favorites setup. When Favorites was a page this wasn't a problem as toggling happened on a different page than the favorites were loaded.
- I removed the current `useLocalStorage` hook and have it run for only setting the initial state of a users' first visit. 
- the setting of storage items are now handled where the action happens while also allow favorites to be set to a state than the DOM can recognize changes to it.

### Favorites Hook
- Inspired by the conversation that @herbig and I had. wanted to create a hook that setup easy to use methods and data as an object in state. `useAccountFavorites` to have this almost feel like a class with methods. 

closes #651 

## Testing
Navigate to a DAO
Click on the favorited Star

expect:
- Star Fill should toggle gold and empty
- Favorited DAO should appear/disappear from favorites list in header


src/pages/DaoDashboard/Info/InfoDAO.tsx
```diff
@@ -9,13 +9,22 @@ export function InfoDAO() {
...
+  } = useFractal();
+
   // @todo replace mocked values
-  const MOCK_IS_FAVORITE = false;
   const MOCK_IS_SUB_DAO = false;
-
+  if (!safe.address) {
+    // @todo replace with a loader
+    return <div />;
+  }
```
Da-Colon, line 27: I imagine we will need loaders for each of the sections at the top. Once that Loader I made is merged and see what it looks like here. and do some focus where on how these sections look as the data loads

src/pages/DaoDashboard/Info/InfoDAO.tsx
```diff
@@ -30,7 +39,19 @@ export function InfoDAO() {
...
+          aria-label="Favorite Toggle"
+          icon={
+            isConnectedFavorited ? (
+              <StarGoldSolid boxSize="1.5rem" />
+            ) : (
+              <StarOutline boxSize="1.5rem" />
+            )
+          }
+          onClick={() => toggleFavorite(safe.address!)}
+        />
```
Da-Colon, line 54: Shouldn't put onClicks on Icons (per Chakra-Docs) they recommends this `IconWrapper` component. `variant="unstyled"` and `minWidth="0px"` to override defaults of 'Button' styling

src/providers/fractal/FractalProvider.tsx
```diff
@@ -38,14 +42,24 @@ export function FractalProvider({ children }: { children: ReactNode }) {
...
+  const [account, accountDispatch] = useReducer(
+    connectedAccountReducer,
+    connectedAccountInitialState,
+    initializeConnectedAccount
+  );
+
   // @todo update to handle new contracts
   // const daoLegacy: IDaoLegacy = useDAOLegacy(dao.daoAddress);
 
+  useLocalStorage();
```
Da-Colon, line 54: This setups localStorage on load. really this should only ever fire for a user once unless we update with a new cache item (see notes)

src/App.tsx
```diff
@@ -4,23 +4,25 @@ import Header from './components/ui/Header';
...
 
   useActionToast({
     toastId: 'audit:toast',
     testId: 'toast-audit',
-    isVisible: !notAuditedAcceptance,
+    isVisible: hasAccepted !== undefined && !hasAccepted,
     titleTranslationKey: 'auditDisclaimer',
     buttonTranslationKey: 'accept',
-    buttonOnClick: () => setNotAuditedAcceptance(true),
+    buttonOnClick: () => acceptAudit(),
```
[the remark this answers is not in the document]
Da-Colon: yeah opps, fixed

src/providers/fractal/hooks/account/useAccountFavorites.ts
```diff
@@ -0,0 +1,87 @@
...
+   */
+  const toggleFavorite = useCallback(
+    (address: string) => {
+      const normalizedAddress = ethers.utils.getAddress(address);
+      const rawFavorites = localStorage.getItem(CacheKeys.FAVORITES);
+
+      let updatedFavorites = [] as string[];
+      if (rawFavorites) {
+        const parsedFavorites = JSON.parse(rawFavorites);
+        if ([...favoritesList].includes(normalizedAddress)) {
```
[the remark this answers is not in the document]
Da-Colon: habit. updated

src/providers/fractal/hooks/account/useAccountFavorites.ts
```diff
@@ -0,0 +1,87 @@
...
+  const toggleFavorite = useCallback(
+    (address: string) => {
+      const normalizedAddress = ethers.utils.getAddress(address);
+      const rawFavorites = localStorage.getItem(CacheKeys.FAVORITES);
+
+      let updatedFavorites = [] as string[];
+      if (rawFavorites) {
+        const parsedFavorites = JSON.parse(rawFavorites);
+        if (favoritesList.includes(normalizedAddress)) {
+          updatedFavorites = favoritesList.filter(favorite => favorite !== normalizedAddress);
```
herbig, line 44: a map would probably be more efficient for favorites than a list, that way you don't ever need to iterate over the whole thing, you just add or delete the key, mapped to itself.

it probably doesn't matter until you get to a whole lot of favorites though, at which point this iteration may lag, but I have no idea how many that would take

src/providers/fractal/hooks/account/useAccountFavorites.ts
```diff
@@ -0,0 +1,87 @@
...
+          const newValue = JSON.stringify({
+            ...parsedFavorites,
+            [chainId]: updatedFavorites,
```
herbig, line 47: nice, were we doing this by chainId before, or is this new?

src/providers/fractal/hooks/account/useAccountFavorites.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
+      }
+      setFavorites(updatedFavorites);
+    },
+    [favoritesList, chainId]
+  );
+
+  const loadFavorites = useCallback(() => {
+    const rawFavorites = localStorage.getItem(CacheKeys.FAVORITES);
+    if (rawFavorites) {
```
herbig, line 66: since you set up `defaultValue` for cache keys, you shouldn't need to check here, you'd always have a value

src/providers/fractal/hooks/account/useAccountFavorites.ts
```diff
@@ -0,0 +1,87 @@
...
+   * if favorited it will remove from local storage and state
+   * if not favorited, given address will be saved to local storage with the give id
+   */
+  const toggleFavorite = useCallback(
+    (address: string) => {
+      const normalizedAddress = ethers.utils.getAddress(address);
+      const rawFavorites = localStorage.getItem(CacheKeys.FAVORITES);
+
+      let updatedFavorites = [] as string[];
+      if (rawFavorites) {
```
herbig, line 41: also with this check, this should always be true I believe

src/providers/fractal/hooks/account/useAccountAudit.ts
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useState, useCallback } from 'react';
+import { AccountAction } from './../../constants/actions';
+import { CacheKeys } from './useLocalStorage';
+
+interface IUseAccountAudit {
+  safeAddress?: string;
+  accountDispatch: any;
+}
+
+export const useAccountAudit = ({ accountDispatch }: IUseAccountAudit) => {
```
herbig, line 10: favorites is a special enough case to have its own hook probably, but I wonder if this could instead be a generalized `useLocalStorage` that takes a key to return?

herbig: otherwise we need a new hook for every local storage key we create

src/pages/DaoDashboard/Info/InfoDAO.tsx
herbig: yeah it might look a bit weird in all of these boxes, but we'll see

src/providers/fractal/hooks/account/useAccountFavorites.ts
Da-Colon: yeah we were doing it by chainId before.

herbig (approved):
some comments for discussion, but looks great

src/providers/fractal/hooks/account/useAccountFavorites.ts
Da-Colon: so this is partly for Typescript, partly because new users when they first load wont have these for a second. The Cached items are created quickly but I rather not risk problems for new users

src/providers/fractal/hooks/account/useAccountAudit.ts
Da-Colon: so..I thought through this, I built this with the thought of Caching in the future.

`useLocalStorage` is used to define are Cache items and defaultValues

I spread this into its own hook as it seemed out of place in the main `useAccount` hook we expand more data that will go here.

I think its a good pattern that we can update the current Governance and Gnosis hooks into to group everything together.

Issue #661, opened 2022-11-14
Populate Treasury Card in DAO homepage


herbig:
### Overview

For Usul and Safe

<img width="1411" alt="Screen Shot 2022-11-14 at 9 54 17 AM" src="https://user-images.githubusercontent.com/384559/201692205-0c4f982c-ae4e-4af5-bfc3-57afc75daf22.png">

### Link to documentation

https://www.figma.com/file/uRV3Nfowdzt5CknW1FgUkQ/%F0%9F%93%9A-Fractal-Design-System-%7C-Base-Components-%7C-October-2022?node-id=2158%3A12667&t=ztkrKXthwIf9Hvc0-0

PR #666, opened 2022-11-14
treasury info card

herbig:
## Description

Adds total USD to the DAO homepage

closes https://github.com/decent-dao/fractal-interface/issues/661

## Testing

View a DAO with Safe assets that are worth USD.

Screen Shot 2022-11-14 at 4 31 20 PM


src/pages/Treasury/hooks/useTreasuryTotalUSD.tsx
```diff
@@ -0,0 +1,11 @@
+import { useFractal } from '../../../providers/fractal/hooks/useFractal';
+import { formatUSD } from '../../../utils/numberFormats';
+
+export function useTreasuryTotalUSD(): string {
+  const {
+    treasury: { assetsFungible },
+  } = useFractal();
+  let totalUSD = 0;
+  assetsFungible.map(asset => (totalUSD += Number(asset.fiatBalance)));
```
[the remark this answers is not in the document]
herbig: Yeah good call. forEach makes a lot more sense

For reduce, it seemed like I need to return the same value at each iteration, which doesn't work with iterating through a list of `GnosisAssetFungible`

src/pages/DaoDashboard/Info/InfoTreasury.tsx
```diff
@@ -28,7 +28,7 @@ export function InfoTreasury({}: IDAOGovernance) {
         textStyle="text-lg-mono-semibold"
         color="grayscale.100"
       >
-        {MOCK_TREASURY_TITLE}
+        {useTreasuryTotalUSD()}
```
Da-Colon, line 31: hooks shouldn't be used in the return here. the hook should return a variable that can be used here. 

```ts
const totalUSD = useTreasuryTotalUSD()

return (
   ...
    <Flex
	@@ -28,7 +28,7 @@ export function InfoTreasury({}: IDAOGovernance) {
        textStyle="text-lg-mono-semibold"
        color="grayscale.100"
      >
      {totalUSD}
      ...
```

Da-Colon: https://reactjs.org/docs/hooks-custom.html

Da-Colon: basically when you call a function in the render like this, everytime the component 're-renders' this method is called again and calculated each time. if you set the it to state or `useMemo` it won't be recalculated unless something has changed.

herbig: @Da-Colon ah, that makes sense, thanks

Does the latest update work?

Da-Colon (approved):

src/pages/DaoDashboard/Info/InfoTreasury.tsx
Da-Colon: Works for me, sir 🚀

PR #670, opened 2022-11-16
adds retries to handle flaky tests

Da-Colon:
## Description
Quick fix: This will help with these flaky tests. I have an idea, that might help optimize these tests when there is time that we can talk about when the chance comes up.


Da-Colon:
Because of the viewports of the test browsers I don't think this is the case. 1280x1000 any toasts would appear far below the links. My guess here is that these are being flaky because of the nature of opening a new tab and it running multiple tests that open in new tabs. We may need to refactor these tests to probably run under the same test and just add a message to the `expect(testParam, 'should log in')`. This current pattern is my fault I think from my experience with Pupeteer, but I think using the expect's second arugment to make clear what fails is the better pattern for Playwright

herbig (approved):

PR #675, opened 2022-11-17
Refactor | Add Safe SDK for Transaction Service API

Da-Colon:
## Description

Adds Safe's Transaction Service Client SDK. This should allow us to easily interact with the transaction-service api. 

## Notes
- Made a few updates to the typings for Treasury. 
- Got rid of our typings for the responses
- Had some webpack errors so I had to update the overrides and add packages 'assert' and 'stream-browserfiy'
- Replaced the API requests in 'useGnosisApiServices` and updated the action types of the treasuryDispatch


src/providers/fractal/hooks/useGnosisApiServices.ts
```diff
@@ -18,63 +15,82 @@ import { TreasuryAction } from './../constants/actions';
...
-    state: { chainId },
+    state: { chainId, account },
   } = useWeb3Provider();
 
+  const {
+    state: { signerOrProvider },
+  } = useWeb3Provider();
+
+  useEffect(() => {
+    if (!account || !signerOrProvider) {
```
herbig, line 31: @Da-Colon did you find any solution to not requiring a signer here at some point?

Da-Colon: @herbigo sorry thought I mentioned this. I found this yesterday. https://github.com/safe-global/safe-core-sdk/issues/258. 

TL:DR its a WIP coming soon. need to track this as we wouldn't have to update much as it lines up with our setup

herbig: Ah ok, cool that's good to know we should have an option to do this in the future.

Issue #679, opened 2022-11-17
Bug: Eth appears in treasury even though there's 0 balance


herbig:
### Overview

Apparently the Gnosis API returns an Eth record even if there is none.  We should remove this.

### Steps to reproduce

Create a new DAO, then view the Treasury.

### Screenshots / Video

<img width="587" alt="Screen Shot 2022-11-17 at 5 10 40 PM" src="https://user-images.githubusercontent.com/384559/202570963-82c73432-966a-4935-a38e-ec1b766daea7.png">

### Version

Dev (app.dev.fractalframework.xyz/)

### Browser

All

PR #690, opened 2022-11-18
Treasury bug

herbig:
## Description

Fixes a bug where a line for Eth appears in the treasury when there is zero balance.

This is because the Eth asset is returned from the Gnosis API regardless of a zero amount.

Closes https://github.com/decent-dao/fractal-interface/issues/679

## Testing

Create a new DAO and navigate to the treasury.

There should be no line item for $0 Eth in the treasury assets.

### DIFF
--- a/src/i18n/locales/en/sidebar.json
+++ b/src/i18n/locales/en/sidebar.json
@@ -3,7 +3,7 @@
   "proposals": "Proposals",
   "nodes": "Nodes",
   "treasury": "Treasury",
-  "support": "Support",
+  "faq": "FAQ",
   "documentation": "Documentation",
   "ariaLabelFractalBrand": "Navigate Fractal Home",
   "ariaLabelHome": "Navigate DAO Home",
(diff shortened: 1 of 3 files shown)


src/i18n/locales/en/sidebar.json
```diff
@@ -3,7 +3,7 @@
   "proposals": "Proposals",
   "nodes": "Nodes",
   "treasury": "Treasury",
-  "support": "Support",
+  "faq": "FAQ",
```
herbig, line 6: @Da-Colon I threw this change in here as well, I think the left sidebar should tooltip this as FAQ, to be consistent with the "FAQ" link we have on the dashboard page as well, since it links to the FAQ and uses the FAQ question mark icon.

src/pages/Treasury/hooks/useFormatCoins.tsx
```diff
@@ -16,11 +16,12 @@ export interface TokenDisplayData {
 
 export function useFormatCoins(assets: GnosisAssetFungible[]) {
   let totalFiatValue = 0;
-  let displayData: TokenDisplayData[] = new Array(assets.length);
+  let displayData: TokenDisplayData[] = [];
```
herbig, line 19: the array is dynamic rather than specifying the initial size, since it could be less if eth is zero

PR #692, opened 2022-11-19
precision percentages with BigNumbers

herbig:
## Description

Precision percentages using BigNumbers.

The issue is that BigNumbers are integers, so division using them will always lose precision, e.g. 1/2 = 0.

This PR uses the bigDecimal library to handle dividing BigNumbers, as well as adjusts the percent formatter method to take string, number, or BigDecimal.

## Testing

I used @mudrila 's test DAO at http://localhost:3000/#/daos/0xeA4b1Ba101ed338BC72a35990a6c962c852Bc075/proposals/0 to view the changes

Screen Shot 2022-11-18 at 10 43 46 PM

### DIFF
--- a/src/components/Proposals/ProposalVotes.tsx
+++ b/src/components/Proposals/ProposalVotes.tsx
@@ -11,7 +11,7 @@ import { BigNumber } from 'ethers';
 import { useTranslation } from 'react-i18next';
 import useDisplayName from '../../hooks/useDisplayName';
 import { Proposal, ProposalVote } from '../../providers/fractal/types';
-import { formatCoin } from '../../utils/numberFormats';
+import { formatCoin, formatPercentage } from '../../utils/numberFormats';
 import ContentBox from '../ui/ContentBox';
 import ProgressBar from '../ui/ProgressBar';
 import StatusBox from '../ui/StatusBox';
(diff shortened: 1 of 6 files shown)


src/components/Proposals/ProposalVotes.tsx
```diff
@@ -21,52 +21,52 @@ const MOCK_VOTES: ProposalVote[] = [
   {
     voter: '0x6a0db4cef1ce2a5f81c8e6322862439f71aca29d',
     choice: 'no',
-    weight: BigNumber.from(150),
+    weight: BigNumber.from('150000000000000000000'),
```
herbig, line 24: @mudrila these were appearing in the UI as 0 FRACT because the "formatCoin" method takes a full raw token amount, so in wei in the case of Eth.  150 here basically zero.

src/pages/ProposalDetails/index.tsx
```diff
@@ -19,7 +19,8 @@ import useProposals from '../../providers/fractal/hooks/useProposals';
 import { Proposal } from '../../providers/fractal/types';
 import { DAO_ROUTES } from '../../routes/constants';
 
-const MOCK_GOV_TOKEN_TOTAL_SUPPLY = BigNumber.from('500000000000000000000000000');
+const MOCK_GOV_TOKEN_TOTAL_SUPPLY = BigNumber.from('3475000000000000000000');
```
herbig, line 22: Adjusted this to be the actual sum of the 10 mock tokens, so the percentages would be correct.

src/pages/ProposalDetails/index.tsx
```diff
@@ -19,7 +19,8 @@ import useProposals from '../../providers/fractal/hooks/useProposals';
...
+const MOCK_GOV_TOKEN_DECIMALS = 18;
```
herbig, line 23: we'll need to pass this in, since not all tokens have 18 decimals

PR #711, opened 2022-11-28
Update | Fractal-UI Release v0.1

Da-Colon:
## Description

Based off #708 

Paired with the release of `@decent-org/fractal-ui` release v0.1. [Release PR](https://github.com/decent-dao/decent-ui/pull/63).

This PR combs through the app removing custom component imports, replacing them with the components imported directly from Chakra-UI library.

In future release more custom `InputGroup` and `NumberInput` Implementations will be possible. but the current variations of where the props need to be passed is quite tricky. to combine these two for all use cases. 

See Release PR above for details on theme updates details
- Fixes `Input` Theme (defaults now matches figma properly)
- Fixes `Button` Theme (fontWeight was being overridden by Chakra-UI default)
- Update `Progress` theme to match documentation
- Adds `NumberInput` theme

## Notes
- Imports `Button`, `Input` components from `@chakra-ui/react`
- Replaces number `Input` components with `NumberInput`
- Create `useFormHelpers` hook to handle decimals on `NumberInput`
- Uses `InputGroup` for Left and Right Element Add-ons

## Testing
Comb through App
- Text `Input` should match figma
- `Input` units (addons) should appear correctly
- `NumberInputs` should be decimals locked where appropriate
- `NumberInputs` should not allow odd characters
- Button should match Figma now


src/hooks/utils/useFormHelpers.tsx
```diff
@@ -0,0 +1,19 @@
...
+      if (!value || !event.key.match('[0-9]')) {
+        return event;
+      }
+      const [, dec] = value.toString().split('.');
+      if (!!dec && !!maxDecimals && dec.length >= maxDecimals) {
+        event.preventDefault();
+      }
+    },
+    []
+  );
```
Da-Colon, line 16: Slightly altered but similar implementation of `limitDecimals` method in the `Input` component. 

Based on a note from @herbig I've removed the hardcoded 'Backspace' and 'Delete' match with a simpler if it isn't a number return the event.

The `NumberInput` components handles all the other restricted characters for us

src/components/DaoCreator/SubsidiaryFunding/index.tsx
```diff
@@ -175,17 +184,30 @@ export function SubsidiaryFunding() {
...
+                  <InputRightElement>
+                    <Button
+                      variant="text"
+                      onClick={() => maxFundToken(index)}
+                    >
+                      Max
+                    </Button>
+                  </InputRightElement>
+                </InputGroup>
+              </NumberInput>
```
Da-Colon, line 210: This is a good example of How to use the `NumberInput` and `InputGroup` together.

NumberInput: https://chakra-ui.com/docs/components/number-input/usage
Input (InputGroup | down the page): https://chakra-ui.com/docs/components/input/usage

`InputGroup` is needed to add elements within the Input Field. Normally this just wraps an Input (see docs).

Notice that you have to wrap the `NumberInputField` in the `InputGroup` rather than the `NumberInput` as a whole. This will allow the added element to properly positioned

I'm working on the padding when there is add-ons added. Its a work in progress but `baseLeftAddon`, `baseWithAddons`, `xlRightAddon`, etc; can be used as the size property in the `NumberInput` or `Input` fields.

mobiusTripper-crypto (approved):

src/components/DaoCreator/SubsidiaryFunding/index.tsx
herbig: @decent-dao/fractal

src/components/DaoCreator/gnosis/GnosisConfig.tsx
```diff
@@ -83,31 +83,27 @@ export function GnosisConfig() {
           subLabel={t('helperSigThreshold')}
           errorMessage={thresholdError}
         >
-          <Input
-            data-testid="gnosisConfig-thresholdInput"
-            size="base"
-            type="text"
+          <NumberInput
             value={signatureThreshold}
```
herbig, line 87: signature threshold still allows decimals, so should probably have precision 0 I believe?

`numberOfSigners` doesn't allow decimals, but I suspect that's because the `handleSignersChanges` is doing something to prevent it, rather than the number input itself.

herbig: @Da-Colon I think we should remove whatever the onChange is doing, and allow the user to delete everything in the field (currently deleting replaces it with 0), and let the numberinput precision handle everything

src/components/DaoCreator/tokenVotingGovernance/TokenDetails.tsx
```diff
@@ -71,15 +71,14 @@ function TokenDetails() {
           label={t('labelTokenSupply')}
           subLabel={t('helperTokenSupply')}
         >
-          <Input
-            size="base"
-            type="number"
+          <NumberInput
             data-testid="tokenVoting-tokenSupplyInput"
             value={govToken.tokenSupply.value}
```
herbig, line 76: this now limits to only 4 decimal places, rather than 18

src/components/DaoCreator/tokenVotingGovernance/GovernanceDetails.tsx
```diff
@@ -56,16 +62,18 @@ function GovernanceDetails() {
             label={t('labelVotingPeriod', { ns: 'daoCreate' })}
             subLabel={t('helperVotingPeriod', { ns: 'daoCreate' })}
           >
-            <Input
-              size="base"
-              type="number"
+            <NumberInput
               value={govModule.votingPeriod.toString()}
```
herbig, line 66: it's impossible to delete the first number here (defaults to 6), so you have to like, highlight the whole thing then type a number to get anything not starting with 6

herbig: Same with the other fields, though now I'm realizing that was the case before as well.

Is there some way we can allow them to delete everything in the field here?

src/components/DaoCreator/tokenVotingGovernance/TokenDetails.tsx
Da-Colon: opps I left in the a testing variable. should be fixed now.

```ts
          onKeyDown={e =>
            limitDecimalsOnKeyDown(e, tokenAllocation.amount.value, DEFAULT_TOKEN_DECIMALS)
          }
```

src/components/DaoCreator/tokenVotingGovernance/GovernanceDetails.tsx
Da-Colon: yeah it was like that before. I believe the field is probably not allowed to be undefined (blank field). This can be made into an issue and fixed in another PR?

src/components/DaoCreator/gnosis/GnosisConfig.tsx
Da-Colon: yeah I'll branch off from this PR and fix this and the below comment's issue.

herbig (approved):

src/components/DaoCreator/gnosis/GnosisConfig.tsx
mobiusTripper-crypto: it also allows _+Ee ... maybe using something like this

`isValidCharacter={(v: string) => /^[0-9]$/.test(v)}` on <NumberInput> 
 
or in the helper method

Da-Colon: blah. I could have sworn NumberInput did this. I added `restrictChars` method that is used by the `limitDecimalsOnKeyDown` method and can also be applied to any other Input seperately

PR #714, opened 2022-11-29
Feature/671/add guard

### LINKED ISSUE #671: Add Veto Guard to subDAO creation (Safe/Usul)

christopherdancy:
## Description

This PR does the following:

- Attaches Fractal, VetoGuard, VetoVoting to a gnosis safe
- There are 4 noble paths - Root Gnosis, Sub Gnosis, Root Usul, Sub Usul
- Updated contract addresses


src/hooks/useAddresses.ts
```diff
@@ -0,0 +1,232 @@
...
+    timelock?: EthAddress;
+    claimModule?: EthAddress;
+    gnosisWrapper?: EthAddress;
+    gnosisSafe?: EthAddress;
+    multiSend?: EthAddress;
+    fractalModuleMasterCopy?: EthAddress;
+    fractalNameRegistry?: EthAddress;
+    vetoGuardMasterCopy?: EthAddress;
+    vetoMultisigVotingMasterCopy?: EthAddress;
+    vetoERC20VotingMasterCopy?: EthAddress;
```
Da-Colon, line 32: This file can be removed.

src/pages/SubDaoCreate/index.tsx
```diff
@@ -1,6 +1,7 @@
 import { useNavigate } from 'react-router-dom';
 import DaoCreator from '../../components/DaoCreator';
 import { GnosisDAO } from '../../components/DaoCreator/provider/types';
+// eslint-disable-next-line import/namespace, import/default, import/no-named-as-default
```
Da-Colon, line 4: nit picky, but might be better to update this hook to an export versus having it default exported and having to add this.

Da-Colon (dismissed):
How can I test that I have successfully added a guard?

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -137,15 +153,194 @@ const useBuildDAOTx = () => {
...
+            ethers.utils.defaultAbiCoder.encode(
+              ['address', 'address', 'address', 'address[]'],
+              [
+                parentDAOAddress ? parentDAOAddress : safeContract.address, // Owner -- Parent DAO
+                safeContract.address, // Avatar
+                safeContract.address, // Target
+                [], // Authorized Controllers
+              ]
+            ),
+          ]);
```
sethhrbek, line 169: This code fragment could be refactored to its own method, since it is used in both DAO types (multisig + usul).

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -56,7 +66,7 @@ const useBuildDAOTx = () => {
                   ...gnosisDaoData.trustedAddresses.map(trustedAddess => trustedAddess.address),
                   multiSendContract.address,
                 ],
-            hasUsul ? 1 : gnosisDaoData.trustedAddresses.length, // threshold
+            1, // Threshold
```
herbig, line 69: what's this parameter doing, and why did that change to 1 in both cases?

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -137,15 +153,194 @@ const useBuildDAOTx = () => {
...
+            ),
+          ]);
+
+        const fractalByteCodeLinear =
+          '0x602d8060093d393df3363d3d373d3d3d363d73' +
+          fractalModuleMasterCopyContract.address.slice(2) +
+          '5af43d82803e903d91602b57fd5bf3';
+        const fractalSalt = solidityKeccak256(
+          ['bytes32', 'uint256'],
+          [solidityKeccak256(['bytes'], [setModuleCalldata]), saltNum]
```
herbig, line 177: does this `saltNum` need to match where it was already used above? if not it probably makes sense to generate a new one

herbig: Same with the uses of `saltNum` below.  Do those all need to match or should we have a new random salt for each?

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -137,15 +153,194 @@ const useBuildDAOTx = () => {
...
-        const internaltTxs: MetaTransaction[] = [
-          buildContractCall(
-            fractalNameRegistryContract,
-            'updateDAOName',
-            [gnosisDaoData.daoName],
-            0,
-            false
-          ),
-        ];
+        // Fractal Module
```
herbig, line 156: @tbwebb22 it would be good to have you review this portion as well, as you and @christopherdancy are the contract experts

src/hooks/useAddresses.ts
christopherdancy: For sure -- will remove this now

src/hooks/DAO/useBuildDAOTx.ts
christopherdancy: Great question -- the previous method would cause the tx to revert for the following reason:

1. A gnosis safe requires the threshold be met before a tx may be executed.
2. If there is more than 1... set as the threshold then the multisend contract cannot act as the only signer.
3. Since the trusted addresses could not be -- the signer which is executing the tx, then they would not be able to sign for this tx.

therefore this should be set as one and changed after the multisend contract has completed its setup txs.

src/hooks/DAO/useBuildDAOTx.ts
christopherdancy: Makes sense -- I am updating as we speak.

christopherdancy: well not as we speak but right after.

herbig (approved):
This looks good to me, would just like @tbwebb22 to weigh in as well

src/hooks/DAO/useBuildDAOTx.ts
tbwebb22: Cool, will give this a review today

tbwebb22 (changes requested):
Why was `votesTokenMasterCopy` changed to `votesMasterCopy` everywhere?

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -137,15 +203,171 @@ const useBuildDAOTx = () => {
...
+          parentDAOAddress,
+          safeContract
+        );
+
+        let internaltTxs: MetaTransaction[];
+        if (
+          parentDAOAddress &&
+          vetoGuardMasterCopyContract &&
+          vetoMultisigVotingMasterCopyContract
+        ) {
```
tbwebb22, line 217: The purpose of this if / else block is to select between whether the DAO is a root or a child, correct? Why does it include the `vetoGuardMasterCopyContract` and `vetoMultisigVotingMasterCopyContract`?

christopherdancy:
> Why was `votesTokenMasterCopy` changed to `votesMasterCopy` everywhere?

I am not sure -- this was not m

Da-Colon:
hit enter too quick. Updates look good to me! lol

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -137,15 +203,171 @@ const useBuildDAOTx = () => {
...
+          const vetoMulitVotingSalt = solidityKeccak256(
+            ['bytes32', 'uint256'],
+            [solidityKeccak256(['bytes'], [setVetoMultiVotingCalldata]), saltNum]
+          );
+          const predictedVetoMultisigAddress = getCreate2Address(
+            zodiacModuleProxyFactoryContract.address,
+            vetoMulitVotingSalt,
+            solidityKeccak256(['bytes'], [vetoMultisigByteCodeLinear])
+          );
+          const vetoMultiContract = VetoMultisigVoting__factory.connect(
```
tbwebb22, line 235: Can we rename this one to `vetoMultisigVotingContract`? Just getting confusing with there also be the `vetoGuard` contract

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -137,15 +203,171 @@ const useBuildDAOTx = () => {
...
+          const vetoSalt = solidityKeccak256(
+            ['bytes32', 'uint256'],
+            [solidityKeccak256(['bytes'], [setVetoGuardCalldata]), saltNum]
+          );
+          const predictedVetoModuleAddress = getCreate2Address(
+            zodiacModuleProxyFactoryContract.address,
+            vetoSalt,
+            solidityKeccak256(['bytes'], [vetoByteCodeLinear])
+          );
+
```
tbwebb22, line 267: I think we should be more explicit with variable naming here. We have VetoVoting and VetoGuard contracts, so just naming something "Veto" doesn't really identify it well

tbwebb22:
I think we should be a little more explicit with the variable naming, more like:
`multisigVetoGuard`
`usulVetoGuard`
`multisigVetoVoting`
`erc20VetoVoting`

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -337,37 +577,193 @@ const useBuildDAOTx = () => {
...
-          ),
-        ];
+        // Fractal Module
+        const { predictedFractalModuleAddress, setModuleCalldata } = buildFractalModuleData(
+          parentDAOAddress,
+          safeContract
+        );
+
+        let internaltTxs: MetaTransaction[];
+        if (parentDAOAddress && vetoERC20VotingMasterCopyContract && vetoGuardMasterCopyContract) {
```
tbwebb22, line 587: Same question as the other if statement - why are we checking these two contracts here?

src/hooks/DAO/useBuildDAOTx.ts
christopherdancy: Due to the possible undefined state of these two contracts. They are not required if a user wanted to deploy a root gnosis safe

src/hooks/DAO/useBuildDAOTx.ts
christopherdancy: same answer as above.

src/hooks/DAO/useBuildDAOTx.ts
tbwebb22: So if I'm trying to create a subDAO, but one of those master copy contracts isn't defined, then it will still create a DAO tx, but it won't have any of the necessary child DAO configuration (veto guard, fractal module, etc). I don't think that's how we want this to work right?

christopherdancy: This is not for a sub dao. The same method is used for a root dao or a sub dao. The veto contract is only needed for the subdao. Therefore it is okay, during root dao creation, to not have one of the mastercopy contract defined.

mobiusTripper-crypto (approved):

src/hooks/DAO/useBuildDAOTx.ts
tbwebb22: The conditional in the if statement I have highlighted above, is supposed to only evaluate to true when we are creating a subDAO, no?

christopherdancy: It will only be true if there is a subdao correct.

tbwebb22: Right, so if I'm trying to create a subDAO, but one of these master copy contract is undefined, this conditional evaluates to false, and we go to the `else` block, and end up still creating a DAO tx, but that DAO will be configured as a root DAO, not a subDAO. I don't believe that's how we want this to work, right?

christopherdancy: I see what you are saying - just made changes.

tbwebb22: Hmm I don't see the changes

christopherdancy: I see where the issue was -- I updated this on the other branch. Should be here now.

tbwebb22 (approved):
LGTM

Issue #701, opened 2022-11-23
Bug: Limit Safe Signers input


herbig:
### Overview

The Safe signers input allows any number, but at about 10000 the UI will freeze as it attempts to render 10,000 components.

It's unreasonable to have even 1,000 signers added via this UI, so we should limit it to prevent issues.

### Steps to reproduce

Input 10000 in the "How many trusted users for Gnosis Safe" input in Safe DAO creation

### Version

Dev (app.dev.fractalframework.xyz/)

### Browser

All

PR #716, opened 2022-11-30
Misc bug fixes

herbig:
## Description

- adds bottom margin to the app main content
- fixes treasury "days ago" text that changed when proposals text changed it
- prevents rendering too many inputs in the Safe DAO creation page

closes https://github.com/decent-dao/fractal-interface/issues/701

Before
Screenshot 2022-11-29 at 6 59 13 PM
Screenshot 2022-11-29 at 6 59 54 PM

After
Screenshot 2022-11-29 at 6 59 37 PM
Screenshot 2022-11-29 at 6 59 47 PM

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -61,6 +61,7 @@ function App() {
           maxWidth="container.xl"
           mt={6}
           px="0"
+          paddingBottom="2rem"
           mx={[6, null, null, null, null, 'auto']}
         >
           <FractalRoutes />
--- a/src/components/DaoCreator/gnosis/GnosisConfig.tsx
+++ b/src/components/DaoCreator/gnosis/GnosisConfig.tsx
@@ -30,6 +30,11 @@ export function GnosisConfig() {
 
   const handleSignersChanges = (event: ChangeEvent<HTMLInputElement>) => {
     let numOfSigners = Number(event.target.value);
+
+    // don't let the initial set of signers to be greater than 999,
+    // since rendering that many inputs can freeze the browser
+    if (numOfSigners > 999) return;
+
     if (trustedAddresses.length !== numOfSigners) {
       const gnosisAddresses = [...trustedAddresses];
       const trustedAddressLength = trustedAddresses.length;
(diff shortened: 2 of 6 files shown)


src/App.tsx
```diff
@@ -61,6 +61,7 @@ function App() {
           maxWidth="container.xl"
           mt={6}
           px="0"
+          paddingBottom="2rem"
```
mobiusTripper-crypto, line 64: I'm pretty sure that's equivalent to `pb={8)` in chrakra? Do we want to standardize on a spacing styel for padding and margin? `24px` or `2rem` or the chakra `{8}`? or is it not that important?

the pb is a minor point but i would probably not use paddingBottom on a charka component.

src/components/DaoCreator/gnosis/GnosisConfig.tsx
```diff
@@ -30,6 +30,11 @@ export function GnosisConfig() {
 
   const handleSignersChanges = (event: ChangeEvent<HTMLInputElement>) => {
     let numOfSigners = Number(event.target.value);
+
+    // don't let the initial set of signers to be greater than 999,
+    // since rendering that many inputs can freeze the browser
+    if (numOfSigners > 999) return;
```
mobiusTripper-crypto, line 36: Thanks! that happened to me the 1st time i used the app.

mobiusTripper-crypto (approved):
Minor question let in a comment

src/App.tsx
herbig: @mobiusTripper-crypto what should I use for chakra instead of paddingBottom?

src/components/DaoCreator/gnosis/GnosisConfig.tsx
sethhrbek: If we aren't supporting > 999,  could this instead be set as the max number allowed on the input itself via something like `max={999}`

sethhrbek (approved):

src/components/DaoCreator/gnosis/GnosisConfig.tsx
herbig: yeah, that's a better solution. @Da-Colon can we add max{int} to number inputs and maxLength{int} to text inputs in the design system?

src/App.tsx
herbig: hmm, yeah I don't believe we've decided on a standard, so I'll leave it for now.

@Da-Colon maybe we should discuss a standard doc for this stuff

PR #719, opened 2022-11-30
Feature | DAO Dashboard Multisig Transactions

Da-Colon:
## Description

Adds Multisig Transactions to Dashboard.

## Notes
There is still work to be done to add USUL proposals to Dashboard.

## Testing
Load in any Multisig DAO

localhost_3000_(Desktop) (7)


src/pages/DaoDashboard/Activities/hooks/useActivities.ts
```diff
@@ -1,132 +1,118 @@
...
+import { useEffect, useMemo, useState } from 'react';
+import { BadgeLabels } from '../../../../components/ui/badges/Badge';
 import { useFractal } from '../../../../providers/Fractal/hooks/useFractal';
-import { GnosisTransactionsResponse } from '../../../../providers/Fractal/types';
-import { buildGnosisApiUrl } from '../../../../providers/Fractal/utils';
-import { useWeb3Provider } from '../../../../providers/Web3Data/hooks/useWeb3Provider';
-import { ActivityEventType, GnosisTransferType, SortBy } from '../../../../types';
+import { ActivityEventType, SortBy } from '../../../../types';
 import { formatWeiToValue } from '../../../../utils';
+import { useActivityParser } from './useActivityParser';
```
Da-Colon, line 11: extracted some of the larger logic into a separate file so that this file can remain fairly readable and short

src/pages/DaoDashboard/Activities/hooks/useActivities.ts
```diff
@@ -1,132 +1,118 @@
...
-    return transactions.results
-      .filter(t => !!t.transfers.length)
-      .map(transaction => {
-        const isDeposit = transaction.transfers.every(
-          t => t.to.toLowerCase() === safe.address!.toLowerCase()
-        );
+    return transactions.results.map((transaction, i, transactionArr) => {
+      const isMultiSigTransaction = transaction.txType === 'MULTISIG_TRANSACTION';
+      const multiSigTransaction = transaction as SafeMultisigTransactionWithTransfersResponse;
+      const ethereumTransaction = transaction as EthereumTxWithTransfersResponse;
```
Da-Colon, line 28: Makes it a little easier to extract typing instead of `as` everywhere

src/pages/DaoDashboard/Activities/hooks/useActivityParser.ts
```diff
@@ -0,0 +1,79 @@
+import {
+  TransferWithTokenInfoResponse,
+  SafeMultisigTransactionWithTransfersResponse,
+} from '@gnosis.pm/safe-service-client';
+import { constants, BigNumber } from 'ethers';
+import { useCallback } from 'react';
+import { GnosisTransferType } from '../../../../types';
+export const useActivityParser = () => {
+  const totalsReducer = useCallback((prev: Map<any, any>, cur: TransferWithTokenInfoResponse) => {
```
Da-Colon, line 9: This method is same as before just extracted out of `useActivities`

src/pages/DaoDashboard/Activities/hooks/useActivityParser.ts
```diff
@@ -0,0 +1,79 @@
...
+        return parseTransactions(dataDecoded);
+      });
+    };
+
+    const dataDecoded = multiSigTransaction.dataDecoded as any;
+    if (dataDecoded && isMultiSigTransaction) {
+      parseTransactions(dataDecoded.parameters);
+    }
+    return eventTransactionMap;
+  };
```
Da-Colon, line 77: Recursion of the win. So TL:DR of this method is that is looks through the parameters and their parameters for 'to' fields for each target it counts it as a single transactions. We use the size of the mapping to determine how many transactions took place in the multisig transaction

src/components/ui/badges/Badge.tsx
```diff
@@ -4,14 +4,24 @@ import { useTranslation } from 'react-i18next';
 
 type BadgeType = { [key: string]: { Icon: any; bg: string; color: string } };
 
+export enum BadgeLabels {
```
mobiusTripper-crypto, line 7: Are their benefits to labeling these with strings (maybe when debugging?) as opposed to just letting them be auto numbered? (You don't have to answer that here as I'm just curious as I don't really know. I'm still influenced by C# enums and maybe typescript enums are different)

src/pages/DaoDashboard/Activities/ActivityDescription.tsx
```diff
@@ -0,0 +1,60 @@
+import { Flex, Text } from '@chakra-ui/react';
+import { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+import { Activity } from '../../../types';
+import { ActivityAddress } from './ActivityAddress';
+
+function TransferDescription({ asset }: { asset: Activity }) {
+  const { t } = useTranslation();
+
+  const transferTypeStr = useMemo(() => {
```
mobiusTripper-crypto, line 10: I generally have no idea when to use useMemo, but I was under the impression that generally when returning a privative, useMemo should not be used. or in this case, should we be memorizing our use of `t`. More just wondering about this case, not really a question.

mobiusTripper-crypto (approved):

src/components/ui/badges/Badge.tsx
Da-Colon: yessir. The strings here have 2 purposes. Allows the use of this ENUM as a key to index an object. Also the string that its holding is also the label key for translations

src/pages/DaoDashboard/Activities/ActivityDescription.tsx
Da-Colon: err probably more of a habit of mine and in this case I can switch to just a enum. but don't like global values that use something from state not being wrapped. but I know it most cases this is overkill. I'll update these.

src/providers/Fractal/hooks/useGnosisGovernance.ts
```diff
@@ -13,7 +13,7 @@ export const useGnosisGovernance = (
     state: { account },
   } = useWeb3Provider();
 
-  const { deploySubDao, deploySubDAOPending } = useSubDAODeploy();
+  // const { deploySubDao, deploySubDAOPending } = useSubDAODeploy();
```
herbig, line 16: are these commented out because they're coming back at some point?

Da-Colon: I believe it should. I had originally had the idea that all things would just be pulled out of state and handled by the fractal provider versus in any of the UI components. They would only need to call 

```ts
const { governance: { deploySubDao }  } = useFractal() 
```

without any extra logic concerning itself of what kinda of Governance

sethhrbek: recommend putting a `@todo` whenever commenting something out so it doesn't accidentally get treated as dead code.

src/pages/DaoDashboard/Activities/hooks/useActivities.ts
```diff
@@ -1,132 +1,118 @@
...
+      const eventNonce = multiSigTransaction.nonce;
+
+      // Check to see if a proposal has been successfully executed to reject current transaction
+      const isRejected = transactionArr.find(tx => {
+        const multiSigTx = tx as SafeMultisigTransactionWithTransfersResponse;
+        return (
+          multiSigTx.nonce === eventNonce &&
+          multiSigTx.safeTxHash !== multiSigTransaction.safeTxHash &&
+          multiSigTx.isExecuted
+        );
```
sethhrbek, line 85: Is this the only place this logic will be used? Wondering if there should be both a proposal state / proposal TX state utility which can return the state of each of them. This may already exist (or not be necessary), but wanted to ask.

sethhrbek (approved):

src/pages/DaoDashboard/Activities/ActivityDescription.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+function TransferDescription({ asset }: { asset: Activity }) {
+  const { t } = useTranslation();
+
+  const transferTypeStr = asset.isDeposit ? t('receive') : t('send');
+  const transferDirStr = asset.isDeposit ? t('from') : t('to');
+
+  if (!asset.transaction.transfers.length) {
+    return null;
+  }
```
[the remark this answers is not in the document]
Da-Colon: fair

src/pages/DaoDashboard/Activities/hooks/useActivities.ts
sethhrbek: Oh no I didn't mean in a single method -- just meant having some specific file with multiple functions for looking up TX states / proposal states, depending on DAO type. But if this is a specific case just for this place in the code and most states are just like `<daoTx>.state` and not derived using logic, then this is probably over engineering and can be ignored.

Da-Colon: ah I see what your are saying. heard.

PR #732, opened 2022-12-02
Misc Bug Fixes

herbig:
## Description

- moved DAO creation strings to json rather than hardcoding
- updated Sentry version to the latest
- Changed hover text from 'Support' to 'FAQ' in left sidebar, to match where it links to and the text in the splash page
- Removed references to unsupported chains 'Rinkeby' and 'Sepolia'
- Changed toast message to say chain name instead of id
- Updated USD formatting to not remove trailing zero while still not showing decimals if they aren't needed


Da-Colon (approved):

mobiusTripper-crypto (approved):
looks good. looking at the formatUSD code. eventually maybe a more sophisticated number formatting library like 'react-number-format'

tbwebb22 (approved):
LGTM

herbig:
> looks good. looking at the formatUSD code. eventually maybe a more sophisticated number formatting library like 'react-number-format'

yeah, I couldn't find a better way to do exactly what I wanted with this.. which is why I wrote the spec test to make sure it worked properly

PR #733, opened 2022-12-05
Language Switcher

herbig:
## Description

This adds the language picker component to the UI.

## Notes
Logic has been included to not show the language picker in production if there aren't any other languages supported.

https://www.figma.com/file/uRV3Nfowdzt5CknW1FgUkQ/%F0%9F%93%9A-Fractal-%7C-Design-System?node-id=4150%3A23217&t=UtKHwiGziJtArvgj-0

## Testing

You should be able to see the three test options, but since we don't have translations for anything yet, they won't do anything.

Testing should just include visual confirmation that the component otherwise appears and operates as expected.

Screenshot 2022-12-07 at 3 59 04 PM


src/App.tsx
```diff
@@ -42,41 +42,40 @@ function App() {
...
+        area={'nav'}
+        display="flex"
+        flexDirection="column"
+        flexGrow="1"
+        bg="chocolate.900"
+        position="fixed"
+        w="4.25rem"
+        minHeight="100vh"
+      >
+        <Sidebar />
```
herbig, line 78: These changes were done by @mobiusTripper-crypto, in order to fix z ordering of components, where we saw the picker appearing underneath app content.

src/components/ui/Sidebar/index.tsx
```diff
@@ -13,20 +13,23 @@ import { Fragment, useCallback } from 'react';
...
-    <Tooltip
-      closeDelay={250}
-      gutter={10}
-      hasArrow
-      label={label}
-      placement="right"
-    >
-      {children}
-    </Tooltip>
+    <Center>
```
herbig, line 22: Centers the sidebar components within the tooltip wrapper itself, since centering the content of the sidebar caused horizontal dividers to be clipped

src/components/ui/menus/OptionMenu.tsx
```diff
@@ -21,7 +22,7 @@ export interface Option {
 
 interface IOptionMenu {
   trigger: ReactNode;
-  titleKey: string;
+  titleKey?: string;
```
herbig, line 25: made title optional, since the language picker doesn't have one

src/i18n/LanguageSwitcher.tsx
```diff
@@ -1,20 +1,50 @@
...
 import { useTranslation } from 'react-i18next';
-import Translate from '../components/ui/svg/Translate';
+import { OptionMenu } from '../components/ui/menus/OptionMenu';
+import { supportedLanguages } from '.';
 
-function LanguageSwitcher() {
+export function LanguageSwitcher() {
   const { i18n } = useTranslation();
+
+  let supported = Object.keys(supportedLanguages).map(function (languageCode) {
```
herbig, line 9: The is the actual, ultimate list of supported languages, which will be all the keys in the `supportedLanguages` object.

src/i18n/index.ts
```diff
@@ -32,7 +33,7 @@ import TREASURY_EN from './locales/en/treasury.json';
  * https://www.i18next.com/translation-function/interpolation
  */
 
-const resources = {
+export const supportedLanguages = {
```
herbig, line 36: Changed to `supportedLanguages` and exported, so I could get the list of keys in the picker

src/components/ui/menus/OptionMenu.tsx
```diff
@@ -21,7 +22,7 @@ export interface Option {
 
 interface IOptionMenu {
```
mobiusTripper-crypto, line 23: what if you made this extend MenuProps (from chakra) like
```
interface IOptionMenu extends Omit<MenuProps, 'children'> {
  trigger,
  titleKey,
    ...
  children,
  closeOnSelect = true,
  ...rest
}
```
and then
```
    <Menu
      isLazy
      {...rest}
    >
```

Then you could pass in properties that related to Menu, in particular I would probably replace `gutter` with something like       `offset={[16, 8]}` ... which could then be set in LanguageSwitch. And OptionsMenu wouldn't be tied to that offset if we need to reuse it sometime?

with the offset property set?
image

mobiusTripper-crypto (approved):

src/components/ui/menus/OptionMenu.tsx
herbig: Awesome, yeah I didn't know I could do that @mobiusTripper-crypto, thanks!

PR #752, opened 2022-12-07
add Proposal Details for create proposal page

mobiusTripper-crypto:
## Description
Added Proposal Detail info to the Create Proposal page per the design documents. 
Also made some fixed to the GoveranceInfo components as they load very similar data.

## Notes
NOTE: There are still some loading issue in which loading a favorite just show the loading bars, I believe that is fixed in an open PR. But refresh and the data will load correctly.

## Testing
- Open a DAO
- From the Proposals Page choose Create Proposal.
- Verify details displayed in the panel on the right are correct for the currently selected DAO.

image
image


src/providers/Fractal/governance/hooks/useGovernanceTokenData.ts
```diff
@@ -32,9 +33,15 @@ export interface IGoveranceTokenData extends ITokenData, ITokenAccount, VotingTo
   isLoading?: boolean;
 }
 
+interface ValueFormattedPair {
+  value: BigNumber;
+  formatted?: string;
+}
+
```
mobiusTripper-crypto, line 40: I'm not entirely please with my naming of this object, so any suggestions are welcomed. But it will make both the BigNumber value, and a formatted version of that value available to consumers of this hook.

src/components/ProposalCreate/Transactions.tsx
```diff
@@ -76,6 +76,9 @@ function Transactions({
                     onClick={() => removeTransaction(index)}
                     minWidth="auto"
                     _hover={{ color: 'gold.500' }}
+                    _disabled={{ opacity: 0.4, cursor: 'default' }}
+                    sx={{ '&:disabled:hover': { color: 'inherit', opacity: 0.4 } }}
+                    disabled={pending}
                   />
```
mobiusTripper-crypto, line 82: Not really part of this PR. but the Delete Transaction button was not being disabled.
@epicbadtiming - maybe we could get IconButton into the decent-ui (or if you suggest a different approach for this implementation,)

src/providers/Fractal/governance/hooks/useGovernanceTokenData.ts
```diff
@@ -140,20 +150,29 @@ const useTokenData = (votingContract?: OZLinearVoting, tokenContract?: VotesToke
     (async () => {
       // @todo handle errors
       const votingPeriod = await votingContract.votingPeriod();
-      const blockNumber = await provider.getBlockNumber();
-      // @note Two is subtracted from current block number to ensure that blocks are in sync with other providers
-      const quorum = await votingContract.quorum(blockNumber - 2);
+      const quorum = await votingContract.quorumNumerator();
```
tbwebb22, line 153: So this is quorum percentage, correct? Maybe we should rename this variable to `quorumPercentage` or something. I'm thinking of of quorum as the raw number of token votes, whereas `quorumPercentage` is expressed as a percentage of total token supply

tbwebb22 (approved):
One minor non-blocking comment, LGTM

src/providers/Fractal/governance/hooks/useGovernanceTokenData.ts
mobiusTripper-crypto: Yes, agreed. we should be more explicit with the naming of quorum since it can mean a few different things.

PR #769, opened 2022-12-13
Nav tooltips appear under the NavBar

mobiusTripper-crypto:
## Description
Display order of the GridItems in App.txs doesn't allow for popover items to display in the correct z-index order. Change the order so both the tooltip and language menu display correctly.

image


herbig (approved):
🛳🛳🛳

Da-Colon:
Hmmm I had tested and thought just adding the Popover was fine?

mobiusTripper-crypto:
@Da-Colon but the order of the GridItems on app.tsx is what keeps everything in the correct order, somehow on your merge they got changed back to where they didn't work. I don't think using z-index in this situation is reliable.

Da-Colon:
I guess I'm confused on how order of gridItems is related to Zindex of the tooltips? if this works not a big deal lets ship, I can ask questions later.

Da-Colon (approved):

mobiusTripper-crypto:
In this case I believe the 'stacking context', in conjuntion with setting the transitions and position properties overrides z-index. So with the conflict in the Language Menu and Proposal list. Some of the items in the proposal list have transitions (can't recall the element off the top of my head not), and some have relative position set. Both of those properties will result in them being displayed in the order in which they are added to the DOM. By switching the order in app.tsx, then the menu is in the DOM first. I'm pretty sure zindex only really works in side a single stacking context, and in this case `nav`, `header`, and `toolbar` are all in different contexts ... the same problem occurred with the favorites menu actually.

mobiusTripper-crypto:
@Da-Colon I didn't address your actual question. which is why the tooltip is under the sidebar ... which I think we could discuss later, becuase my brain hurts from thinking about zindex

PR #771, opened 2022-12-13
adds cursor pointer and increases size of copy icon

Da-Colon:
## Description
Quick Fix.


herbig (approved):

mobiusTripper-crypto (approved):
I wonder if that Flex box would make more sense as a button? no point in redoing that now, but a button would have gotten us a pointer included.

PR #774, opened 2022-12-14
Nodes page bug fixes

herbig:
## Description

1. Moves title to json file
2. defaults to expanded view
3. fixes child nodes always having the "favorited" state of the parent

### DIFF
--- a/src/components/ui/cards/DAOInfoCard.tsx
+++ b/src/components/ui/cards/DAOInfoCard.tsx
@@ -10,6 +10,7 @@ import { BACKGROUND_SEMI_TRANSPARENT } from '../../../constants/common';
 import useDAOName from '../../../hooks/DAO/useDAOName';
 import { useCopyText } from '../../../hooks/utils/useCopyText';
 import useDisplayName from '../../../hooks/utils/useDisplayName';
+import { useIsFavorite } from '../../../providers/Fractal/hooks/account/useAccountFavorites';
 import { useFractal } from '../../../providers/Fractal/hooks/useFractal';
 import { ManageDAOMenu } from '../menus/ManageDAO/ManageDAOMenu';
 
@@ -29,12 +30,13 @@ export function DAOInfoCard({
 }: IDAOInfoCard) {
   const {
     account: {
-      favorites: { isConnectedFavorited, toggleFavorite },
+      favorites: { toggleFavorite },
     },
   } = useFractal();
   const copyToClipboard = useCopyText();
   const { accountSubstring } = useDisplayName(safeAddress);
   const { daoRegistryName } = useDAOName({ address: safeAddress });
+  const isFavorite = useIsFavorite(safeAddress);
   return (
     <Flex
       justifyContent="space-between"
@@ -84,11 +86,7 @@ export function DAOInfoCard({
               minWidth="0px"
               aria-label="Favorite Toggle"
               icon={
-                isConnectedFavorited ? (
-                  <StarGoldSolid boxSize="1.5rem" />
-                ) : (
-                  <StarOutline boxSize="1.5rem" />
-                )
+                isFavorite ? <StarGoldSolid boxSize="1.5rem" /> : <StarOutline boxSize="1.5rem" />
               }
               onClick={() => toggleFavorite(safeAddress)}
             />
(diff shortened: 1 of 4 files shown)


src/components/ui/cards/DAOInfoCard.tsx
```diff
@@ -29,12 +30,13 @@ export function DAOInfoCard({
 }: IDAOInfoCard) {
   const {
     account: {
-      favorites: { isConnectedFavorited, toggleFavorite },
```
herbig, line 32: using isConnectedFavorited on child nodes made them all have the favorited state of the parent.  We needed a way to tell if a Node was favorited, regardless of whether it's the connected one (below).

src/pages/FractalNodes/index.tsx
```diff
@@ -11,8 +12,9 @@ export function FractalNodes() {
   const {
     gnosis: { safe, parentDAOAddress, childNodes },
   } = useFractal();
+  const { t } = useTranslation();
   const [isParentExpanded, setIsParentExpended] = useState(true);
-  const [isChildrenExpanded, setIsChildrenExpanded] = useState(false);
+  const [isChildrenExpanded, setIsChildrenExpanded] = useState(true);
```
herbig, line 17: I think defaulting to expanded makes sense, otherwise it's easy to miss you can expand.  This expand/collapse will make more sense when we have deeper node trees.

src/components/ui/cards/DAOInfoCard.tsx
Da-Colon: isn't it checking the address of the child itself?

Da-Colon: o I see now....This used to be a serperate component. DAOInfoCard is connected directly to the safe address

Da-Colon: heard. WE can create an issue and get this straighten out quick enough

src/pages/FractalNodes/index.tsx
Da-Colon: +1

Da-Colon (approved):
LGTM

src/components/ui/cards/DAOInfoCard.tsx
herbig: > heard. We can create an issue and get this straighten out quick enough

yeah the fix is in this PR!

Da-Colon: ha! re read, re looked. We needed, past tense :D

src/components/ui/cards/DAOInfoCard.tsx
```diff
@@ -84,11 +86,7 @@ export function DAOInfoCard({
               minWidth="0px"
               aria-label="Favorite Toggle"
               icon={
-                isConnectedFavorited ? (
-                  <StarGoldSolid boxSize="1.5rem" />
-                ) : (
-                  <StarOutline boxSize="1.5rem" />
-                )
+                isFavorite ? <StarGoldSolid boxSize="1.5rem" /> : <StarOutline boxSize="1.5rem" />
```
mobiusTripper-crypto, line 89: I'm not sure a hook makes sense here, since you could get favoriteList:
```
const {
    account: {
      favorites: { favoritesList, toggleFavorite },
    },
  } = useFractal();
```

and replace isFavorite with `favoritesList.includes(safeAddress)`

unless you think `useIsFavorite` is going to be extended someday. I think the use of `is` in the hook name seems funny.

mobiusTripper-crypto (approved):
A few comments that might need addressed or not.

src/providers/Fractal/hooks/account/useAccountFavorites.ts
```diff
@@ -85,3 +86,15 @@ export const useAccountFavorites = ({ safeAddress, accountDispatch }: IUseFavort
     });
   }, [accountDispatch, isConnectedFavorited, toggleFavorite, favoritesList]);
 };
+
+/**
+ * @returns favorited status of the provided safe address
+ */
+export const useIsFavorite = (safeAddress: string) => {
```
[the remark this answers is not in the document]
herbig: yeah, I think `isConnectedFavorites` is used elsewhere, since aside from this Nodes page, that's generally what you're interested in, so I've kept it.

src/components/ui/cards/DAOInfoCard.tsx
herbig: yeah, agreed, I just removed the hook and used the list in that way instead

PR #775, opened 2022-12-14
don't show zero eth

herbig:
## Description

The Gnosis API returns regular unwrapped eth in the list of assets even if the balance is zero.

This change ignores this if it's zero, so it won't display in the list.

## Screenshots (before / after)

Screenshot 2022-12-13 at 7 45 33 PM

Screenshot 2022-12-13 at 7 46 30 PM


Da-Colon (approved):
🔢

src/pages/Treasury/hooks/useFormatCoins.tsx
```diff
@@ -16,9 +16,10 @@ export interface TokenDisplayData {
 
 export function useFormatCoins(assets: SafeBalanceUsdResponse[]) {
   let totalFiatValue = 0;
-  let displayData: TokenDisplayData[] = new Array(assets.length);
+  let displayData: TokenDisplayData[] = [];
   for (let i = 0; i < assets.length; i++) {
     let asset = assets[i];
+    if (asset.balance === '0') continue;
```
mobiusTripper-crypto, line 22: This is totally fine, but technically balance is a BigNumber, so checked for a string '0' seems strange, but it works. And this is a ways off, but at some point down the road we might want to filter out for dust, yeah, that's a different issue.

mobiusTripper-crypto: The use of continue here really throws me off also. it's really strange to see a "real" for loop and not a forEach and a filter ....

mobiusTripper-crypto (approved):

src/pages/Treasury/hooks/useFormatCoins.tsx
herbig: We're currently typing this as a string, but we should probably be using BigNumber actually...

PR #782, opened 2022-12-14
Update App.tsx

mobiusTripper-crypto:
## Description
The order in which the main grid items are added to the DOM determines how they are shown in the browser.

## Notes

Because I was careless with the last PR dealing with how elements are stacked in the DOM, I needed to make this fix also.

## Testing

Try to find any popups that will go over other element. Favorites and Language Switcher are the two obvious ones.

image

image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -42,17 +42,6 @@ function App() {
       gridTemplateRows={`${HEADER_HEIGHT} minmax(${CONTENT_HEIGHT}, 100%)`}
       position="relative"
     >
-      <GridItem area={'header'}>
-        <Box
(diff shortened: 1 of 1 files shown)


Da-Colon (approved):
I've got to look into this more. is this strictly a `grid` thing or is this have to do with Chakra-ui grid-item component itself? I am very much less familiar with CSS Grid as I'd like to be

mobiusTripper-crypto:
@Da-Colon I think it's a DOM / browser thing. Maybe we can have a chit chat. I am by no means an expert on this at all.

herbig (approved):

PR #784, opened 2022-12-14
copy tweaks for veto/freeze screen

herbig:
## Description

1. Hides the Veto Votes Threshold input, for now
2. Flips the order of Execution and Timelock Period to put Timelock first, since that's the order they occur in.
3. Adds a title to the Freeze parameters
4. fixes incorrect seconds to 2 days conversion
5. Copy tweaks because I'm a copy nerd.

## Screenshots (before / after)

Screenshot 2022-12-13 at 9 00 30 PM
Screenshot 2022-12-13 at 8 59 03 PM


mobiusTripper-crypto (approved):
Looks good. I pushed up a change to remove the `{ ns: 'daoCreate' }` since the `common` namespace wasn't even being used anywhere

PR #785, opened 2022-12-14
Update | Dashboard Activities Loading

Da-Colon:
## Description
The loading was targeting the wrong thing. updated wait for contracts to actually be loaded. I imagine this in short term will be replaced with Skeletons, once we get those implemented.

## Testing
Visit each type of DAO, and ensure that you don't see 'no Activity' flash


Da-Colon:
Not quick like I expected. closing for now

PR #795, opened 2022-12-14
Update | Adds delay to DAO Creation

### LINKED ISSUE #632: Handle Gnosis API indexing delay
### Overview

There may be a delay between when a new DAO is created and when Gnosis indexes data for that DAO (and can be served by their API)

The code that fetches data from the Gnosis API should be updated to detect if the DAO address is a Fractal DAO. If the Gnosis API is failing to serve API data for a valid DAO (404 / 500):

1. If a 404 is returned, determine if it's safe to assume this is b/c Gnosis hasn't indexed the data yet, and if so, display a message to the user that their DAO will be ready shortly. 

2. If a 500 is returned, then determine best way to handle this in the UX. 

3. We could also put a small delay (~2 seconds) after DAO creation before redirecting to give extra buffer time.

It may help to also put a refetch (or backoff function) on the API at an interval in either of these cases.

Da-Colon:
## Description
Adds a way to wait for the DAO to be indexed by Safe Api before redirecting to DAO Dashboard

## Notes
- I only added this to Parent DAO creation

Closes #632 

## Testing
- Create Usul and Multisig DAO (parent). should have no problems loading in.


src/hooks/utils/useAsyncRetry.ts
```diff
@@ -0,0 +1,38 @@
...
+        setTimeout(() => resolve(func().catch(reject)), 3000);
+      }).catch(() => null);
+
+      if (funcResponse) {
+        return funcResponse;
+      }
+
+      if (!currentRetries) {
+        // @todo maybe a toast Error?
+        return null;
```
herbig, line 27: yeah probably the caller can determine what to do with a null response

herbig (approved):
🚀

mobiusTripper-crypto (approved):
looks good

src/hooks/utils/useAsyncRetry.ts
sethhrbek: Yeah, displaying some error on caller side is a good idea. 

You don't need to change anything here ---- just notes ---- this could throw whatever was caught in the last failure of the function run. May not matter for whatever the user sees in the UX, but would help with debugging (unless we just assume its b/c gnosis API is boinked if we get here, in which case ignore all these comments).

A similar way I've seen this done is below, where if the last loop fails, it will throw the caught error, which can be caught in the caller.

```
const MAX_RETRIES = 5;
const INTERVAL_BETWEEN_RETRIES = 5000;
export const wait = ms => new Promise(res => setTimeout(res, ms));

export const retryFn = useCallback(async (func: () => Promise<any>, depth:number = 0) => {
  try {
    return await fn();
  } catch (e) {
    if (depth > MAX_RETRIES) {
      throw e;
    }

    await wait(INTERVAL_BETWEEN_RETRIES);

    return retryFn(fn, depth + 1);
  }
}, []);

const myResult = await retryFn(whateverFunctionToRetry);
```

herbig: @decent-dao/fractal merged this in before @sethhrbek got this comment in, but useful suggestions here

PR #796, opened 2022-12-14
Send Assets Modal

mobiusTripper-crypto:
## Description
Allow user to create a new proposal to send DAO assets to another address.

## Notes
A new issue has been create to add functionality to send ETH (native token). So currently you can only send ERC20.

The input for the token amount doesn't handle BigNumbers correctly (like any of the token inputs in the app) ... so there can be some error if attempting to send very large or very small amounts.

Also the code that calculates USD price for the token can be refactored right away.

## Testing
Make certain that you send an ERC20 token to the DAO
And make sure your use is either a DAO owner or has tokens deletaged for voting.
From the Treasury page, given the above condition, the send asset button should be in the upper right.
Use the Sent Asset Modal to create a proposal, then vote, queue and execute the proposal to verify the token is sent to the correct address.

image


src/hooks/utils/useAddress.ts
```diff
@@ -47,24 +47,22 @@ const useAddress = (addressInput: string | undefined) => {
...
-
-        setValidAddress(false);
-        setLoading(false);
+        setIsValidAddress(false);
+        setIsAddressLoading(false);
       });
   }, [provider, addressInput]);
 
-  return [address, validAddress, loading] as const;
+  return { address, isValidAddress, isAddressLoading };
```
mobiusTripper-crypto, line 65: I refactored this hook to return an object, as returning that array didn't seem quite right. I also renamed the booleans to have `is`. That the coding standard I've become accustom too.

Da-Colon: so...using the array pattern technically is a correct pattern and a preference of a few devs. I prefer objects myself. but in this case we have a lot of things called `address` through the codebase by using an array, you are able to use whatever name you need to within the component itself that uses this hook. 

I'm not saying this needs to be changed back, I have no preference, but food for thought

Da-Colon (commented):
image

These Conversions shouldn't be user facing. After I enter 0.2 on blur it adds decimal places

image

getting an address error before entering anything in. Probably can add something to ensure the value is entered before checking validity

src/components/ui/modals/SendAssetsModal.tsx
```diff
@@ -131,7 +155,7 @@ export function SendAssetsModal({ close }: { close: () => void }) {
           </Flex>
           <NumberInput
             placeholder="0"
-            precision={0}
+            precision={isEth(selectedAsset) ? 18 : selectedAsset.token.decimals}
```
Da-Colon, line 158: realized. My comment on the input wasn't the change to the bigNumber but rather this precision. I would use `limitDecimalsOnKeyDown` instead here.

src/components/ui/modals/SendAssetsModal.tsx
```diff
@@ -131,7 +155,7 @@ export function SendAssetsModal({ close }: { close: () => void }) {
           </Flex>
           <NumberInput
```
Da-Colon, line 156: Might be a good idea to just go ahead and switch this out to a regular `Input` instead of the `NumberInput`

Da-Colon (commented):
thoughts that randomly hit me.

src/hooks/utils/useAddress.ts
mobiusTripper-crypto: Noted, thanks. I was under the impression you could name address if whatever you wanted with 
` const { address: Myaddress, isValidAddress, isAddressLoading } = useAddress(searchString);`

src/components/ui/modals/SendAssetsModal.tsx
mobiusTripper-crypto: I can make this change for now, but I kind of phoned it in on the Input because it doesn't really work for BigNumbers anyway and we have an open issue to fix those.

src/components/ui/modals/SendAssetsModal.tsx
mobiusTripper-crypto: Switched

src/components/ui/modals/SendAssetsModal.tsx
```diff
@@ -33,11 +34,41 @@ export function SendAssetsModal({ close }: { close: () => void }) {
...
+        isEth(selectedAsset) ? 18 : selectedAsset.token.decimals
+      );
+    } catch {
+      return BigNumber.from('0');
+    }
+  };
+
   const sendAssets = useSendAssets({
-    normalizedBalance: amountInput,
+    transferAmount: transferAmount(),
```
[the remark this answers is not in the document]
mobiusTripper-crypto: yes, thanks, i got lazy at the end, fixing this up correctly

src/pages/Treasury/hooks/useSendAssets.ts
```diff
@@ -1,29 +1,47 @@
...
+  transferAmount: BigNumber;
   asset: SafeBalanceUsdResponse;
   destinationAddress: string;
 }) => {
-  const rawBalance =
-    Number(normalizedBalance) * (10 ^ (asset?.token?.decimals ? asset.token.decimals : 18));
+  const { submitProposal } = useSubmitProposal();
+
+  const successCallback = () => {
+    //TODO: need to add a toast and close modal
```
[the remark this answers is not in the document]
mobiusTripper-crypto: I have a follow up PR for this. I just wanted to get the basic transfer function working ...

src/pages/Treasury/hooks/useSendAssets.ts
```diff
@@ -1,29 +1,47 @@
...
+        transferAmount,
+      ]),
+    ];
+
+    const proposalData: ProposalExecuteData = {
+      targets: [tokenAddress],
+      values: [BigNumber.from('0')],
+      calldatas: calldatas,
+      title: 'Send Assets',
+      description: 'to:' + destinationAddress + ' amt:' + transferAmount.toString(),
```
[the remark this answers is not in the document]
mobiusTripper-crypto: fixed

src/pages/Treasury/index.tsx
```diff
@@ -4,19 +4,34 @@ import PageHeader from '../../components/ui/Header/PageHeader';
...
+  } = useWeb3Provider();
+
   const { t } = useTranslation('treasury');
-  // TODO show only if the user is a DAO signer or delegated token holder and assets > 0
-  // TODO needs transaction implementation
-  const showButton = false && assetsFungible.length > 0;
+
+  const isOwnerOrDelegate =
+    type === GovernanceTypes.GNOSIS_SAFE
+      ? owners?.includes(account || '')
```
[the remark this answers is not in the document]
mobiusTripper-crypto: `includes` only takes a string. i can figure a different way to handle the null, but account is typed at `string | null`

src/components/ui/modals/SendAssetsModal.tsx
mobiusTripper-crypto: changed

src/hooks/utils/useAddress.ts
herbig: yeah I don't have a return preference, mostly been copying what I see you guys do, but I'll try to default to objects

herbig:
@mobiusTripper-crypto code looks good, one thing, could you fix the modal opening with `$NaN` as the fiat total, and also when you've typed only a `.` while inputting a decimal?

Screenshot 2022-12-15 at 8 22 11 PM

herbig (approved):
🛳

PR #801, opened 2022-12-15
BugFix | Handle duplicate Address allocation error

### LINKED ISSUE #786: Cannot Deploy DAO with duplicate Trusted Addresses
Add validation to check for duplicate Trusted Addresses.

Image

Da-Colon:
## Description
Updates Token Allocation to handle duplicate address error

## Notes
Couple other additional updates
- Updates handles blank input values. (button is disabled)
- Handles 0 allocationed (button is disabled)
- Cleans up update methods as callbacks and include 'snapshots' of state to ensure correct data at time of call.

Closes #786 

## Testing
Go to Token Allocations during Usul creation.
- Try to add an additional allocation with same address
- Try to add an additional allocation different address no amount
- Try to add an additional allocation different address 0

- Finish off by making sure that Valid values will pass still.

localhost_3000_(Desktop) (26)


src/i18n/locales/en/daoCreate.json
```diff
@@ -52,6 +52,7 @@
     "exampleFreezePeriod": "604800 seconds = 1 week",
     "vetoGuardDescription": "These configuration values may be changed later on by passing a proposal to do so on this child DAO.",
     "errorConnectWallet": "Connect wallet to validate.",
+    "errorDuplicateAddress": "Duplicate Address",
```
herbig, line 55: A small nitpick but all the error messages end in punctuation now!

herbig (approved):
Awesome, looks good

mobiusTripper-crypto:
I was able to get it into this state.

image

So enter a dup, then change the 1st one. Dup error still there.

Honestly, it seems like we need to address validation at an App level. And stop trying to debug all this individual errors?

mobiusTripper-crypto:
And does this seem correct. Button is enabled but the amount is still blank for one value?

image

Da-Colon:
> And does this seem correct. Button is enabled but the amount is still blank for one value?

nice catch. left a `.every()` instead a `.some()` on the validation check

Da-Colon:
@mobiusTripper-crypto this is ready for re-review. do your worst

mobiusTripper-crypto (approved):
Slight dup problem, but looks good.

PR #802, opened 2022-12-15
remove duplicate useDAOName hook and fix ENS display logic

herbig:
## Description

Fixes inconsistently showing ENS names in the app by removing duplicate implementations of `useDAOName`

## Testing

View `ensname.eth` DAO and ensure the name appears as `ensname.eth` throughout the app.


src/components/ui/cards/DAOInfoCard.tsx
```diff
@@ -28,13 +27,13 @@ export function DAOInfoCard({
   numberOfChildrenDAO,
 }: IDAOInfoCard) {
   const {
+    gnosis: { daoName },
```
herbig, line 30: using the daoName set in useFractal() throughout, rather than calling useDAOName

src/providers/Fractal/hooks/useDAOName.ts
```diff
@@ -12,27 +13,30 @@ export default function useDAOName({
   gnosisDispatch: Dispatch<GnosisActions>;
 }) {
   const { fractalNameRegistryContract } = useSafeContracts();
-  const { displayName } = useDisplayName(address);
```
herbig, line 15: using `displayName` doesn't work here, because it will return either the ENS or a truncated address, but we need the preference of:

1. ENS
2. name register contract
3. truncated address

src/providers/Fractal/hooks/useDAOName.ts
```diff
@@ -12,27 +13,30 @@ export default function useDAOName({
...
+        return createAccountSubstring(address!);
       }
       const events = await fractalNameRegistryContract.queryFilter(
         fractalNameRegistryContract.filters.FractalNameUpdated(_address)
       );
 
       const latestEvent = events[0];
       if (!latestEvent) {
-        return displayName;
+        return createAccountSubstring(address!);
```
herbig, line 32: we can do ! here because there is a check before calling the `getDaoName` function that address is truthy

herbig:
I just realized this broke Favorites.. because that needs the other useDAOName hook...

Setting to draft and will figure that out

src/components/ui/cards/DAOInfoCard.tsx
herbig, line 30: using the dispatched `daoName` here, rather than the `useDAOName` hook

src/providers/Fractal/hooks/useDispatchDAOName.ts
```diff
@@ -0,0 +1,27 @@
+import { Dispatch, useEffect } from 'react';
+import useDAOName from '../../../hooks/DAO/useDAOName';
+import { GnosisAction } from '../constants';
+import { GnosisActions } from '../types';
+
+export default function useDispatchDAOName({
```
herbig, line 6: this is a renaming of the duplicate `useDAOName` above, refactored to just call the other `useDAOName`

src/hooks/utils/useDisplayName.ts
```diff
@@ -6,6 +6,14 @@ export const createAccountSubstring = (account: string) => {
...
 
+/**
+ * Gets the 'display name' of an Ethereum address, by first checking if there is a corresponding
+ * Primary ENS Name (reverse record), otherwise returning a truncated address in the form
+ * 0xbFC4...7551
+ *
+ * This is intended to be used for NON Fractal DAO display names.  If you would like to get the
+ * display name for a Fractal DAO, use the useDAOName hook instead.
+ */
 const useDisplayName = (account?: string | null) => {
```
herbig, line 17: I'm not sure we even need this method either.

I think we should probably use the useDAOName logic for everything, in case the address you're checking is in fact a Fractal DAO.

The way the app is currently set up, none of the addresses in Proposals List or Details will display DAO names from the Fractal name registry, if the address is a Fractal DAO.

herbig: That adds a lot of overhead though, as we'd be checking all personal wallets to see if there's an event in the Fractal name registry giving it a name.

Da-Colon (approved):
LGTM

PR #805, opened 2022-12-15
Multi tx

tbwebb22:

### DIFF
--- a/src/hooks/DAO/proposal/useSubmitProposal.ts
+++ b/src/hooks/DAO/proposal/useSubmitProposal.ts
@@ -45,7 +45,7 @@ export default function useSubmitProposal() {
 
         setPendingCreateTx(true);
 
-        let to, data;
+        let to, data, operation;
         if (proposalData.targets.length > 1) {
           // Need to wrap it in Multisend function call
           to = multiSendContract.address;
@@ -62,10 +62,13 @@ export default function useSubmitProposal() {
           data = multiSendContract.interface.encodeFunctionData('multiSend', [
             encodeMultiSend(tempData),
           ]);
+
+          operation = 1;
         } else {
           // Single transaction to post
           to = proposalData.targets[0];
           data = proposalData.calldatas[0];
+          operation = 0;
         }
 
         try {
@@ -79,6 +82,7 @@ export default function useSubmitProposal() {
               {
                 to,
                 data,
+                operation,
                 nonce: (await gnosisContract.nonce()).toNumber(),
               }
             )
--- a/src/hooks/DAO/proposal/useSubmitProposal.ts
+++ b/src/hooks/DAO/proposal/useSubmitProposal.ts
@@ -1,19 +1,22 @@
 import { TypedDataSigner } from '@ethersproject/abstract-signer';
 import { GnosisSafe__factory } from '@fractal-framework/fractal-contracts';
 import axios from 'axios';
-import { Signer } from 'ethers';
+import { BigNumber, Signer } from 'ethers';
 import { useCallback, useState } from 'react';
-import { buildSafeAPIPost } from '../../../helpers';
+import { buildSafeAPIPost, encodeMultiSend } from '../../../helpers';
+
 import { logError } from '../../../helpers/errorLogging';
 import { useFractal } from '../../../providers/Fractal/hooks/useFractal';
 import { buildGnosisApiUrl } from '../../../providers/Fractal/utils';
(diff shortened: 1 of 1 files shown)


herbig:
@tbwebb22 for some reason this shows in the proposals list as 3 transactions, when there's only 2.

Also this breaks the ability to decode the transactions in the details, is there a way we can still do that?

Screenshot 2022-12-15 at 10 51 33 AM

herbig:
Also, just FYI on your local env you can click the "Create Proposal" title to quickly add to transactions in the create screen

Da-Colon:
> @tbwebb22 for some reason this shows in the proposals list as 3 transactions, when there's only 2.
> 
> Also this breaks the ability to decode the transactions in the details, is there a way we can still do that?
> 
> Screenshot 2022-12-15 at 10 51 33 AM

I think this more the parsing fault than his multisend. Lets spin this to a seperate issue

herbig:
> I think this more the parsing fault than his multisend. Lets spin this to a seperate issue

Sounds good, created an issue https://github.com/decent-dao/fractal-interface/issues/806

And we'll ignore that for this PR

src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -34,10 +37,40 @@ export default function useSubmitProposal() {
       }
 
       if (!usulContract || !votingStrategiesAddresses || !safe.address) {
-        if (!safe.address || !signerOrProvider) {
+        // Submit a multisig proposal
+
+        if (!safe.address || !signerOrProvider || !multiSendContract) {
```
herbig, line 42: would it make sense to do `|| (!multiSendContract && proposalData.targets.length > 1)` here?

herbig (approved):
🚀

src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -47,8 +80,9 @@ export default function useSubmitProposal() {
               signerOrProvider as Signer & TypedDataSigner,
               chainId,
               {
-                to: proposalData.targets[0],
-                data: proposalData.calldatas[0],
+                to,
+                data,
+                operation,
                 nonce: (await gnosisContract.nonce()).toNumber(),
```
Da-Colon, line 86: Should nonce is changed, when a proposal is executed yes? if I were to create one proposal (don't execute) then create another would this be made with the same nonce? (testing this now)

Da-Colon (changes requested):
one other thing. if the create proposal is successful. can you call

```ts
const { actions: { getGnosisSafeTransactions } } = useFractal()
```

This will refresh this list when it redirects

src/hooks/DAO/proposal/useSubmitProposal.ts
Da-Colon: image

Da-Colon: https://app.safe.global/gor:0xEC1e72322649e757b753E9e8c12ba385cfDf055e/transactions/queue

src/hooks/DAO/proposal/useSubmitProposal.ts
tbwebb22: Yeah, makes sense. I just pushed a commit to move that !multiSendContract check down within the if block where we're checking for more than one tx.

src/hooks/DAO/proposal/useSubmitProposal.ts
tbwebb22: Yes that is true. We probably need to handle this differently. 
However this is also a problem for single tx proposals right?

tbwebb22:
> one other thing. if the create proposal is successful. can you call
> 
> ```ts
> const { actions: { getGnosisSafeTransactions } } = useFractal()
> ```
> 
> This will refresh this list when it redirects

Nice, pushed this up

mobiusTripper-crypto (approved):
look good

herbig:
@Da-Colon can you recheck this before we merge?

src/hooks/DAO/proposal/useSubmitProposal.ts
Da-Colon: yessir. We discussed a little more this morning. Probably need to create a ticket and update Proposal Creation to look up the lastest proposal and add one Nonce.

Da-Colon (approved):

PR #809, opened 2022-12-15
BugFix | Clean DAO data when not viewing DAO

### LINKED ISSUE #793: Side NavItems show on Welcome screen.
To Reproduce:
- Load in a DAO
- Click on Fractal F logo (top-right)

Image

Da-Colon:
## Description
Added a useEffect to clear state when not within DAO.

## Notes
- I tried a few different approaches here. I would prefer to have `DAOController` make this update, but because of how React Route is works. params of nested routes won't be recongnized by parent. So moving `useDAOController` isn't really an option and doesn't work. 
- There isn't any real way (and you shouldn't) update state on unMount.
- I believe I had a version of this before and had deleted it as it was causing some issues with loading. This version using routeMatch to check if state should be reset. instead of data from state.

Closes #793

## Testing
Go to all the different routes and back home again. make sure state is changing correctly

### DIFF
--- a/src/hooks/DAO/useDAOController.ts
+++ b/src/hooks/DAO/useDAOController.ts
@@ -66,14 +66,4 @@ export default function useDAOController() {
       }
     }
   }, [errorMessage, account, loading, isProviderLoading, gnosisDispatch, address, navigate]);
-
-  useEffect(() => {
-    return () => {
-      if (!params.address) {
-        gnosisDispatch({ type: GnosisAction.RESET });
-        treasuryDispatch({ type: TreasuryAction.RESET });
-        governanceDispatch({ type: GovernanceAction.RESET });
-      }
-    };
-  }, [params.address, gnosisDispatch, treasuryDispatch, governanceDispatch]);
 }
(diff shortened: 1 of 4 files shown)


herbig:
Not sure what's up, but the deploy preview for this just hangs on a DAO homepage

https://deploy-preview-809--fractal-framework-interface-dev.netlify.app/#/daos/0xbFC43fba83971596709cA0f66B1C54B89aad7551

Da-Colon:
> Not sure what's up, but the deploy preview for this just hangs on a DAO homepage
> 
> https://deploy-preview-809--fractal-framework-interface-dev.netlify.app/#/daos/0xbFC43fba83971596709cA0f66B1C54B89aad7551

almost went down a rabbit hole. is your account connected? if not that would likely cause that.

herbig:
> > Not sure what's up, but the deploy preview for this just hangs on a DAO homepage
> > https://deploy-preview-809--fractal-framework-interface-dev.netlify.app/#/daos/0xbFC43fba83971596709cA0f66B1C54B89aad7551
> 
> almost went down a rabbit hole. is your account connected? if not that would likely cause that.

🤦

herbig (approved):
Looks good, and works as expected!

PR #812, opened 2022-12-16
Feature/freeze

### LINKED ISSUE #615: Freeze function on subDAO homepage
### Overview

Implement the ability of a parent DAO to freeze a childDAO.

### Link to documentation

https://www.notion.so/hulking-cosmos-3e8/Veto-Freeze-UI-components-94827eb64843494aba4055943753f6b3

https://www.figma.com/file/uRV3Nfowdzt5CknW1FgUkQ/%F0%9F%93%9A-Fractal-Design-System-%7C-Base-Components-%7C-October-2022?node-id=2685%3A33148&t=P1b27Lz50AnVyiV2-0

christopherdancy:
## Description

This feature completes the following:
1. Freeze Activity UI
2. Cast Freeze Tx
3. Event Listener

image

image


src/components/Activity/ActivityCard.tsx
```diff
@@ -3,16 +3,23 @@ import { Calendar } from '@decent-org/fractal-ui';
 import { ReactNode } from 'react';
 import { ActivityBox } from '../ui/containers/ActivityBox';
 
-interface IActivityCard {
-  eventDate: string;
+interface IAcitivityCard {
```
sethhrbek, line 6: mispelling here

src/components/ui/badges/Badge.tsx
```diff
@@ -5,6 +5,11 @@ import { TxProposalState } from '../../../providers/Fractal/types';
 
 type BadgeType = { [key: string]: { Icon?: any; bg: string; color: string } };
 
+export enum BadgeLabels {
+  STATE_FREEZE_INIT = 'stateFreezeInit',
+  STATE_FROZEN = 'stateFrozen',
+}
```
sethhrbek, line 11: Maybe these could be defined as `FreezeState` to match the pattern of `TxProposalState`.

src/components/ui/badges/Badge.tsx
```diff
@@ -28,7 +35,7 @@ const BADGE_SIZES: BadgeSize = {
 
 interface IBadge {
   size: 'sm' | 'base';
-  labelKey: TxProposalState;
+  labelKey: TxProposalState | string;
```
sethhrbek, line 38: Should `string` here be changed to `BadgeLabels` type (or whatever that is renamed to if you rename it)

src/hooks/safe/useSafeContracts.ts
```diff
@@ -66,11 +66,11 @@ export default function useSafeContracts() {
       fractalUsulMasterCopy,
       fractalModuleMasterCopy,
       fractalNameRegistry,
+      votesTokenMasterCopy,
```
sethhrbek, line 69: (minor / whatever) does this break the alphabetical order?

src/hooks/safe/useSafeContracts.ts
```diff
@@ -128,6 +128,18 @@ export default function useSafeContracts() {
       VetoERC20Voting__factory.connect(vetoERC20VotingMasterCopy, signerOrProvider)
     );
 
+    setGnosisVetoGuardMasterCopyContract(
+      VetoGuard__factory.connect(gnosisVetoGuardMasterCopy, signerOrProvider)
+    );
+    setUsulVetoGuardMasterCopyContract(
+      UsulVetoGuard__factory.connect(usulVetoGuardMasterCopy, signerOrProvider)
+    );
+    setvetoMultisigVotingMasterCopyContract(
```
sethhrbek, line 137: minor -- formatting should be `setVeto..`

src/pages/DaoDashboard/Activities/ActivityFreeze.tsx
```diff
@@ -0,0 +1,107 @@
...
+          .div(86400)
+          .toNumber()
+      )
+    : 0;
+
+  return (
+    <ActivityCard
+      Badge={
+        <Badge
+          labelKey={freezeData.isFrozen ? 'stateFrozen' : 'stateFreezeInit'}
```
sethhrbek, line 64: This should utilize the `BadgeLabels` enum

src/pages/DaoDashboard/Activities/ActivityFreeze.tsx
```diff
@@ -0,0 +1,107 @@
...
+          alignItems="center"
+          gap="2rem"
+        >
+          <Text textStyle="text-base-sans-regular">
+            {!freezeData.isFrozen && freezeData.freezeVotesThreshold.gt(0) && (
+              <Tooltip
+                label={
+                  freezeData.freezeProposalVoteCount.toString() +
+                  ' / ' +
+                  freezeData.freezeVotesThreshold.toString() +
```
sethhrbek, line 81: recommend breaking this to a utility method, since this code is repeated below

src/pages/DaoDashboard/Activities/index.tsx
```diff
@@ -28,41 +35,51 @@ export function Activities() {
...
+        gap="1rem"
+      >
+        {(freezeData.isFrozen ||
+          freezeData.freezeProposalCreatedTime
+            .add(freezeData.freezeProposalPeriod)
+            .sub(currentTime)
+            .gt(0)) && (
+          <ActivityFreeze
+            freezeData={freezeData}
+            vetoContract={guardContracts}
```
sethhrbek, line 49: Is there a reason not to pass in the `guardContracts.vetoVotingContract` directly here, as that is the contract being utilized by the component?

src/i18n/locales/en/dashboard.json
```diff
@@ -13,7 +13,13 @@
     "titlePassed": "Passed",
     "titleTreasury": "Treasury",
     "noActivity": "No Activity",
-    "proposalDescription": "Proposal to execute {{txCount}} transactions on",
+    "proposalDescription": "Proposal to execute {{txCount}} transactions",
+    "freezeDescription": "Freeze Initiated on DAO",
+    "frozenDescription": "DAO is Frozen",
+    "freezeButton": "Freeze",
+    "freezeVotedButton": "Freeze Vote Casted",
+    "freezeDaysLeft": " days left",
```
herbig, line 21: for translation purposes, we should have this string be `{{count}} days left`, since depending on the language there might be text before or after the number of days.

src/pages/DaoDashboard/Activities/index.tsx
```diff
@@ -28,41 +35,51 @@ export function Activities() {
...
-            if (activity.eventType === ActivityEventType.Module) {
+      <Flex
+        flexDirection="column"
+        gap="1rem"
+      >
+        {(freezeData.isFrozen ||
+          freezeData.freezeProposalCreatedTime
+            .add(freezeData.freezeProposalPeriod)
+            .sub(currentTime)
+            .gt(0)) && (
```
sethhrbek, line 46: should this be checking if the `vetoGuardContract.vetoVotingContract` exists first? Looking at `useFreeze` hook, it looks like that contract must exist to generate the freeze data, but maybe I'm misunderstanding.

src/i18n/locales/en/dashboard.json
herbig: then translators can do `"freezeDaysLeft": "blah blah {{count}} blah",` in the translated json file

src/pages/DaoDashboard/Activities/ActivityFreeze.tsx
```diff
@@ -0,0 +1,107 @@
...
+                {freezeData.freezeProposalVoteCount.toString() +
+                  ' / ' +
+                  freezeData.freezeVotesThreshold.toString()}
+              </Tooltip>
+            )}
+          </Text>
+          {daysLeft > 0 && (
+            <Text textStyle="text-base-sans-regular">{daysLeft + t('freezeDaysLeft')}</Text>
+          )}
+          {!freezeData.isFrozen && vetoContract.vetoVotingContract && (
```
sethhrbek, line 95: Left a comment below alluding to this -- but wondering if this component should not be rendered unless the `vetoContract.vetoVotingContract` exists

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
```diff
@@ -1,22 +1,31 @@
...
-      { optionKey: 'optionInitiateFreeze', onClick: () => {} }, // TODO freeze hook (if parent voting holder)
-    ],
-    [safeAddress, navigate]
-  );
+  const options = useMemo(() => {
+    const createSubDAOOption = {
+      optionKey: 'optionCreateSubDAO',
+      onClick: () => navigate(DAO_ROUTES.newSubDao.relative(safeAddress)),
+    };
+    if (guardContracts.vetoVotingContract) {
```
herbig, line 19: Can we also check that the user has the ability to start a freeze vote? So delegated token holder / multisig on the parent.

src/components/Activity/FreezeButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+  isFrozen,
+  userHasFreezeVoted,
+  vetoVotingContract,
+}: {
+  isFrozen: boolean;
+  userHasFreezeVoted: boolean;
+  vetoVotingContract: VetoERC20Voting | VetoMultisigVoting;
+}) {
+  const [pending, setPending] = useState<boolean>(false);
+  const { t } = useTranslation(['dashboard', 'common']);
```
herbig, line 17: doesn't look like you're using `common` here at all so it can be removed

src/providers/Fractal/constants/state.ts
```diff
@@ -1,10 +1,26 @@
-import { IGovernance } from '../governance/types';
+import { BigNumber } from 'ethers';
+import { IGovernance, VetoVotingType } from '../governance/types';
+// eslint-disable-next-line import/namespace
```
sethhrbek, line 3: do we need this? ran the linter without this and it didnt throw an error on my end

src/providers/Fractal/governance/types.ts
```diff
@@ -73,6 +77,28 @@ export interface GnosisConfig {
...
+  vetoVotingContract: VetoERC20Voting | VetoMultisigVoting | undefined;
+  vetoVotingType: VetoVotingType;
+}
+
+export interface IGnosisFreezeData {
+  freezeVotesThreshold: BigNumber; // Number of freeze votes required to activate a freeze
+  freezeProposalCreatedTime: BigNumber; // Block number the freeze proposal was created at
+  freezeProposalVoteCount: BigNumber; // Number of accrued freeze votes
+  freezeProposalPeriod: BigNumber; // Number of blocks a freeze proposal has to succeed
+  freezePeriod: BigNumber; // Number of blocks a freeze lasts, from time of freeze proposal creation
```
sethhrbek, line 97: love the comments! very helpful

src/pages/DaoDashboard/Activities/ActivityFreeze.tsx
```diff
@@ -0,0 +1,107 @@
...
+  vetoContract,
+  currentTime,
+}: {
+  freezeData: IGnosisFreezeData;
+  vetoContract: IGnosisVetoContract;
+  currentTime: BigNumber;
+}) {
+  const { t } = useTranslation('dashboard');
+
+  const daysLeft = freezeData.isFrozen
```
herbig, line 37: this should probably use `formatDatesDiffReadable`, so that it will say hours/minutes once you're less than a day

There's an example in `ProposalTime` that uses that for a "minutes/hours/days left" string

herbig: in which case we don't need the `freezeDaysLeft`, since we have `timeLeft`

And now that I'm looking at `timeLeft`, I think we should refactor that so it's not always appending to the front, but I can take care of that after we merge this, ha

src/providers/Fractal/reducers/gnosis.ts
```diff
@@ -27,7 +28,29 @@ export const gnosisReducer = (state: IGnosis, action: GnosisActions): IGnosis =>
...
+    case GnosisAction.SET_GUARD_CONTRACTS:
+      return { ...state, guardContracts: { ...action.payload } };
+    case GnosisAction.SET_FREEZE_DATA:
+      return { ...state, freezeData: { ...action.payload } };
+    case GnosisAction.FREEZE_VOTE_EVENT: {
+      const freezeData = state.freezeData;
+      const { isVoter, freezeProposalCreatedTime, freezeProposalVoteCount } = action.payload;
+      const userHasFreezeVoted = isVoter;
+      const isFrozen = freezeProposalVoteCount.gte(
+        freezeData.freezeVotesThreshold || BigNumber.from(0)
```
herbig, line 40: what does the `|| BigNumber.from(0)` do here?  Is there a case where `freezeVotesThreshold` isn't available for the event?

src/pages/DaoDashboard/Activities/index.tsx
sethhrbek: I guess it wouldn't make a difference to check that contract, as this will use all the default freeze values set in the provider. But makes me wonder if the initial state for `freezeData` should just be `undefined`, b/c that is more accurate to it's actual state when a voting guard contract isn't present. There shouldn't be a case where the veto voting guard isn't present and we need those freeze guard default values, b/c the `useFreeze` hook will always set them when the guard contract is present.

src/providers/Fractal/constants/state.ts
```diff
@@ -1,10 +1,26 @@
...
+  },
+  freezeData: {
+    freezeVotesThreshold: BigNumber.from(0),
+    freezeProposalCreatedTime: BigNumber.from(0),
+    freezeProposalVoteCount: BigNumber.from(0),
+    freezeProposalPeriod: BigNumber.from(0),
+    freezePeriod: BigNumber.from(0),
+    userHasFreezeVoted: false,
+    isFrozen: false,
+  },
```
sethhrbek, line 23: see my comment above -- wondering if the default for this should be `undefined`

src/components/ui/badges/Badge.tsx
Da-Colon: Could we not add freeze to TxProposalState enum?

src/components/ui/badges/Badge.tsx
Da-Colon: (comment above) then we wouldn't have to do this

src/pages/DaoDashboard/Activities/index.tsx
```diff
@@ -28,41 +35,51 @@ export function Activities() {
...
-              <ActivityTreasury
-                key={i}
-                activity={activity as TreasuryActivity}
-              />
-            );
-          })}
-        </Flex>
-      ) : (
-        <EmptyBox emptyText={t('noActivity')} />
-      )}
```
Da-Colon, line 65: This should get reverted, Deleted portion is lastest

src/providers/Fractal/hooks/useFreezeData.ts
```diff
@@ -0,0 +1,113 @@
...
+        },
+      });
+    };
+
+    vetoVotingContract.on(filter, listenerCallback);
+
+    return () => {
+      vetoVotingContract.off(filter, listenerCallback);
+    };
+  }, [account, vetoGuardContract, gnosisDispatch]);
```
sethhrbek, line 112: I think these last 2 `useEffects` can be combined into 1 -- would need to add a conditional to check the veto voting type:

```
if (![VetoVotingType.ERC20, VetoVotingType.MULTISIG].includes(vetoVotingType)) { return }

const vetoVotingContract = 
   vetoGuardContract.vetoVotingType === VetoVotingType.ERC20 ?
      vetoGuardContract.vetoVotingContract as VetoERC20Voting :
      vetoGuardContract.vetoVotingContract as VetoMultisigVoting
```

src/components/ui/svg/Alert.tsx
```diff
@@ -9,7 +9,7 @@ function Alert() {
     >
       <path
         d="M12.8227 3.48644L21.8727 19.5405C21.9561 19.6884 22 19.8562 22 20.027C22 20.1978 21.9561 20.3656 21.8727 20.5135C21.7893 20.6614 21.6694 20.7842 21.525 20.8696C21.3806 20.955 21.2168 21 21.05 21H2.95001C2.78324 21 2.61942 20.955 2.475 20.8696C2.33058 20.7842 2.21065 20.6614 2.12727 20.5135C2.04389 20.3656 2 20.1978 2 20.027C2 19.8562 2.0439 19.6884 2.12728 19.5405L11.1773 3.48644C11.2607 3.33855 11.3806 3.21573 11.525 3.13034C11.6694 3.04495 11.8332 3 12 3C12.1668 3 12.3306 3.04495 12.475 3.13034C12.6194 3.21573 12.7393 3.33855 12.8227 3.48644V3.48644ZM4.59546 19.054H19.4045L12 5.91888L4.59546 19.054V19.054ZM11.05 16.1351H12.95V18.0811H11.05V16.1351ZM11.05 9.3243H12.95V14.1892H11.05V9.3243Z"
-        fill="#F5F5F5"
+        fill={'#F5F5F5'}
       />
     </svg>
   );
```
Da-Colon, line 15: I think at this point All these SVG can be removed @herbig 

@christopherdancy Whats this update?

src/pages/DaoDashboard/Activities/ActivityFreeze.tsx
Da-Colon: TxProposalState~

Da-Colon (changes requested):

src/components/ui/svg/Alert.tsx
herbig: ah, yeah we have `<Alert>` as an icon from the library now right?

src/pages/DaoDashboard/Activities/index.tsx
Da-Colon: yikes, scratch that. missed the it in the changes. ignore comment

src/components/ui/svg/Alert.tsx
Da-Colon: yessir. I think for some reason the linter made this change? but yeah either way a reminder we can delete most/all SVGs

src/providers/Fractal/reducers/gnosis.ts
sethhrbek: I think with the way the `useEffects` are setup in `useFreezeData`, its possible `freezeData` wouldn't be set yet (but should verify).

I was thinking we should set `freezeData` initial state to `undefined`  -- if that was done, then the `useEffect(s)` in `useFreezeData` that calls this `FREEZE_VOTE_EVENT` action, could check if `freezeData` is defined before running so this code would know `freezeData` is defined here.

src/providers/Fractal/hooks/useFreezeData.ts
```diff
@@ -0,0 +1,113 @@
...
+    account,
+    vetoGuardContract,
+    vetoGuardContract.vetoVotingContract,
+  ]);
+
+  useEffect(() => {
+    if (
+      !account ||
+      vetoGuardContract.vetoVotingType !== VetoVotingType.MULTISIG ||
+      !vetoGuardContract.vetoVotingContract
```
sethhrbek, line 56: The previous use effect checks `!vetoGuardContract`, does that need to be done here as well?

herbig (changes requested):
@christopherdancy if a DAO is already Frozen, the option to initiate a Freeze still appears, but clicking it does nothing.  We should not show that option.
Screenshot 2022-12-16 at 5 42 34 PM

src/providers/Fractal/hooks/useFreezeData.ts
sethhrbek: oh right, that's exactly why I thought it also needs to be here -- b/c this could be executed separately so needs to check `!vetoGuardContract` instead of assuming its defined here.

src/providers/Fractal/hooks/useFreezeData.ts
sethhrbek: Is there only 1 vetoGuardContract.vetoVotingType when this hook runs? I was assuming only one listener is going ever going to be setup here when this hook runs, in which case:

1. this code is 99% the same except for typing the contract:`as VetoMultisigVoting` vs. `as VetoERC20Voting` 

2. Combining these makes it clearer when reading there's just 1 listener, and also makes it easier to update logic in a single place if anything changes.

But if this assumption about one listener is wrong, then I agree they could be separate.

sethhrbek: or create separate functions for the contract lookup 
```
let { vetoVotingContract: vvc, vetoVotingType } = vetoGuardContract;
const vetoVotingContract = getVetoContractByVotingType(vvc, vetoVotingType);
```

src/pages/DaoDashboard/Activities/index.tsx
christopherdancy: correct, this is true! It is also the case that these params will need to be checked either way. I guess is this a question about overall state management or if this will render improperly.

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
christopherdancy: Yes -- rendering this state is not so simple. Will be looking into this soon.

src/providers/Fractal/hooks/useFreezeData.ts
christopherdancy: I attempted one listener which updated the vetoVotingContract depending on its type. However, the filter object does not like the possible type differences.

src/hooks/DAO/useCastFreezeVote.ts
```diff
@@ -0,0 +1,32 @@
...
+import { useCallback, useEffect } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useTransaction } from '../../providers/Web3Data/transactions';
+
+const useCastFreezeVote = ({
+  vetoVotingContract,
+  setPending,
+}: {
+  vetoVotingContract: VetoERC20Voting | VetoMultisigVoting;
+  setPending: React.Dispatch<React.SetStateAction<boolean>>;
```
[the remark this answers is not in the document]
christopherdancy: I followed the pattern which appears in useCastVote.ts -- which one should be followed moving forward? Is this something which can be addressed at a later time?

christopherdancy:
Here is an image showing ui updates depending if the user has voting power.

image

sethhrbek (approved):

herbig (approved):
🚀🚀🚀🚀🚀🚀🚀

Da-Colon (approved):
Looks good to me

PR #821, opened 2022-12-16
Hotfix | Show correct DAO name on Node Page.

Da-Colon:
localhost_3000_(Desktop) (27)

### DIFF
--- a/src/components/ui/cards/DAOInfoCard.tsx
+++ b/src/components/ui/cards/DAOInfoCard.tsx
@@ -7,6 +7,7 @@ import {
   ArrowRightSm,
 } from '@decent-org/fractal-ui';
 import { BACKGROUND_SEMI_TRANSPARENT } from '../../../constants/common';
+import useDAOName from '../../../hooks/DAO/useDAOName';
 import { useCopyText } from '../../../hooks/utils/useCopyText';
 import useDisplayName from '../../../hooks/utils/useDisplayName';
 import { useFractal } from '../../../providers/Fractal/hooks/useFractal';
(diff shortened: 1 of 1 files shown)


src/components/ui/cards/DAOInfoCard.tsx
```diff
@@ -27,12 +28,18 @@ export function DAOInfoCard({
   numberOfChildrenDAO,
 }: IDAOInfoCard) {
   const {
-    gnosis: { daoName },
+    gnosis: {
+      safe: { address },
+      daoName,
```
herbig, line 33: what is this name? If it's not the DAO registry name..

src/components/ui/cards/DAOInfoCard.tsx
```diff
@@ -27,12 +28,18 @@ export function DAOInfoCard({
...
+    },
     account: {
       favorites: { favoritesList, toggleFavorite },
     },
   } = useFractal();
   const copyToClipboard = useCopyText();
+  const { daoRegistryName } = useDAOName({
+    address: address !== safeAddress ? safeAddress : undefined,
```
mobiusTripper-crypto, line 41: it seems like we should only be retrieving the daoName from one place, maybe 
```
  const { daoRegistryName } = useDAOName({
    address: address !== safeAddress ? safeAddress : address,
  });
```
and stop getting if from the gnosis object? (I have no idea what other consequences that may cause)

mobiusTripper-crypto (approved):
a comment was left

src/components/ui/cards/DAOInfoCard.tsx
Da-Colon: undefined I believe

src/components/ui/cards/DAOInfoCard.tsx
Da-Colon: I guess the thought here is to save on duplicate calls as renders happen.

src/components/ui/cards/DAOInfoCard.tsx
Da-Colon: is that what you meant?

Da-Colon: ah its not. useDAOName is used when its retrieved. so it goes through the same ens -> DAO name -> substring

herbig (approved):

Issue #810, opened 2022-12-15
Allow ETH transfer on the Sent Asset modal


mobiusTripper-crypto:
when unwrapped ETH is selected for transfer on the Send Assets dialog ... create a proposal with: `target` is the TO address,  `calldata` is empty and `value` is the amount of ETH to transfer.

Also add toast / close dialog on successCallback()

Issue #754, opened 2022-12-08
Bug: Unable to enter maximum amount for NumberInputs


mobiusTripper-crypto:
We are using NumberInput which maxes out at Javascript `number` type. Change to Input and limit to ethers --- constants.MaxInt256 (<- I think)

![Image](https://user-images.githubusercontent.com/99622829/206479405-fa1745d2-4cb1-47f3-a82b-9d23139fa949.png)

herbig:
@mobiusTripper-crypto @Da-Colon this is something we should probably fix before launch

sethhrbek:
@herbig @mobiusTripper-crypto @Da-Colon  -- I think the scope for this task should be as below. Leave feedback if you feel differently

1. Change from `NumberInput` to use normal string Input. We can use `NumberInput` as a guide for the new component.
2. This new input should:
a. Handle Decimals -- if the user doesn't enter a decimal, then the number entered is considered whole units (1 ETH, 2 ETH, etc)
3. Input should have validation in place to not allow more than "x" characters to the right of the decimal point, where "x" is the number of decimals in the ERC20
4. Don't allow negative numbers
5. Add min/max if necessary

**Nice to have:**
Get the ERC20 symbol (ETH, FRAC, etc) and include that as a "rightIcon" on the input (or whatever chakra calls it)

If we need this input to handle integers, make `hasDecimals` property. If this property is false, then we assume this input is an integer, and can parse like so.

---

Once this is done, we can either, in a different task:
1. Replace the relevant inputs in the app with this new component locally
2. Abstract this component to the Design system and import it from there to use in the app

mobiusTripper-crypto:
Upon further evaluation, I do not think min / max on the input is a good idea, as the value changes without any feedback to the user if we just change the value. I think using validation is a more appropriate solution to insure min / max values.

PR #826, opened 2022-12-18
Allow ETH transfer on the Sent Asset modal

mobiusTripper-crypto:
## Description
Added the ability to sent native ETH using the Send Assets Modal.

This incorporates (https://github.com/decent-dao/fractal-interface/issues/754) with the addition of the input component BigNumberInput. Upon this PRs testing and completion, this component can be use throughout the rest app where other big numbers are used.

## Notes
Disable USD switch and created issue to re-enable with correct functionality.
No toast / notification on transaction submit, created issue to add.

https://github.com/decent-dao/fractal-interface/issues/810

## Testing
From the Treasury page. Insure DAO has ETH in the treasury and the current user is a delegate or owner.
Open the Send Assets modal
Test all possible combination of input on the token amount input component, i.e. leading zeros, non number characters, copy / paste, etc.
Create a proposal to send native ETH using the Send Assets modal, and verify the ETH is transfer once the proposal is executed.
 
image


src/components/ui/BigNumberInput/BigNumberInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+      });
+      if (newValue !== event.target.value) {
+        resetCaretPositionForInput(event);
+      }
+      setInputValue(newValue);
+    },
+    [decimalPlaces, onChange, truncateDecimalPlaces]
+  );
+
+  // if the decimalPlaces change, need to update the value
```
mobiusTripper-crypto, line 87: When the asset is changed on the modal, we need to re-run the truncateDecimalPlaces. I don't feel like my use of the useEffect here is 100% correct. But I couldn't figure out how to trigger onChange on the input?

Da-Colon (approved):
I won't hold PR, if you want to just do it in another one.

src/components/ui/modals/SendAssetsModal.tsx
```diff
@@ -1,105 +1,71 @@
...
 
-  const isSubmitDisabled =
-    isEth(selectedAsset) || // remove this check once we are able to send ETH
-    !isValidAddress ||
-    !amountInput ||
-    overDraft;
+  const isSubmitDisabled = !isValidAddress || inputAmount?.bigNumberValue.isZero() || overDraft;
 
   const onSubmit = () => {
     sendAssets();
```
Da-Colon, line 67: Might be a quick add to go ahead and add toasts here using `useTransaction()` hook. see `src/components/Proposals/MultisigProposalDetails/TxActions.tsx`. its pretty straightforward

src/components/ui/BigNumberInput/BigNumberInput.tsx
Da-Colon: doesn't appear to effect the 'controlled' or 'uncontrolled' state of the Input. in my book this looks good and works well

src/pages/Treasury/hooks/useSendAssets.ts
```diff
@@ -16,12 +16,13 @@ const useSendAssets = ({
   const { submitProposal } = useSubmitProposal();
 
   const successCallback = () => {
-    //TODO: need to add a toast and close modal
+    //TODO: need to figure out what toast / notification to add
```
herbig, line 19: @Da-Colon mentioned there is a `useTransaction` hook that can manage pending / success toasts, but we can add that as a followup

src/components/ui/modals/SendAssetsModal.tsx
herbig: oh just noticed this comment, yeah could add in this PR or a followup

herbig (changes requested):
@mobiusTripper-crypto the USD slider no longer works, it's permanently disabled now for some reason

mobiusTripper-crypto:
> @mobiusTripper-crypto the USD slider no longer works, it's permanently disabled now for some reason

Yes, I think it needs to switch between the token amount and the USD value. the way it works now it just confusing (and I think doesn't actually work). I am creating a follow up issue.

herbig (approved):

src/pages/Treasury/hooks/useSendAssets.ts
```diff
@@ -31,16 +32,16 @@ const useSendAssets = ({
     ];
 
     const proposalData: ProposalExecuteData = {
-      targets: [tokenAddress],
-      values: [BigNumber.from('0')],
-      calldatas: calldatas,
+      targets: [isEth ? destinationAddress : asset.tokenAddress],
+      values: [isEth ? transferAmount : BigNumber.from('0')],
+      calldatas: isEth ? ['0x'] : calldatas,
       title: 'Send Assets',
```
herbig, line 38: just a note for @decent-dao/fractal, any on chain metadata we submit in the background (title/description) should be hardcoded in English, as this is.

translating these per the user's settings doesn't make sense, as they are not necessarily submitting it to a DAO that is specific to their language

Issue #807, opened 2022-12-15
Freeze / Veto Default Values


herbig:
Currently the default values in the SubDAO creation for both `multisig` and `Usul` for Freeze and Veto votes thresholds are set to 1.

Veto votes threshold is also hidden from the form via a Chakra `<Hide>` component, but will still submit its default value in the SubDAO creation.

We would like to set sensible defaults here, rather than 1, which should be:

**Multisig Parent**: the _Signature Threshold_ of the **ParentDAO** that is creating this **SubDAO** proposal.

**Usul Parent**: 50% of the total number of tokens in existence for the **ParentDAO** creating this **SubDAO** proposal.

Additionally, we'd like to adjust the helper text under these threshold inputs to say:

**helperFreezeVotesThreshold**
`Total votes required by the parent (out of {{totalVotes}}) to freeze this DAO entirely.`

**helperVetoVotesThreshold**
`Total votes required by the parent (out of {{totalVotes}}) to veto a proposal on this DAO.`

<img width="480" alt="Screenshot 2022-12-15 at 12 36 03 PM" src="https://user-images.githubusercontent.com/384559/207930578-e2468442-402f-4bf1-92c2-1471ab8741a7.png">

PR #827, opened 2022-12-19
dynamic freeze/veto thresholds

herbig:
## Description

Sets the Veto and Freeze Votes Threshold default values as half the total amount of voting power on the parent.

closes https://github.com/decent-dao/fractal-interface/issues/807

## Testing

Check the create child DAO flow via the Create Node option for both multisig and token voting, and ensure that the Freeze Votes Threshold field shows the default as half the total votes and that you can create the proposal successfully.

Screenshot 2022-12-18 at 8 40 42 PM

Screenshot 2022-12-18 at 8 41 18 PM


src/components/DaoCreator/GuardDetails.tsx
```diff
@@ -65,6 +74,55 @@ function GuardDetails() {
...
+
+  useEffect(() => {
+    if (!isInitialized) {
+      if (governanceIsLoading || !safe || !governanceToken) return;
+
+      let totalVotes: number;
+      switch (type) {
+        case GovernanceTypes.GNOSIS_SAFE_USUL:
+          totalVotes = parseInt(
+            ethers.utils.formatUnits(governanceToken.totalSupply!, governanceToken.decimals)
```
mobiusTripper-crypto, line 86: I guess I would rather see `governanceToken.totalSupply || '0'` here as opposed to the trailing ! (i might not be familiar enough on how that really works, as what if would the parameter value would be if it did happened to be undefined). I can't seem to reproduce it, but I was getting an error on this line reporting a value of "".

mobiusTripper-crypto (approved):
one comment that needs updating.

src/components/DaoCreator/GuardDetails.tsx
```diff
@@ -65,6 +74,55 @@ function GuardDetails() {
   const { t } = useTranslation(['daoCreate']);
   const votes = t('votes');
   const seconds = t('seconds');
+
+  useEffect(() => {
+    if (!isInitialized) {
```
mobiusTripper-crypto, line 79: I am no expert on useEffects, but I think if you want this to run only once on re-render, then simply removing all the dependencies (other than eslint complaining) will work the same as the isInitialized? maybe other people have opinions on this.

src/components/DaoCreator/GuardDetails.tsx
Da-Colon:
```suggestion
            ethers.utils.formatUnits(governanceToken.totalSupply || '0', governanceToken.decimals)
```
I agree (trying out feature)

herbig: good call, thanks for the easy commit button @Da-Colon !

src/components/DaoCreator/GuardDetails.tsx
herbig: Yeah I wasn't sure what the best way to do this would be.  I read about some lifecycle methods for components, but for functions the only thing I found was to have this initialized state.

It feels better than having lint errors that need to be ignored.. @Da-Colon any thoughts?

herbig: actually, since I'm using `totalParentVotes` I can just remove the initialized state and check that instead

src/components/DaoCreator/GuardDetails.tsx
```diff
@@ -66,6 +73,53 @@ function GuardDetails() {
   const { t } = useTranslation(['daoCreate', 'common', 'proposal']);
   const votes = t('votesTitle', { ns: 'proposal' });
   const seconds = t('seconds', { ns: 'common' });
+
+  useEffect(() => {
```
christopherdancy, line 77: Question? why do we need the useffect? could we just create a function which updates the data?

src/components/DaoCreator/GuardDetails.tsx
```diff
@@ -151,7 +205,7 @@ function GuardDetails() {
         <InputBox>
           <LabelWrapper
             label={t('labelFreezeVotesThreshold')}
-            subLabel={t('helperFreezeVotesThreshold')}
+            subLabel={freezeHelper}
```
christopherdancy, line 208: If we have time, we should also create a error which checks if the users added threshold is above the totalvotes.

src/components/DaoCreator/GuardDetails.tsx
herbig: yeah I think the useEffect is necessary here, data needs to be loaded in the background, not when you click something, so the function would need to be called in some way.  Also, it should update when `governanceToken` loads

src/components/DaoCreator/GuardDetails.tsx
```diff
@@ -66,6 +73,53 @@ function GuardDetails() {
...
+          );
+          break;
+        case GovernanceTypes.GNOSIS_SAFE:
+        default:
+          totalVotes = safe.owners?.length || 0;
+      }
+      setTotalParentVotes(totalVotes);
+
+      const childThresholds = Math.ceil(totalVotes / 2);
+      dispatch({
```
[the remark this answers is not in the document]
herbig: yeah good call. It does require two calls to it but I think better to reuse that method here

src/components/DaoCreator/GuardDetails.tsx
```diff
@@ -66,6 +73,53 @@ function GuardDetails() {
...
+    governance,
+    governanceIsLoading,
+    governanceToken,
+    safe,
+    safe.threshold,
+    totalParentVotes,
+    type,
+  ]);
+
+  const showVetoFreezeHelpers = totalParentVotes > 0;
```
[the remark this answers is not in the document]
herbig: yeah `totalParentVotes` should always be greater than zero.  If there is no veto guard this `GuardDetails` screen won't be shown, so I'm assuming >0.

The reason there is logic to hide it is for the case where data might still be loading for a fraction of a second, so we don't display it as 0, then update to the true amount.

src/components/DaoCreator/GuardDetails.tsx
herbig: good call.  This whole screen and most of DAO creation in general needs a lot more input validation...

I'll create an issue for validation as a whole, I think that could be a good project to tackle all together

herbig: https://github.com/decent-dao/fractal-interface/issues/846

src/components/DaoCreator/GuardDetails.tsx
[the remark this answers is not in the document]
herbig: shoot, @christopherdancy after pushing I see why I didn't call fieldUpdate, that caused a dependency chain that required more refactoring of `fieldUpdate`, so I think having the single dispatch with both params is better here
Screenshot 2022-12-21 at 1 56 02 PM

src/components/DaoCreator/GuardDetails.tsx
```diff
@@ -66,6 +73,53 @@ function GuardDetails() {
...
+  useEffect(() => {
+    if (totalParentVotes === 0) {
+      if (governanceIsLoading || !safe || !governanceToken) return;
+
+      let totalVotes: number;
+      switch (type) {
+        case GovernanceTypes.GNOSIS_SAFE_USUL:
+          totalVotes = parseInt(
+            ethers.utils.formatUnits(governanceToken.totalSupply || '0', governanceToken.decimals)
+          );
```
sethhrbek, line 86: Can `governanceToken.totalSupply` whole units be larger than max JS int (9007199254740991)? Do we allow user to create enter a number this large for tokens when setting up a DAO? If so, then may need to handle this case with Big Number division rather than casting to an int.

src/components/DaoCreator/GuardDetails.tsx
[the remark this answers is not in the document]
sethhrbek: This comment is outside of the scope of this PR, but regarding the `fieldUpdate` method.

`(value: any, field: string)`

Would recommend swapping the order of the args (just to follow key => value conventions),

And then I believe could type `value: BigNumber`, as the guard config has all `BigNumber` values

```
type DAOVetoGuardConfig = {
  executionPeriod: BigNumber;
  timelockPeriod: BigNumber;
  vetoVotesThreshold: BigNumber;
  freezeVotesThreshold: BigNumber;
  freezeProposalPeriod: BigNumber;
  freezePeriod: BigNumber;
};
```

src/components/DaoCreator/GuardDetails.tsx
```diff
@@ -66,6 +73,53 @@ function GuardDetails() {
...
+          break;
+        case GovernanceTypes.GNOSIS_SAFE:
+        default:
+          totalVotes = safe.owners?.length || 0;
+      }
+      setTotalParentVotes(totalVotes);
+
+      const childThresholds = Math.ceil(totalVotes / 2);
```
sethhrbek, line 94: This `int` value is expected to be a `BigNumber` when it is stored (see `DAOVetoGuardConfig`), however this may be fine and be auto-coerced at some point, just wanted to point it out.

src/components/DaoCreator/GuardDetails.tsx
christopherdancy: This is handled in my PR -- it is an input rather then a number input

christopherdancy (approved):

src/components/DaoCreator/GuardDetails.tsx
herbig: hmm, yeah you're right, I think we should handle this with BigNumbers

herbig: @sethhrbek https://github.com/decent-dao/fractal-interface/pull/827/commits/40b35ef22670141232b7d17d53f0436a6ea4fdb5

herbig: @christopherdancy which PR do you mean?

sethhrbek (approved):

src/components/DaoCreator/GuardDetails.tsx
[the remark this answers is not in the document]
herbig: done, good call

PR #832, opened 2022-12-19
Copy edits

herbig:
## Description

- Adds copy edits to all json strings in the app
- Deletes UK strings that have changed and would need to be retranslated
- adds helper text under allocations

Screenshot 2022-12-19 at 2 22 15 PM


src/components/Proposals/index.tsx
```diff
@@ -90,12 +85,7 @@ export default function Proposals() {
                 justifyContent="flex-end"
                 onClick={() => setFilters([])}
               >
-                <Text
-                  color="grayscale.100"
-                  fontStyle="text-sm-mono-regular"
-                >
-                  {t('clear', { ns: 'common' })}
-                </Text>
+                <Text color="grayscale.100">{t('clear', { ns: 'common' })}</Text>
```
herbig, line 88: these I noticed were using `fontStyle` rather than `textStyle`, so it wasn't actually doing anything.  Luckily the default textStyle was already what was needed there, so it wasn't noticed

mobiusTripper-crypto (approved):
I looked over the copy, I'm not much of a proof reader, so no guarantees on that stuff.

PR #835, opened 2022-12-19
Sent Asset bug fixes

mobiusTripper-crypto:
## Description
Fixed various bugs on the Send Asset Modal:

- removed USD switch from the app as we determine it's usefulness was limited.
- filtered assetsFungible to remove items with zero balance
- do not show Sent Assets button if there are no fungible assets to send
- added toast to Send Assets, to complete this I added toast to useSubmitProposal and then updated all references of that to include appropriate messages.

## Testing

- Verify Toast messages when submitting a new proposal, creating a sub-DAO, and sent assets.
- Verify Send Assets modal work correctly. That the list of assets in the the dropdown only contains items with a balance.
- Verify the Send Assets button does not appear on the Treasury page for a DAO with no assets.


src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -90,8 +108,14 @@ export default function useSubmitProposal() {
...
           );
           await getGnosisSafeTransactions();
-          successCallback(safe.address);
+          if (successCallback) {
+            successCallback(safe.address);
+          }
+          toast.dismiss(toastId);
+          toast(successToastMessage);
         } catch (e) {
+          toast.dismiss(toastId);
```
herbig, line 117: could put this dismiss in the finally rather than once in the try block and once here

src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -119,10 +143,16 @@ export default function useSubmitProposal() {
...
+          if (successCallback) {
+            successCallback(safe.address);
+          }
         } catch (e) {
+          toast.dismiss(toastId);
+          toast(failedToastMessage);
           logError(e, 'Error during Usul proposal creation');
         } finally {
+          toast.dismiss(toastId);
+          toast(successToastMessage);
```
herbig, line 155: should the success message be in finally? it seems like that will end up toasting success when there's an exception

src/i18n/locales/en/daoCreate.json
```diff
@@ -74,5 +74,8 @@
     "titleAllocations": "Initial Token Allocations",
     "titleTokenParams": "Token Metadata",
     "titleAddress": "Address",
-    "titleAmount": "Amount"
+    "titleAmount": "Amount",
+    "createSubDAOPendingToastMessage": "Deploying your sub-DAO",
+    "createSubDAOSuccessToastMessage": "Sub-DAO deployment successful",
+    "createSubDAOFailureToastMessage": "Sub-DAO deployment failed"
```
herbig, line 80: we killed the dash in the copy updates, so we're not calling it "SubDAO"

src/hooks/DAO/proposal/useSubmitProposal.ts
mobiusTripper-crypto: This is incorrect. Thanks for finding that.

src/i18n/locales/en/daoCreate.json
mobiusTripper-crypto: Fixed. copy is not my strong suit

src/hooks/DAO/proposal/useSubmitProposal.ts
mobiusTripper-crypto: yeah, i thought i needed to be this way, but it works the same in finally.

src/i18n/locales/en/daoCreate.json
herbig: yeah me neither, but now I have an idea of what the copy folks prefer, ha 

Thanks!

herbig (approved):
🛳

PR #839, opened 2022-12-20
disable button if user is not an owner or they have signed

christopherdancy:
## Description

Very simple PR which simply disables the approve button under the following circumstances:
1. user has already signed the tx
2. user is not an owner on the safe

image


src/components/Proposals/MultisigProposalDetails/TxActions.tsx
```diff
@@ -117,7 +117,7 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
   const hasReachedThreshold = proposal.confirmations.length >= (safe.threshold || 1);
 
   const isOwner = safe.owners?.includes(account || '');
-  const isPending = asyncRequestPending || contractCallPending;
+  const isPending = asyncRequestPending || contractCallPending || hasSigned || !isOwner;
```
mobiusTripper-crypto, line 120: I think i would put those checks on `isButtonActive`, as we are double checking `isOwner` which seem strange. And now that you have added these other checks, it doesn't seem like it's just testing `pending` state.

mobiusTripper-crypto (approved):
small comment

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
sethhrbek: we may need a few checks -- if it's executable, we will want to show the button, but if it isn't and they aren't the owner or they have signed, then return blank component. 

```
if (!isExecutable && (!isOwner || hasSigned) {
   return <></>
} 
```

sethhrbek (approved):

PR #840, opened 2022-12-20
Update | No account DAO viewing

Da-Colon:


mobiusTripper-crypto (approved):
looks good to me, but I did find this, when there are no proposals, the Create Proposal button works here. Seems like it should be hidden when not connected

image

Da-Colon:
> looks good to me, but I did find this, when there are no proposals, the Create Proposal button works here. Seems like it should be hidden when not connected
> 
> image

o darn missed this one good catch.

christopherdancy (approved):
This looks good. However, I am a tad confused on how we will alert users. It seems like they may be confused if not prompted to connect.

Da-Colon:
> This looks good. However, I am a tad confused on how we will alert users. It seems like they may be confused if not prompted to connect.

Since we no longer 'require' to be connected. I don't think we should prompt a user on DAO screens. We are still covering the Create DAO workflow I believe. but otherwise. they are just more or less viewing a DAO

herbig (approved):

PR #842, opened 2022-12-20
check if there are any votes before render

### LINKED ISSUE #833: Bug: Votes component showed when empty
### Overview

If there are no votes cast yet, we should not show this component

### Steps to reproduce

n/a

### Screenshots / Video

Screenshot 2022-12-19 at 2 56 39 PM

### Version

### Browser

Firefox

### Operating System

Mac

### web3 wallet

None

christopherdancy:
## Description

This simple PR checks if there are votes before rendering the vote history component.

## Notes

There are no listeners set up re-render state once a vote occurs. This requires a manual refresh to see the new state after a vote is completed.
## Testing

1. Create a usul dao
2. Create a random proposal
3. If you have not voted, the component should not render.

image

### DIFF
--- a/src/components/Proposals/ProposalVotes.tsx
+++ b/src/components/Proposals/ProposalVotes.tsx
@@ -163,28 +163,30 @@ function ProposalVotes({
           </GridItem>
         </Grid>
       </ContentBox>
-      <ContentBox bg={BACKGROUND_SEMI_TRANSPARENT}>
-        <Text textStyle="text-lg-mono-medium">{t('votesTitle', { ns: 'proposal' })}</Text>
-        <Divider
-          color="chocolate.700"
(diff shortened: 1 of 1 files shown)


herbig (approved):

mobiusTripper-crypto (approved):

mobiusTripper-crypto:
I've thought about this someone, maybe it would be more informative to the user to display `No Votes` message where the list is, instead of just completely hiding the box? Just another thought ....

herbig:
> I've thought about this someone, maybe it would be more informative to the user to display `No Votes` message where the list is, instead of just completely hiding the box? Just another thought ....

I think in this case just hiding the empty component works fine.  You can see from the right sidebar that there are no votes.

Issue #830, opened 2022-12-19
Bug: DAO Search improvements


herbig:
### Overview

1. Don't clear the field when you click on it
2. Don't clear the field when you type a space (though spaces shouldn't be allowed)
3. Pressing Enter while focused in search should click View DAO if it's there

### Steps to reproduce

n/a

### Screenshots / Video

<img width="602" alt="Screenshot 2022-12-19 at 1 14 15 AM" src="https://user-images.githubusercontent.com/384559/208360387-2185886b-6319-4f18-b9f5-79cb19e30261.png">

### Version

Dev (app.dev.fractalframework.xyz/)

### Browser

Firefox

### Operating System

Mac

### web3 wallet

None

Issue #780, opened 2022-12-14
Bug: Valid ENS names that are not DAOs don't give error message in search


herbig:
### Overview

See screenshots

### Steps to reproduce

See screenshots

### Screenshots / Video

<img width="568" alt="Screenshot 2022-12-13 at 8 07 47 PM" src="https://user-images.githubusercontent.com/384559/207479977-be4c4650-60f5-4ca2-9a26-533ae278b9fe.png">
<img width="618" alt="Screenshot 2022-12-13 at 8 07 06 PM" src="https://user-images.githubusercontent.com/384559/207479994-afa0c01c-673b-40bb-88bc-05147a0f3612.png">

### Version

Dev (app.dev.fractalframework.xyz/)

### Browser

Firefox

### Operating System

Mac

### web3 wallet

None

herbig:
@mudrila this also seems to be the case for any valid eth address:

<img width="549" alt="Screenshot 2022-12-19 at 1 16 55 AM" src="https://user-images.githubusercontent.com/384559/208360519-5e324599-075e-4438-b946-2893e473f52c.png">

herbig:
also just 'eth', weird
<img width="537" alt="Screenshot 2022-12-19 at 9 30 47 AM" src="https://user-images.githubusercontent.com/384559/208448980-57e9cda1-18cf-4187-948a-c4a9fd2c5a32.png">

herbig:
hey @mudrila not sure how much you've looked into this already, but I'm making changes to the search component and will fix this in the process

PR #844, opened 2022-12-21
Search refactor

herbig:
## Description

A bit of a refactor of components involved in DAO search to improve usability.

closes https://github.com/decent-dao/fractal-interface/issues/830
closes https://github.com/decent-dao/fractal-interface/issues/780
Also fixes attempting to load an address via the URL manually for a non Safe
Search field no longer requires clicking twice to focus it

Screenshot 2022-12-21 at 1 39 50 AM

### DIFF
--- a/src/components/ui/menus/DAOSearch/index.tsx
+++ b/src/components/ui/menus/DAOSearch/index.tsx
@@ -8,8 +8,9 @@ import {
   InputLeftElement,
 } from '@chakra-ui/react';
 import { Search } from '@decent-org/fractal-ui';
-import { useCallback, useRef } from 'react';
+import { useCallback, useEffect, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
+import { useLocation } from 'react-router-dom';
 import { useSearchDao } from '../../../../hooks/DAO/useSearchDao';
 import { SearchDisplay } from './SearchDisplay';
 
@@ -44,6 +45,12 @@ export function DAOSearch() {
     unFocusInput();
   };
 
+  const location = useLocation();
+  useEffect(() => {
+    setSearchString('');
+    unFocusInput();
+  }, [location, setSearchString]);
+
   return (
     <Box
       width="full"
@@ -53,8 +60,8 @@ export function DAOSearch() {
       <Menu
         matchWidth
         isLazy
-        defaultIsOpen={true}
         onOpen={focusInput}
+        onClose={unFocusInput}
       >
         <MenuButton
           h="full"
--- a/src/components/ui/menus/DAOSearch/index.tsx
+++ b/src/components/ui/menus/DAOSearch/index.tsx
@@ -8,30 +8,18 @@ import {
   InputLeftElement,
 } from '@chakra-ui/react';
 import { Search } from '@decent-org/fractal-ui';
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
-import useSearchDao from '../../../../hooks/DAO/useSearchDao';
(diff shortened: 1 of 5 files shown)


src/hooks/DAO/useSearchDao.ts
```diff
@@ -1,92 +1,44 @@
...
-
-  /**
-   * handles loading state of search
-   */
   useEffect(() => {
-    if (isAddressLoading !== undefined) {
-      setLoading(isAddressLoading || isGnosisSafeLoading);
-    }
-  }, [isAddressLoading, isGnosisSafeLoading]);
+    setIsLoading(isAddressLoading !== false || isSafeLoading === true);
```
christopherdancy, line 16: nitpick -- does this statement {isAddressLoading !== false || isSafeLoading === true} === {isAddressLoading || isGnosisSafeLoading}

if not, how is it diff?

christopherdancy (approved):
Left one comment but it is not blocking... great job!

src/hooks/safe/useIsSafe.ts
```diff
@@ -4,29 +4,29 @@ import { useEffect, useState } from 'react';
...
+export const useIsGnosisSafe = (address: string | undefined) => {
+  const [isSafeLoading, setSafeLoading] = useState<boolean>(false);
   const [isSafe, setIsSafe] = useState<boolean>();
   const {
     state: { chainId },
   } = useWeb3Provider();
 
   useEffect(() => {
+    setSafeLoading(true);
+    setIsSafe(undefined);
```
sethhrbek, line 16: Do we need the triple state (undefined/true/false)? If not, then could the `isSafe` state var default to `false` so this line is uneccessary?

src/components/ui/menus/DAOSearch/index.tsx
```diff
@@ -46,17 +35,21 @@ export function DAOSearch() {
...
+
+  const location = useLocation();
   useEffect(() => {
-    return () => {
-      resetErrorState();
-    };
-  });
+    setSearchString('');
+    unFocusInput();
+  }, [location, setSearchString]);
```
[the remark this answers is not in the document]
herbig: yeah good call, will update

src/hooks/DAO/useSearchDao.ts
herbig: yeah, as @sethhrbek said, I simplified this and just set it to be if either loading state is true or undefined

`useAddress` is used in a couple other places, so I didn't want to go refactor that and break something else, but ideally it would be as simple as `isAddressLoading || isSafeLoading`

Automated testing would make refactoring things like that a lot easier

src/hooks/safe/useIsSafe.ts
sethhrbek: Good point -- if we treat a valid address that is *not* a safe differently from a non-valid address, then this should be separate. Would recommend updating the type to be `boolean | undefined` to be explicit about this.

sethhrbek (approved):

src/hooks/safe/useIsSafe.ts
```diff
@@ -4,29 +4,40 @@ import { useEffect, useState } from 'react';
...
+ * Safe smart contract address on the currently connected chain (chainId).
+ *
+ * The state can be either true/false or undefined, if a network call is currently
+ * being performed to determine that status.
+ *
+ * @param address the address to check
+ * @returns isSafe: whether the address is a Safe,
+ *  isSafeLoading: true/false whether the isSafe status is still being determined
+ */
+export const useIsGnosisSafe = (address: string | undefined) => {
```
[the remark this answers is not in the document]
herbig: whoops, missed this before I merged, but yeah `address?` is better..

PR #850, opened 2022-12-22
Unable to enter maximum amount for NumberInputs

mobiusTripper-crypto:
## Description
Where an token amount is input in the UI, swap `NumberInput` with `BigNumberInput`.

## Notes
I changed the `value` type of `BigNumberInput` to be a `BigNumber` as it seems like in most cases returning and saving `BigNumberValuePair` was unnecessary. If a consumer of the component needs the converted numeric value they can retrieve that from the `onChange` since is still returns the `BigNumberValuePair`.

'ethers.utils.formatUnits()` adds a trailing .0 when it returns. we can create an issue to fix that if we think it's a problem. (will show the trailing 0 on re-render so from the token allocation page, prev page, then return will show the 0.

A new issue is created to replace the remaining NumberInput, but these are whole number and BigNumberInput needs a few tweaks to handle that case.

## Issue 
https://github.com/decent-dao/fractal-interface/issues/754

## Testing
Pages to test where there new component is implemented.

Create new Token Voting DAO, TokenAllocations and Token Amount. Verify 18 decimal numbers can be entered, and the proper values are used to create the new DAO
Treasury -> Send Asset


src/components/DaoCreator/provider/types/index.ts
```diff
@@ -115,14 +110,14 @@ export interface SubDAO extends GnosisConfig, TokenGovernanceDAO {
 export interface TokenGovernanceDAO extends DAODetails {
   tokenName: string;
   tokenSymbol: string;
-  tokenSupply: BigNumberInput;
+  tokenSupply: BigNumber;
```
herbig, line 113: nice, yeah being able to use `BigNumber` is great

src/components/DaoCreator/tokenVotingGovernance/TokenAllocations.tsx
```diff
@@ -1,20 +1,18 @@
...
 import InputBox from '../../ui/forms/InputBox';
-import { DEFAULT_TOKEN_DECIMALS } from '../provider/constants';
-import { BigNumberInput } from '../provider/types';
 import TokenAllocationInput from './TokenAllocationInput';
 
 interface TokenAllocationsProps {
   tokenAllocations: TokenAllocation[];
-  supply: BigNumber | null;
-  parentAllocationAmount?: BigNumberInput;
+  supply: BigNumber | undefined;
```
herbig, line 14: `supply?: BigNumber;` is the same here right? @mudrila recently called me on that, ha.  It's a nitpick, but probably better for consistency

src/components/DaoCreator/tokenVotingGovernance/TokenAllocations.tsx
```diff
@@ -44,59 +40,43 @@ function TokenAllocations({
...
     fieldUpdate(updatedTokenAllocations, 'tokenAllocations');
   };
 
-  const onParentAllocationChange = (value: string) => {
-    fieldUpdate(
-      { value, bigNumberValue: utils.parseUnits(value || '0', DEFAULT_TOKEN_DECIMALS) },
-      'parentAllocationAmount'
-    );
+  const onParentAllocationChange = (value: BigNumberValuePair) => {
+    fieldUpdate(value.bigNumberValue, 'parentAllocationAmount');
```
herbig, line 63: I realize you didn't create this piece initially, but have these existing `fieldUpdate` functions all over and @sethhrbek recently pointed out that we should refactor these to be `fieldUpdate(key, value)` rather than the other way around, so would be good to do that in this PR while you're here.

src/components/ui/BigNumberInput.tsx
```diff
@@ -1,25 +1,30 @@
...
+export interface BigNumberInputProps
+  extends Omit<InputElementProps, 'value' | 'onChange'>,
+    FormControlOptions {
+  value: BigNumber | undefined;
   onChange: (value: BigNumberValuePair) => void;
   decimalPlaces?: number;
 }
 
+// the value property can be either a BigNumber or a BigNumberValuePair
+// choose if you need to use the value converted to a Number, then use the Pair, otherwise, just use a BigNumber
```
herbig, line 19: If you do a `/** */` style comment here rather than `//` VS code will show you the documentation on hover

src/components/ui/BigNumberInput.tsx
```diff
@@ -73,7 +75,7 @@ export function BigNumberInput({
       }
 
       onChange({
-        value: removeOnlyDecimalPoint(newValue),
+        value: Number(removeOnlyDecimalPoint(newValue)),
```
herbig, line 78: @mobiusTripper-crypto maybe I'm missing something, why is this converted to a number here?

herbig (approved):
just a few minor suggestions but works great, looks good

Issue #838, opened 2022-12-20
Bug: Child Multisig Doesn't Recognize Veto Guard


herbig:
### Overview

In testing the proposal flow for childDAO multisig, I noticed that approved proposals go straight to Queued state.

A transaction should be required to enqueue the proposal, after which there is the queued timer, then it is executable.

### Steps to reproduce

1. Create and approve a proposal on a multisig childDAO (you can use `childmultisig.eth`, however note that only one signature is required to approve, so it's not a bug that it goes right to approved after creation).
2. Observe that the proposal says it is in the `Queued` state.  There is no queue timer, the proposal will stay in this state indefinitely.
3. The execute button appears, but attempting to execute fails (presumably because of the veto guard).

Multisig proposals that are approved should have a "Queue" button, which calls the contract to Queue it, after which there is no action until the queued timer runs out, then it is executable.

### Version

Dev (app.dev.fractalframework.xyz/)

### Browser

All

### Operating System

None

### web3 wallet

None

PR #853, opened 2022-12-22
Multisig tx state

tbwebb22:
## Description
This PR:
1) Updates the proposal TX logic for multisig DAOs to support the concept of queuing, queued state, queuable state, expired state, and executing state.
2) Adds a 'queue' button to the proposal details page when the TX is in the 'queuable' state

https://github.com/decent-dao/fractal-interface/issues/838

## Testing
1) Create a root multisig DAO
2) Test creating, signing and executing proposals on this DAO
3) From this parent DAO, create a child DAO that will have the VetoGuard attached
4) Test creating, signing, queueing, executing proposals on this child DAO. 
5) Test edge cases such as allowing an 'executing' proposal to exceed its execution period and become 'expired'


src/hooks/utils/useSafeActivitiesWithState.tsx
```diff
@@ -0,0 +1,162 @@
...
+                multiSigTransaction.refundReceiver as string
+              );
+
+              const queuedTimestamp = (
+                await vetoGuard.getTransactionQueuedTimestamp(vetoGuardTransactionHash)
+              ).toNumber();
+
+              const lastBlockTimestamp = provider
+                ? (await provider.getBlock(await provider.getBlockNumber())).timestamp
+                : 0;
```
tbwebb22, line 70: @Da-Colon @sethhrbek We need this to be able to update on every new block so this app updates without refreshing. Any thoughts on where to do this? I know @Da-Colon had some thoughts on keeping it close to this file so its not causing large portions of the app to re-render every block

src/i18n/locales/en/proposal.json
```diff
@@ -23,6 +23,8 @@
     "stateTimeLocked": "Time Locked",
     "stateFreezeInit": "Freeze Initiated",
     "stateFrozen": "Frozen",
+    "stateQueuable": "Queuable",
+    "stateExpired": "Expired",
```
tbwebb22, line 27: @mudrila Looking for UK translations for these

src/hooks/utils/useSafeActivitiesWithState.tsx
```diff
@@ -0,0 +1,162 @@
...
+                    multiSigTx.nonce === eventNonce &&
+                    multiSigTx.safeTxHash !== multiSigTransaction.safeTxHash &&
+                    multiSigTx.isExecuted
+                  );
+                });
+
+              const isQueuable = multiSigTransaction.confirmations
+                ? multiSigTransaction.confirmations?.length >=
+                  multiSigTransaction.confirmationsRequired
+                : false;
```
sethhrbek, line 49: this could be more terse:
```
const isQueueable = (multiSigTransaction.confirmations?.length || 0) >= multiSigTransaction.confirmationsRequired
```

src/i18n/locales/en/proposal.json
herbig: should be `stateQueueable` though!

src/hooks/utils/useSafeActivitiesWithState.tsx
sethhrbek: Yeah this will require some thought -- one thing is that this is going to get calculated on every `activity` (since it's in a loop) --- this should get moved out of the loop since we should only need to calculate it once.

Then need to figure out most optimized way to determine when this should be calculated and where the interval should live.

src/i18n/locales/en/proposal.json
tbwebb22: Oh man you're right. How many words have 5 vowels in a row lol. Just pushed up this fix

src/hooks/utils/useSafeActivitiesWithState.tsx
tbwebb22: Good call, pushed this up

src/hooks/utils/useSafeActivitiesWithState.tsx
tbwebb22: David and I just came up with a simple solution to this that I just pushed up. 

Since this entire `useSafeActivitiesWithState` hook will re-render every 20 seconds anyways - due to the Gnosis service API call being made, we don't need to put an interval on this for getting each new block. This was just moved up into the first Promise.all

sethhrbek: @tbwebb22 -- sounds great, thanks for the explanation!

src/hooks/utils/useSafeActivitiesWithState.tsx
```diff
@@ -0,0 +1,159 @@
...
+            const isMultiSigTransaction = activity.transaction.txType === 'MULTISIG_TRANSACTION';
+            const isModuleTransaction = activity.transaction.txType === 'MODULE_TRANSACTION';
+            const multiSigTransaction =
+              activity.transaction as SafeMultisigTransactionWithTransfersResponse;
+
+            const eventNonce = multiSigTransaction.nonce;
+
+            const isRejected =
+              isMultiSigTransaction &&
+              activityArr.find(_activity => {
```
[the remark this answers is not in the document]
sethhrbek: Actually, now that I'm re-reading this,-- looks like this function is trying to find some other activity in the array that has same nonce but a different hash.

In this case, do we know the TX was rejected, or could it have just been replaced with a different proposal with the same nonce? I guess if we treat a replaced proposal TX as "rejected", then this should be fine, just wanted to make sure we didn't need to differentiate between rejected + replaced.

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
```diff
@@ -62,6 +65,46 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
     }
   };
 
+  const queueTransaction = async () => {
+    try {
+      if (!multisigTx || !multisigTx.confirmations) {
+        return;
+      }
```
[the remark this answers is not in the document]
tbwebb22: Cool, yeah just cleaned these dependencies up

src/hooks/utils/useSafeActivitiesWithState.tsx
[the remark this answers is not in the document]
Da-Colon: > could it have just been replaced with a different proposal with the same nonce.

The second this happens all other proposals with that same nonce are considered 'rejected'.

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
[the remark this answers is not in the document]
Da-Colon: yeah if these methods are using state at all should be coverted to `useCallback`s good call

src/hooks/utils/useSafeActivitiesWithState.tsx
```diff
@@ -0,0 +1,159 @@
...
+              multiSigTransaction.to,
+              multiSigTransaction.value,
+              multiSigTransaction.data as string,
+              multiSigTransaction.operation,
+              multiSigTransaction.safeTxGas,
+              multiSigTransaction.baseGas,
+              multiSigTransaction.gasPrice,
+              multiSigTransaction.gasToken,
+              multiSigTransaction.refundReceiver as string
+            );
```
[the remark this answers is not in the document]
tbwebb22: Yeah good call, just pushed up this change

src/hooks/utils/useSafeActivitiesWithState.tsx
```diff
@@ -0,0 +1,159 @@
...
+        provider.getBlockNumber().then((blockNumber: number) => provider.getBlock(blockNumber)),
+      ]).then(([timelockPeriod, executionPeriod, lastBlock]) => {
+        Promise.all(
+          activities.map(async (activity, _, activityArr) => {
+            if (activity.eventType !== ActivityEventType.Governance || !activity.transaction) {
+              return activity;
+            }
+
+            const isMultiSigTransaction = activity.transaction.txType === 'MULTISIG_TRANSACTION';
+            const isModuleTransaction = activity.transaction.txType === 'MODULE_TRANSACTION';
```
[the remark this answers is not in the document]
tbwebb22: Cool, just pushed it up

src/hooks/utils/useSafeActivitiesWithState.tsx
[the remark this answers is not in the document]
tbwebb22: Yeah, I think this should be fine as is. Since if the other TX with same nonce but different hash was executed, then our current tx is rejected. I think of rejected and replaced as the same thing.

src/hooks/utils/useSafeActivitiesWithState.tsx
```diff
@@ -0,0 +1,159 @@
...
+              isMultiSigTransaction &&
+              activityArr.find(_activity => {
+                const multiSigTx =
+                  _activity.transaction as SafeMultisigTransactionWithTransfersResponse;
+                return (
+                  multiSigTx.nonce === eventNonce &&
+                  multiSigTx.safeTxHash !== multiSigTransaction.safeTxHash &&
+                  multiSigTx.isExecuted
+                );
+              });
```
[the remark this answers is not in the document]
tbwebb22: Cool, just moved isRejected and isApproved checks into functions in a separate helper file

sethhrbek (approved):
There are

src/components/Proposals/MultisigProposalDetails/index.tsx
```diff
@@ -30,7 +37,12 @@ export function MultisigProposalDetails({ proposal }: { proposal: TxProposal })
       </GridItem>
       <GridItem colSpan={1}>
         <TxDetails proposal={txProposal} />
-        {account && <TxActions proposal={txProposal} />}
+        {account && (
+          <TxActions
+            proposal={txProposal}
+            vetoGuard={vetoGuardContract as VetoGuard}
```
herbig, line 43: we shouldn't change it here, since that's what the contracts are call, but when we inevitably do some contract refactoring it would be great to have those be named like "MultisigVetoGuard" along with the existing "UsulVetoGuard"

src/providers/Fractal/governance/hooks/useUsulProposals.ts
```diff
@@ -208,19 +208,17 @@ export default function useUsulProposals({ governance, governanceDispatch }: IUs
         (prev, proposal) => (proposal.state === TxProposalState.Executed ? prev + 1 : prev),
         0
       );
-      const pendingProposals = mappedProposals.reduce(
-        (prev, proposal) =>
-          proposal.state === TxProposalState.Active || proposal.state === TxProposalState.Pending
-            ? prev + 1
-            : prev,
+      // @todo no queued?
```
herbig, line 211: @tbwebb22 what's this `todo` mean?

src/hooks/utils/useSafeActivitiesWithState.tsx
```diff
@@ -0,0 +1,175 @@
...
+            }
+
+            if (activity.transaction.txType === 'MODULE_TRANSACTION') {
+              return {
+                ...activity,
+                state: TxProposalState.Module,
+              };
+            }
+
+            const isMultiSigTransaction = activity.transaction.txType === 'MULTISIG_TRANSACTION';
```
herbig, line 39: this can be moved above the previous if statement and used there

herbig: oh nevermind, scratch that, I misread...

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -393,11 +393,11 @@ const useBuildDAOTx = () => {
 
           // Veto Guard
           const deployVetoGuardTx = await buildVetoGuardData({
-            executionPeriod: subDAOData.executionPeriod.mul(TIMER_MULT),
```
herbig, line 396: 🤦

herbig (approved):
🚀🚀🚀

src/providers/Fractal/governance/hooks/useUsulProposals.ts
tbwebb22: Hmm looks like David left that comment. Seth and David and I were talking about that this morning, specificallly the data we're getting here for this component 
image
We may want to show proposal counts for more than just Pending and Passed

PR #855, opened 2022-12-23
Update | Readd Multisig DAO creation test and setup API Mocking

Da-Colon:
## Description
Updates tests to mock initial DAO creation requests

## Notes
Sets up pattern to intercept and mock response from api request to gnosis api

- fixes flaky external link tests
- updates `safeService` to use network configs for setup


tests/models/mock/data/testSigners.ts
```diff
@@ -0,0 +1,22 @@
...
+  '0x71bE63f3384f5fb98995898A86B02Fb2426c5788',
+  '0xFABB0ac9d68B0B445fB7357272Ff202C5651694a',
+  '0x1CBd3b2770909D4e10f157cABC84C7264073C9Ec',
+  '0xdF3e18d64BC6A983f673Ab319CCaE4f1a57C7097',
+  '0xcd3B766CCDd6AE721141F452C550Ca635964ce71',
+  '0x2546BcD3c84621e976D8185a91A922aE77ECEc30',
+  '0xbDA5747bFD65F08deb54cb465eB87D40e51B197E',
+  '0xdD2FD4581271e230360230F9337D5c0430Bf44C0',
+  '0x8626f6940E2eb28930eFb4CeF49B2d1F2C9C1199',
+];
```
sethhrbek, line 22: Do these addresses have any significance or are they random addresses? Also looks like only the first address (address[0]) is being used, but assuming having more will come in handy later.

tests/models/mock/data/creation.ts
```diff
@@ -0,0 +1,72 @@
...
+  },
+};
+
+export const createSafeInfo = (safeAddress: string, owners: string[]) => {
+  return {
+    address: safeAddress,
+    nonce: 7,
+    threshold: 1,
+    owners: owners,
+    masterCopy: '0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552',
```
sethhrbek, line 61: This same address is used earlier in the file --- is this also used elsewhere in the application? Wondering if this (and potentially other addresses here) should be named consts

tests/models/mock/NewDAOMockRequests.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export class NewDAOMockRequests extends GnosisRoutes {
+  async newDAORoutes() {
+    // @dev this function is usuful for debughing requests, uncomment as needed
+    // this.page.on('request', request => console.log('>>', request.method(), request.url()));
+    await this.page.route(this.routeGetSafeCreationInfo, async route => {
+      await route.fulfill({
+        body: JSON.stringify(SINGLE_OWNER_MULTISIG_CREATION),
+      });
+    });
```
sethhrbek, line 17: If this was to be updated to handle multiple cases (single & multi-owner), how would the pattern of these mocks be modified?

Would this use a single route, but pass a param to `newDAORoutes` to indicate it should use a different creation object?

src/hooks/safe/useIsSafe.ts
```diff
@@ -19,25 +17,25 @@ export const useIsGnosisSafe = (address: string | undefined) => {
...
-  } = useWeb3Provider();
+    gnosis: { safeService },
+  } = useFractal();
 
   useEffect(() => {
     setSafeLoading(true);
     setIsSafe(undefined);
 
-    if (!address || !isAddress(address)) {
+    if (!address || !isAddress(address) || !safeService) {
```
herbig, line 27: In the case of `!safeService` should isSafe be left as `undefined`?  We don't know in that case whether it's a safe or not.

tests/models/mock/data/testSigners.ts
Da-Colon: yessir. and these accounts are the default created in the local hardhat environment. I'll see if I can extract these from the package versus creating a new object

tests/models/mock/data/creation.ts
Da-Colon: There is some work to do here. currently all the contract addresses in `providers/networkConfig/networks/` are still hardcoded. Might be good to knock this out in a seperate PR. FYI This is GnosisSafe IMPL address. We currently don't include this in any NPM package? I'm not sure if we are pulling GnosisDeployments NPM into this codebase

src/hooks/safe/useIsSafe.ts
Da-Colon: This is only replacing the api call. doesn't affect functionality. Once SafeService is set the check will happen as normal.

tests/models/mock/data/creation.ts
sethhrbek: This is in the localhost network config file, maybe could grab it from there

Da-Colon: o yeah good call, I'll do that for now.

tests/models/mock/NewDAOMockRequests.ts
Da-Colon: So these interceptors are set at the start of each test and cannot be dynamically updated. It must be set prior to Test starting. So for each route that we expect certain response from. we need create a new set of interceptors taylored specifically to what we expect the data to be for that test. 

So in that case we would need response and set of routes for each case that we want to test.

Multisig (Single Owner),
Multisig (Multiple Owners),
TokenVoting (Single Owner),
TokenVoting (Multiple Owners)

and repeat for any additional requests we need back. 

The one issue that may come up as we run these tests are the intervals. As we create a test for subDAO creation will be know more about the challenges that are faced when trying to make it through the happy path correctly.

https://playwright.dev/docs/mock

https://playwright.dev/docs/network

sethhrbek: Would it then make sense to make these mocked responses more targeted per test, rather than setting them all up at one time? If there's any that will be mocked the same no matter what, then we could do a `beforeAll` for the entire test suite so that they get intercepted for all tests.

Then for specific tests that require specific mocked responses, the interceptor would be part of the test itself. That way it is explicit which tests require which mocked responses.

For the interval issue -- where do you anticipate there to be a potential issue? I would think with the mocked responses those would be fine, but are you thinking with contract calls made via RPC?

sethhrbek: Or if a group of tests require a specific mocked response (for example, if a group of tests require a single multi-sig owner), then a `beforeAll` or `beforeEach` could be written around that group of tests with the interceptor.

Da-Colon: So I think what you are refering to is setting up whats called 'Fixtures' it basically does the same thing we are doing with the models except that there an extra step of extending the test object itself. see https://playwright.dev/docs/test-fixtures

I think we might need to talk about this.

sethhrbek: Yeah that looks promising! Could setup fixtures for each DAO type, and then override them as needed.

If we could setup basic fixtures to start, then when devs start writing tests for their pages, they could use these pre-defined fixtures as a template. Edge cases will pop up which may need special mocked responses requiring overrides can be added on-demand as they are uncovered.

Da-Colon:
Sorry for the delay, `playwright` did not like that import to NetworkConfig type locally, was throwing a misleading error. Should be good to go based on our discussions

Da-Colon:
Yessir  @mudrila. This is ready for final review. The plan is @herbig will be taking on the fixtures setup from here (in a new PR).

tests/models/mock/data/testSigners.ts
Da-Colon: We don't have hardhat installed. I think it will be fine to remain hardcoded for now. Those default haven't changed in over 2 years

src/hooks/utils/useUpdateTimer.tsx
```diff
@@ -15,7 +15,8 @@ export const useUpdateTimer = (safeAddress?: string) => {
     []
   );
   useEffect(() => {
-    if (!safeAddress) {
+    if (!safeAddress || process.env.REACT_APP_TESTING_ENVIROMENT) {
+      console.log('HERE');
```
sethhrbek, line 19: this could get yoinked

sethhrbek (approved):

src/hooks/utils/useUpdateTimer.tsx
Da-Colon: yoinked

herbig (approved):
LGTM, I'll take up building out generic mocking in a followup PR!

PR #856, opened 2022-12-23
Replace NumberInput with BigNumberInput component

mobiusTripper-crypto:
## Description
Replace all existing NumberInput that hold a Bignumber value with BigNumberInput input component.

## Testing
DAO creation (both types and children) need to be tested as I have replaced most of the input components on those pages.


src/components/ui/BigNumberInput.tsx
```diff
@@ -0,0 +1,163 @@
...
+}
+
+export interface BigNumberInputProps
+  extends Omit<InputElementProps, 'value' | 'onChange'>,
+    FormControlOptions {
+  value: BigNumber | undefined;
+  onChange: (value: BigNumberValuePair) => void;
+  decimalPlaces?: number;
+  min?: string;
+  max?: string;
```
herbig, line 17: there's no point to making min/max BigNumbers is there? That would probably just clutter the code

src/components/ui/BigNumberInput.tsx
```diff
@@ -0,0 +1,163 @@
...
+}
+/**
+ * This component will add a chakra Input component that accepts and sets a BigNumber
+ *
+ * @param value input value to the control as a BigNumber. If undefined is set then the component will be blank.
+ * @param onChange event is raised whenever the component changes. Sends back a value / BigNumber pair. The value sent back is a string represnetation of the BigNumber as a decimal number.
```
herbig, line 23: just a typo on `representation`

src/components/ui/BigNumberInput.tsx
```diff
@@ -0,0 +1,163 @@
...
+ * @param decimalPlaces number of decimal places to be used to parse the value to set the BigNumber
+ * @param min Setting a minimum value will reset the Input value to min when the component looses focus. Can set decimal number for minimum, but must respect the decimalPlaces value.
+ * @param max Setting this will cause the value of the Input control to be reset to the maximum when a number larger than it is inputted.
+ * @parma ...rest component accepts all properties for Input and FormControl
```
herbig, line 27: ah cool, yeah I keep forgetting about this.  We should add `...rest` to a lot of our other custom components

src/components/ui/BigNumberInput.tsx
mobiusTripper-crypto: I don't think we was them as bn, as the decimals can change on the fly. I wanted them to be number, but that doesn't work either.

src/components/ui/BigNumberInput.tsx
```diff
@@ -0,0 +1,163 @@
...
+ * @returns
+ */
+export function BigNumberInput({
+  value,
+  onChange,
+  decimalPlaces = 18,
+  min = '0',
+  max = constants.MaxUint256.toString(),
```
herbig, line 35: for the Token Supply input, since the decimals is 18 and the max is constants.MaxUint256.toString(), I can input higher than the maxUint will ultimately be.

e.g., the max UINT on chain with 18 decimals should be (notice decimal place):

115792089237316195423570985008687907853269984665640564039457.584007913129639935

which then gets converted to the max int value

But I can input more than that in the token supply

src/components/ui/BigNumberInput.tsx
herbig: and `looses`!

src/components/ui/BigNumberInput.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const initialValue = value
+    ? !value.isZero()
+      ? removeTrailingZeros(utils.formatUnits(value, decimalPlaces))
+      : '0'
+    : '';
+
+  const [inputValue, setInputValue] = useState<string>(initialValue);
+
+  // this will insure the caret in the input button does not shift to the end of the input when the value is changed
+  const resetCaretPositionForInput = (event: React.ChangeEvent<HTMLInputElement>) => {
```
herbig, line 56: it seems like the cursor is reset to one off when you input more than the max, like if max is 1000 and I type 1001, my cursor will be 100{here}1

src/components/ui/BigNumberInput.tsx
herbig: cool, yeah that makes sense

src/components/ui/BigNumberInput.tsx
mobiusTripper-crypto: I think that worked at one point, I'll fix it so it is correct.

herbig: yeah, I'm on Firefox, if that impacts anything

src/components/DaoCreator/GuardDetails.tsx
```diff
@@ -135,19 +99,16 @@ function GuardDetails() {
...
-                  <NumberInputField />
-                  <InputRightElement mr="4">{minutes}</InputRightElement>
-                </InputGroup>
-              </NumberInput>
+              <InputGroup>
+                <BigNumberInput
+                  value={vetoGuard.timelockPeriod}
+                  onChange={value => fieldUpdate('timelockPeriod', value.bigNumberValue)}
+                  decimalPlaces={0}
+                  min="1"
```
herbig, line 107: honestly pretty edge case but if we wanted to be exact here, `max` I think should be `constants.MaxUint256.div(60)` for all these time fields

src/components/ui/BigNumberInput.tsx
mobiusTripper-crypto: I'm not sure I can fix this one(I mean I CAN, i'm not sure I'm willing to add the amount of code for this edge case), as changing it for max case ruins it for when entering too many decimals

src/components/DaoCreator/GuardDetails.tsx
mobiusTripper-crypto: oh right, cause we multiply those values by 60. really if we switch to validation, then just pick a reasonable max number to validate against and can call it a day ...

herbig (approved):

PR #857, opened 2022-12-23
small bug fixes

herbig:
## Description

- lower max multisig signers to 100, since it still causes jankiness at 1000
- fix gov type hardcoded to English in dashboard and display Token Voting rather than Usul
- make Treasury and Proposals cards clickable in Dashboard


src/pages/DaoDashboard/Info/InfoGovernance.tsx
```diff
@@ -5,7 +5,7 @@ import { BarLoader } from '../../../components/ui/loaders/BarLoader';
 import { useFractal } from '../../../providers/Fractal/hooks/useFractal';
 
 export function InfoGovernance() {
-  const { t } = useTranslation('dashboard');
+  const { t } = useTranslation(['dashboard', 'daoCreate']);
```
herbig, line 8: `daoCreate` is a weird namespace to use on Dashboard, but I think string namespaces overall could use a refactor and didn't want to do that here

src/pages/DaoDashboard/Info/InfoProposals.tsx
```diff
@@ -30,7 +32,11 @@ export function InfoProposals({}: IDAOGovernance) {
   }
 
   return (
-    <Box data-testid="dashboard-daoProposals">
+    <Link
+      data-testid="dashboard-daoProposals"
+      to={DAO_ROUTES.proposals.relative(safe.address)}
+      aria-label="Proposals Info Link"
```
herbig, line 38: proposals / treasury cards are now clickable to go to their respective pages, which is something I'm constantly expecting you can do

src/components/DaoCreator/gnosis/GnosisConfig.tsx
```diff
@@ -33,7 +33,12 @@ export function GnosisConfig() {
 
   const handleSignersChanges = (numberStr: string) => {
     let numOfSigners = Number(numberStr);
-    if (numOfSigners > 999) return;
+    // greater than 100 signers is unreasonable for manual input here,
+    // we don't use an error message because we don't want to render
+    // 1000 input fields and lag the app
+    if (numOfSigners > 100) {
+      numOfSigners = 100;
+    }
```
herbig, line 41: it's still super janky trying to render 999 inputs, so just cut that down to 100

src/pages/DaoDashboard/Info/InfoProposals.tsx
mobiusTripper-crypto: right on

src/pages/DaoDashboard/Info/InfoProposals.tsx
```diff
@@ -1,8 +1,10 @@
-import { Box, Flex, Text } from '@chakra-ui/react';
+import { Flex, Text } from '@chakra-ui/react';
 import { Proposals } from '@decent-org/fractal-ui';
 import { useTranslation } from 'react-i18next';
+import { Link } from 'react-router-dom';
 import { BarLoader } from '../../../components/ui/loaders/BarLoader';
 import { useFractal } from '../../../providers/Fractal/hooks/useFractal';
+import { DAO_ROUTES } from '../../../routes/constants';
 
 interface IDAOGovernance {}
```
mobiusTripper-crypto, line 9: This seems like it unused???

src/providers/Fractal/governance/types.ts
```diff
@@ -19,8 +19,8 @@ import { DecodedTransaction, MetaTransaction } from '../../../types';
 import { IGoveranceTokenData } from './hooks/useGovernanceTokenData';
 
 export enum GovernanceTypes {
-  GNOSIS_SAFE = 'Multisig',
-  GNOSIS_SAFE_USUL = 'Usul',
+  GNOSIS_SAFE = 'labelMultisigGov',
```
mobiusTripper-crypto, line 22: I don't know if this is a good idea, if there would be consiquences from changing language? (I guess I could test that myself) ... I still think this should be a MAP, in which the enum is just an integer, and the label is assign in a map ... but if this works i guess it doesn't matter.

mobiusTripper-crypto (approved):
couple comments

src/pages/DaoDashboard/Info/InfoProposals.tsx
herbig: it's used below in the `<Link>`

src/pages/DaoDashboard/Info/InfoProposals.tsx
herbig: hmm, yeah true, let me see what that takes..

src/providers/Fractal/governance/types.ts
herbig: ah true. I tested changing languages and it properly updates.

what would that look like as a map?  We've used this style of enum value with string key in a few other instances in the app, so I did that here as well

src/pages/DaoDashboard/Info/InfoProposals.tsx
herbig: @mobiusTripper-crypto updated!

src/components/ui/containers/InfoBox.tsx
```diff
@@ -26,4 +29,5 @@ export function InfoBox({
       {children}
     </Box>
   );
+  return <>{!to ? content : <Link to={to}>{content}</Link>}</>;
```
Da-Colon, line 32: (nitpick, feel free to ignore since it works) This feels wierd, feel like this should either be 2 Box components are a `ContentWrapper`

Da-Colon (approved):
LGTM, small minor comment that can be ignored

src/components/ui/containers/InfoBox.tsx
herbig: yeah good callout, actually that div wasn't needed, I simplified it to:

`return !to ? content : <Link to={to}>{content}</Link>;`

PR #858, opened 2022-12-23
Info bug fix and various tweaks

herbig:
## Description

- fix BoxInfo width issue on DAO homepage from adding links
- circle crop ENS avatar image
- remove Favorites menu title (redundant)
- expand favorites menu max height to avoid scrollbars
- make favorites scrollbars on the outside edge of padding rather than inside it
- align top of favorites / account menu dropdowns with each other
- update address helper text to match design
- link ENS/wallet avatar to account on Etherscan
- link name text in DAO Hierarchy to the DAO
- top align transaction input when collapsed
- toggle Show / Hide Executable Code

Screenshot 2022-12-23 at 5 25 56 PM
Screenshot 2022-12-23 at 5 25 13 PM


src/constants/common.ts
```diff
@@ -1,3 +1,4 @@
 export const HEADER_HEIGHT = '4rem';
 export const CONTENT_HEIGHT = `calc(100vh - ${HEADER_HEIGHT})`;
 export const BACKGROUND_SEMI_TRANSPARENT = 'black.900-semi-transparent';
+export const ETH_ADDRESS_PLACEHOLDER = '0x1234';
```
Da-Colon, line 4: versus zeroAddress? just curious why this change?

src/components/ui/Header/Avatar.tsx
```diff
@@ -28,7 +28,7 @@ function URLAvatar({ url, size }: { size: AvatarSize; url: string }) {
   return (
     <Box w={avatarSizes[size]}>
       <Image
-        className="rounded-full"
+        borderRadius="full"
```
Da-Colon, line 31: good catch. probably should do a codebase search for `className`. I don't believe there should be any left in the codebase

src/components/ui/menus/AccountDisplay/MenuButtonDisplay.tsx
```diff
@@ -1,10 +1,10 @@
 import { Box, Flex, Text } from '@chakra-ui/react';
-import { ArrowDown } from '@decent-org/fractal-ui';
```
Da-Colon, line 2: Why do we need to switch to a SVG here versus the Design system icon?

Da-Colon (commented):
looks good, got a few questions

src/constants/common.ts
herbig: @Da-Colon the more I saw the zero address the more it felt pretty weird to me and I started to like what the design had.  I dunno, at the very least consolidating them into this constant will allow us to easily change it back or to whatever we want later on.

src/components/ui/Header/Avatar.tsx
herbig: yeah I searched for the term and those were the only instances

src/components/ui/menus/AccountDisplay/MenuButtonDisplay.tsx
herbig: ah, yeah we have two different arrows in the menu, the Favorites uses the SVG one, so I used the same here so they'd be the same size.

Though yeah I should use the fractal-UI one for both instead.  I'll update that.

herbig: @Da-Colon updated!

src/constants/common.ts
Da-Colon: heard. that makes sense to me

sethhrbek (approved):

herbig:
@Da-Colon I think this PR is good to go now, if I could get an approval!

PR #859, opened 2022-12-23
rename "Executing" step to "Executable"

herbig:
## Description

I missed this in previous code reviews, but I think "Executable" is a better term for this step.  Executing sounds like an active process, whereas this is something to act on.

Like `Queueable` / `Queued` we should call this `Executable` / `Executed`

## Notes

This should have no impact other than changing the badge to `Executable`.

Screenshot 2022-12-23 at 5 52 30 PM

### DIFF
--- a/src/components/Proposals/MultisigProposalDetails/TxActions.tsx
+++ b/src/components/Proposals/MultisigProposalDetails/TxActions.tsx
@@ -177,7 +177,7 @@ export function TxActions({
       pageTitle: 'signTitle',
       icon: undefined,
     },
-    [TxProposalState.Executable]: {
+    [TxProposalState.Executing]: {
       action: executeTransaction,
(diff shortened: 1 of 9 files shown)


sethhrbek:
@herbig -- the reason this was left as "Executing" is that this matches the name of this phase in the Usul `ProposalState` enum on their contracts.

Doesn't mean we have to keep it, just noting why it's named what it is.

herbig:
> @herbig -- the reason this was left as "Executing" is that this matches the name of this phase in the Usul `ProposalState` enum on their contracts.
> 
> Doesn't mean we have to keep it, just noting why it's named what it is.

ah right.. blugh.  Executing is really misleading here

Da-Colon (approved):
💯

sethhrbek (approved):

PR #860, opened 2022-12-31
Proposal Filters cleanup

herbig:
## Description

This makes filter options applicable to the proper DAO context, so you'll only see states for multisig proposals on a multisig DAO, for example.

## Testing

Please test each of the four cases (`FILTERS_USUL_BASE`, `FILTERS_USUL_CHILD`, `FILTERS_MULTISIG_BASE`, `FILTERS_MULTISIG_CHILD` to ensure that each case only shows the proposal states applicable to that type of DAO.

**previous**:

Screenshot 2022-12-23 at 4 08 25 PM

**now**:
Screenshot 2022-12-31 at 2 43 55 PM


src/components/Proposals/index.tsx
```diff
@@ -1,22 +1,91 @@
...
 import useProposals from '../../hooks/DAO/proposal/useProposals';
-import { TxProposalState } from '../../providers/Fractal/types';
+import { useFractal } from '../../providers/Fractal/hooks/useFractal';
+import { GovernanceTypes, TxProposalState } from '../../providers/Fractal/types';
 import { SortBy } from '../../types';
 import { Sort } from '../ui/Sort';
 import { OptionMenu } from '../ui/menus/OptionMenu';
 import { ProposalsList } from './ProposalsList';
 
+const FILTERS_USUL_BASE = [
```
herbig, line 13: @tbwebb22 @sethhrbek could you look over these states and let me know if these make sense, or are there some I should add in any of the cases?

Da-Colon (commented):
Changes look good to me pending @tbwebb22 and/or @sethhrbek reviews on the states

src/components/Proposals/index.tsx
```diff
@@ -1,22 +1,91 @@
...
-  const allStates = Object.values(TxProposalState);
+  const {
+    governance: { type, governanceIsLoading },
+    gnosis: { guardContracts, isGnosisLoading },
+  } = useFractal();
+
   const [sortBy, setSortBy] = useState<SortBy>(SortBy.Newest);
-  const [filters, setFilters] = useState<TxProposalState[]>(allStates);
+  const [filters, setFilters] = useState<TxProposalState[]>([]);
+  const [allOptions, setAllOptions] = useState<TxProposalState[]>([]);
```
sethhrbek, line 59: What do each of these represent? Are `filters` the set of all filters, and `allOptions` the ones that show based on the DAO type? If so, maybe rename `filters` to `allFilters` and then `options` to `filters` to indicate one is a subset of the other.

src/components/Proposals/index.tsx
sethhrbek: Will need Taylor's input here --

I thought all DAO types except multisig base would need to have proposals queued, so would need a queueable state. I also am unsure how Timelocked is different from Queued, I thought those represented the same state.

src/components/Proposals/index.tsx
herbig: yeah, that's right.  Good call, I'll update to be more clear.

herbig: @sethhrbek `filters` is the currently selected filters, from the set of all options here.

src/providers/Fractal/governance/types.ts
```diff
@@ -116,25 +116,135 @@ export interface GovernanceContracts {
...
-  Uninitialized = 'stateUninitialized',
+
+  /**
+   *
+   */
   Failed = 'stateFailed',
-  Queueable = 'stateQueueable',
-  // Safe-specific states
-  Queued = 'stateQueued', // Timelock for multisig starts when queued
-  Approved = 'ownerApproved',
```
herbig, line 134: this isn't a proposal state, not sure what it was doing here, ha

Da-Colon (approved):
Looks/works great.

tbwebb22 (approved):
Cool, these states look good to me

PR #868, opened 2023-01-02
Update | Use `@fractal-framework/fractal-contracts` deployments

Da-Colon:
## Description
Quick PR to update the Hardcoded Address

## Notes
We may revist when we figure out versioning for our app. Current each json file much be imported seperated. We do have an option on `hardhat-deploy` for a memoized deployment file. but may just cause more tech debt in the future for versioning. 

#867


herbig (approved):
nice

Issue #863, opened 2023-01-01
Add Eth value in Proposal Creation


herbig:
### Overview

Add Eth value parameter in Proposal Creation

### Link to documentation

https://www.notion.so/hulking-cosmos-3e8/Add-Eth-value-in-Proposal-Creation-54b92aa702174282902bceb4fd08f7ae

PR #875, opened 2023-01-03
Proposal eth value

tbwebb22:
## Description
This PR adds an ETH value input field to the create proposal flow so that ETH can be sent with DAO transactions

https://github.com/decent-dao/fractal-interface/issues/863

## Testing
1) Browse to a DAO you can create proposals on
2) Create a TX / proposal with a non-zero ETH value
3) Verify on the Gnosis UI that the TX will send the specified amount of ETH

image


src/i18n/locales/en/proposal.json
```diff
@@ -8,6 +8,8 @@
     "helperFunctionSignature": "The function of the smart contract (above) to be called if this proposal passes",
     "labelParameters": "Parameters",
     "helperParameters": "Values used to call the function (comma separated)",
+    "labelEthValue": "ETH Value",
+    "helperEthValue": "Amount of Ether to send with the transaction",
```
herbig, line 12: I had a note in the Notion doc, but we should also change some of the existing text to remove the term "value", to avoid confusion with the transaction value.

So

`"Values used to call the function (comma separated)"`

Should be changed to:

`”Parameters used to call the function (comma separated)”`

And 

`parameter values` to `parameter inputs`

herbig: And just remove the Ukrainian strings for those two, and they can be retranslated during a future pass

src/components/ProposalCreate/Transaction.tsx
```diff
@@ -171,6 +180,22 @@ function Transaction({
         }
         errorMessage={transaction.fragmentError}
       />
+      <BigNumberComponent
```
herbig, line 183: I'm able to create a proposal with just a decimal point ( . ), or like `90.`.  Does that do anything weird as input?  Will that just be 0 / 90?

src/i18n/locales/en/proposal.json
tbwebb22: Cool, this has been addressed

src/components/ProposalCreate/Transaction.tsx
tbwebb22: hmm yeah just the decimal point resolves to 0. And "90." resolves to 90

This is reusing the BigNumber input component that Eric created with those PRs back in December

Da-Colon (approved):
Pending a question about using ETHER values versus using WEI on the input value. but otherwise I everything works as intended

herbig:
> Pending a question about using ETHER values versus using WEI on the input value. but otherwise I everything works as intended

I think using Ether makes sense.  Wei is more likely to cause confusion, and the helper text / example gives enough context that you're entering in Ether units I think.

Etherscan's contract input fields use Ether as well.

src/components/ProposalCreate/Transaction.tsx
```diff
@@ -171,6 +180,22 @@ function Transaction({
...
+        subLabel={
+          <HStack>
+            <Text>{`${t('example', { ns: 'common' })}:`}</Text>
+            <Text {...exampleLabelStyle}>{'1.2'}</Text>
+          </HStack>
+        }
+        errorMessage={transaction.fragmentError}
+        value={transaction.ethValue.bigNumberValue}
+        onChange={updateEthValue}
+        decimalPlaces={18}
```
herbig, line 197: minor thing but the component defaults to 18 places

herbig (approved):

PR #888, opened 2023-01-05
Update | Freeze Child DAO option update

### LINKED ISSUE #871: DAO Heirarchy - Freeze Issues
Steps to recreate:

Setup
1. Create a multisig RootDAO, adding yourself as a signer.
2. Create a multisig SubDAO of this RootDAO. Add yourself as a signer to this DAO as well. Use default params for all parent controls.

Parent DAO Freeze:
1. Visit ParentDAO Heirarchy page.
2. See that the Child DAO cannot be frozen (not an option in the menu).

Image

SubDAO Freeze
1. Visit SubDAO Heirarchy page.
2. See that it allows you to freeze both the Parent *and* Child DAOs. The parent should not be freezable.

Image

Da-Colon:
## Description
Small Refactors to the Governance Data-layer to optimize the calls and the re-rendering. (there is more work to be done here).

Main goal of this PR is to improve the functionality of the freeze option. It was not showing up for a user in the case that a vote has happen and at the time of that vote they have voting powers.

The logic here has been update to only do a historical check if a freeze Period or freeze proposal period are currently active.

Also noticed I can make the numberOfChildren data live on DAODashboard

Closes #871

image

image

image


herbig:
> Generally - looks good. I see the same potential problem as with hierarchy PR I just made - it's not really scalable if we're talking about large DAO structure.
> 
> But I also believe there's no use case for that huge DAO structure in the industry atm - so we probably shouldn't caring too much until we will get some feedback from beta 🤔

yeah, agreed.  We shouldn't stress out over it, if that becomes a problem it means we're doing quite well...

herbig (approved):
Looks good, and works as expected!

PR #894, opened 2023-01-08
Bugfix | Proposal Execution Refresh

### LINKED ISSUE #884: SubDAO doesn't populate after deploying (without refresh)
After creating a SubDAO, visit the heirarchy page and the SubDAO doesn't show.

After refreshing the page, the SubDAO does show.

This should be updated so that the SubDAO would show immediately after visiting the heirarchy page.

In addition:

When creating a SubDAO proposal, the toasts should be updated to "Creating SubDAO Proposal" and "Proposal Created" (or something similar) -- right now it says "SubDAO deployed", which isn't accurate.

Da-Colon:
## Description
Going off the pattern I started to setup in #888, I've added a `refresh` method to the actions object in Fractal Provider. To start I've only included Transactions and SafeInfo as these retreive the info we need right now (to save on requests). When call these function don't return anything but update state. 

One other note. Since we don't really know what information we are trying to refresh when a proposal is executed. This isn't always going to catch the new DAO on the initial refresh. This PR also takes care of adding the SafeInfo into the interval as well. So child DAO will appear after some time.

## Notes
I've also updated the copy based on the request in the card.

Closes #884

### DIFF
--- a/src/hooks/DAO/useCreateSubDAOProposal.ts
+++ b/src/hooks/DAO/useCreateSubDAOProposal.ts
@@ -10,7 +10,7 @@ import useBuildDAOTx from './useBuildDAOTx';
 
 export const useCreateSubDAOProposal = () => {
   const { multiSendContract } = useSafeContracts();
-  const { t } = useTranslation(['daoCreate', 'proposal']);
+  const { t } = useTranslation(['daoCreate', 'proposal', 'proposalMetadata']);
 
   const { submitProposal, pendingCreateTx, canUserCreateProposal } = useSubmitProposal();
   const [build] = useBuildDAOTx();
@@ -37,7 +37,7 @@ export const useCreateSubDAOProposal = () => {
           targets: [multiSendContract.address],
           values: [BigNumber.from('0')],
           calldatas: [multiSendContract.interface.encodeFunctionData('multiSend', [safeTx])],
-          title: t('createSubDAO'),
+          title: t('Create a proposal', { ns: 'proposalMetadata' }),
           description: 'to:' + multiSendContract.address + '_ txs:' + safeTx,
           documentationUrl: '',
         };
--- a/src/hooks/DAO/useCreateSubDAOProposal.ts
+++ b/src/hooks/DAO/useCreateSubDAOProposal.ts
@@ -10,7 +10,7 @@ import useBuildDAOTx from './useBuildDAOTx';
 
 export const useCreateSubDAOProposal = () => {
   const { multiSendContract } = useSafeContracts();
-  const { t } = useTranslation('daoCreate');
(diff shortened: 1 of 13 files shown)


src/hooks/DAO/useCreateSubDAOProposal.ts
```diff
@@ -37,15 +37,15 @@ export const useCreateSubDAOProposal = () => {
           targets: [multiSendContract.address],
           values: [BigNumber.from('0')],
           calldatas: [multiSendContract.interface.encodeFunctionData('multiSend', [safeTx])],
-          title: 'Create SubDAO',
+          title: t('createSubDAO'),
```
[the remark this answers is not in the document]
herbig: From discussion in Discord, let's do what @mudrila is suggesting, using English literals as the key value, for proposal metadata, but keep all such strings to a separate metadata specific json file.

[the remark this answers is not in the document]
herbig: Also "Create a subDAO" should be uncapitalized, but that's not a change you made!

src/i18n/locales/en/daoCreate.json
```diff
@@ -73,7 +73,8 @@
     "titleTokenParams": "Token Metadata",
     "titleAddress": "Address",
     "titleAmount": "Amount",
-    "createSubDAOPendingToastMessage": "Deploying your subDAO",
+    "Create SubDAO": "Create SubDAO",
+    "createSubDAOPendingToastMessage": "Creating your SubDAO Proposal",
```
herbig, line 77: `subDAO` should be uncapitalized here

herbig (changes requested):
Looks good, a few instances of subDAO capitalization to fix, and the discussion about moving metadata strings to json with English literals as keys

src/hooks/DAO/useCreateSubDAOProposal.ts
```diff
@@ -37,7 +37,7 @@ export const useCreateSubDAOProposal = () => {
           targets: [multiSendContract.address],
           values: [BigNumber.from('0')],
           calldatas: [multiSendContract.interface.encodeFunctionData('multiSend', [safeTx])],
-          title: t('createSubDAO'),
+          title: t('Create a proposal', { ns: 'proposalMetadata' }),
```
herbig: would "Create a subDAO" be a better title here though?

src/i18n/locales/en/proposalMetadata.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Create a SubDAO proposal": "Create a SubDAO proposal"
```
herbig, line 2: subDAO should be uncapitalized here as well!

herbig (approved):

PR #908, opened 2023-01-11
Delete | RIP Web3Modal provider

Da-Colon:
## Description
Removes old web3 provider from codebase. I've moved remaining items being used to new locations and updated imports


src/providers/Fractal/utils/api.ts
```diff
@@ -1,4 +1,29 @@
...
+  name: string;
+  id: number;
+};
+
+/* We might use something https://github.com/ethereum-lists/chains
+ * if number of supported chains would grew up, this should remain sync with SUPPORTED_CHAIN_IDS of both dev site and prod
+ */
+const CHAINS: ChainMetadata[] = [
+  {
+    name: 'Mainnet',
```
herbig, line 11: not sure if there's any point to translating these, but we can discuss later on

herbig (approved):

PR #913, opened 2023-01-11
Decent feedback quick fixes

herbig:
## Description

- [fix eth value not disabling](https://github.com/decent-dao/fractal-interface/commit/843338cf108c3d51a85beb77dc2c70edcf7f585f)
- [fix proposal button logic](https://github.com/decent-dao/fractal-interface/commit/5394bd36ef3bfb58918f307f793f5e01b244315b)
- [fix mispelling](https://github.com/decent-dao/fractal-interface/commit/53658e71d113701bf63e2d82aed4db8939950b54)
- [update connection text](https://github.com/decent-dao/fractal-interface/commit/f0a8af82d5ec1bcfc9e2215bab3b51b5a54fa144)
- [update search text from feedback](https://github.com/decent-dao/fractal-interface/commit/36a55a3e9c31dba07c1469d230d4529b1ce6f39f)


src/components/Proposals/ProposalsList.tsx
```diff
@@ -20,7 +20,7 @@ export function ProposalsList({ proposals }: { proposals: TxProposal[] }) {
   } = useFractal();
 
   const showCreateButton =
-    type === GovernanceTypes.GNOSIS_SAFE_USUL ?? owners?.includes(account || '');
+    type === GovernanceTypes.GNOSIS_SAFE_USUL ? true : owners?.includes(account || '');
```
herbig, line 23: @mudrila had to switch this back to what I had before, since it removed the button on dev currently

I'm not sure why this logic is different though...

src/i18n/locales/en/common.json
```diff
@@ -55,5 +55,5 @@
     "timeAgo": "{{time}} ago",
     "cancel": "Cancel",
     "example": "Example",
-    "titleNodes": "DAO Hierachy"
+    "titleNodes": "DAO Hierarchy"
```
herbig, line 58: 🤦

src/i18n/locales/en/dashboard.json
```diff
@@ -1,9 +1,9 @@
 {
     "titleFavorites": "Favorites",
     "emptyFavorites": "No favorites",
-    "errorInvalidSearch": "Please use a valid Fractal or Gnosis Safe address, or ENS domain",
+    "errorInvalidSearch": "Please use a valid Fractal address",
```
herbig, line 4: feedback from Adam was that this confusing, since a valid personal ENS domain like michaelherbig.eth should under this text show something, but it does not, which makes sense

I just simplified it to a Fractal address for now.  I realize we can show any Safe, but we haven't yet fixed any issues there, so we'll highlight this as searching for a Fractal only, and let the user discover that a Fractal with an ENS is valid here

Issue #906, opened 2023-01-10
Bug: "ago" appears at the end of "just now"


herbig:
### Overview

I think this issue could warrant a refactor of how we're composing these strings.

We should have different strings for the various uses of these, and not concat two different strings to get what we want.

### Steps to reproduce

create a new transaction into a DAO treasury or anywhere else with a time stamp, and view within 5 minutes

### Screenshots / Video

<img width="193" alt="Screenshot 2023-01-09 at 11 47 17 AM" src="https://user-images.githubusercontent.com/384559/211684105-06eee9e3-1a49-46eb-9692-2743d4c9e37e.png">

### Version

Prod (app.fractalframework.xyz/)

### Browser

Chrome (or Chromium based)

### Operating System

None

### web3 wallet

None

PR #914, opened 2023-01-11
Datetime refactor

herbig:
## Description

Refactors the `useDateTimeDiff` method into a hook with a single parameter, the reference date.

closes https://github.com/decent-dao/fractal-interface/issues/906

Screenshot 2023-01-11 at 4 19 33 PM


src/pages/Treasury/hooks/useFormatTransfers.tsx
```diff
@@ -51,9 +47,7 @@ export function useFormatTransfers(
     const formatted: TransferDisplayData = {
       eventType: safeAddress === transfer.from ? TokenEventType.WITHDRAW : TokenEventType.DEPOSIT,
       transferType: transfer.type as TransferType,
-      dateTimeDisplay: t('timeAgo', {
-        time: formatDatesDiffReadable(new Date(transfer.executionDate), now, t),
-      }),
+      executionDate: transfer.executionDate,
```
herbig, line 50: moved the use of the method to where this data is used, rather than here, because this is called in a loop, which you shouldn't do with hooks

src/i18n/locales/en/common.json
```diff
@@ -35,24 +35,32 @@
...
+    "labelMinutesAgo": "{{count}} minutes ago",
+    "labelHoursAgo_one": "{{count}} hour ago",
+    "labelHoursAgo_other": "{{count}} hours ago",
+    "labelDaysAgo_one": "{{count}} day ago",
+    "labelDaysAgo_other": "{{count}} days ago",
+    "labelMonthsAgo_one": "{{count}} month ago",
+    "labelMonthsAgo_other": "{{count}} months ago",
+    "labelYearsAgo_one": "{{count}} year ago",
+    "labelYearsAgo_other": "{{count}} years ago",
+    "labelNowishLeft": "a few minutes left",
```
herbig, line 48: we could get more granular than < 5 minutes, but unless we're going to have a live countdown, I think keeping it nebulous is better than thinking they have more time than they do

tbwebb22:
Should all timestamps - like this one below - show "time ago" instead of the actual time if it was recent enough? The one below was less than 5 minutes ago, but still shows the timestamp

image

herbig:
> Should all timestamps - like this one below - show "time ago" instead of the actual time if it was recent enough? The one below was less than 5 minutes ago, but still shows the timestamp
> 
> image

hmm, yeah this should have no impact on that, since it wasn't using this method, but we could show the "time ago" there as well, and go to a date if it's past a certain time

tbwebb22:
> 

Ok, yeah if we didn't intend for this to affect those timestamps, then this LGTM

tbwebb22 (approved):

Issue #862, opened 2023-01-01
Token Voting Proposal Metadata


herbig:
### Overview

Add title, description, url fields to Token Voting proposal creation / details.

### Link to documentation

https://www.notion.so/hulking-cosmos-3e8/Token-Voting-Proposal-Metadata-dcc90df28b3f4c04b1b8e3f90c914590

https://www.figma.com/file/cO1nBoNZCmMKCwCCpyTxfx/%F0%9F%A6%84-Fractal-%7C-Pages-%26-Flows-Master?node-id=4204%3A28213&t=oQZiYzaLUJiLW45k-0

PR #915, opened 2023-01-11
Usul metadata

tbwebb22:
## Description
This PR: 
1) Updates the Usul proposal creation process to include a metadata step with fields for "title", "description", and "additional resources" (an external link)
2) Replaces the generated proposal titles from decoded data with the metadata title, if one was provided

https://github.com/decent-dao/fractal-interface/issues/862

## Testing
1) Create various proposals on a Usul DAO, some providing titles, descriptions, and links, and others not.
2) Verify that the proposals are displayed correctly in the various views, and show the metadata values when appropriate


herbig (changes requested):
Hey @tbwebb22 some UI/UX feedback:

**Metadata screen**

- [x] remove the black background and put the fields into the outer card instead, as in the designs
- [x] remove periods from the end of the hint text (I realize the design has this, but we've removed that throughout the app after feedback from Natalie on copy)
- [x] add placeholder text to the inputs from the designs
- [x] add the horizontal bar before the Next button
- [ ] (I realize this wasn't part of this PR) but could you change the copy for "Timelock" to "Queue period", we've decided to call it all Queued rather than Timelocked on the frontend
- [ ] can the Description field wrap to a new line rather than continue to the right?
- [ ] can the Description field have a default height of three lines?
- [ ] can we do URL validation on the documentation link here?

**Proposal transactions screen**

Metadata title should be added here, if it exists (see screenshot)
Screenshot 2023-01-12 at 4 15 16 PM

src/components/Activity/ActivityDescriptionGovernance.tsx
```diff
@@ -76,6 +77,12 @@ export function ActivityDescriptionGovernance({ activity }: IActivityDescription
     count: governanceActivity.targets.length,
   });
 
+  const usulProposalMetaDataTitle = (activity as UsulProposal).metaData?.title;
+
+  if (!!usulProposalMetaDataTitle) {
+    return <Text>{usulProposalMetaDataTitle}</Text>;
+  }
```
sethhrbek, line 84: @herbig @tbwebb22  -- do we want to still include the proposal number even if we have a metadata title?

There is also an `OnChainRejectionMessage` in the current title -- not sure if this should be included as well (not sure what this represents)

src/pages/ProposalCreate/index.tsx
```diff
@@ -55,6 +48,26 @@ function ProposalCreate() {
...
+    documentationUrl: string;
+  }>({ title: '', description: '', documentationUrl: '' });
+
+  useEffect(() => {
+    if (!usulContract) {
+      setIsUsul(false);
+    } else {
+      setIsUsul(true);
+    }
+  }, [usulContract]);
```
sethhrbek, line 65: This could be simplified to:

```
  useEffect(() => {
    setIsUsul(!!usulContract);
  }, [usulContract]);
```

src/pages/ProposalCreate/index.tsx
```diff
@@ -55,6 +48,26 @@ function ProposalCreate() {
...
+
+  useEffect(() => {
+    if (isUsul === undefined) return;
+    setShowTransactionsAndSubmit(!isUsul || inputtedMetadata ? true : false);
```
sethhrbek, line 69: this could be simplified to:
```
setShowTransactionsAndSubmit(!isUsul || inputtedMetadata);
```

src/pages/ProposalCreate/index.tsx
```diff
@@ -94,12 +107,18 @@ function ProposalCreate() {
...
     };
     setProposalData(proposal);
-  }, [transactions, proposalDescription]);
+  }, [
+    transactions,
+    proposalDescription,
+    metadata.title,
+    metadata.description,
+    metadata.documentationUrl,
+  ]);
```
sethhrbek, line 121: This is outside the scope of this PR, but all of these dependencies lead to the `transactions` being validated, even if the user is only typing into the metadata fields (b/c updating `proposalData` triggers `isValidProposal` to run.

Probably makes sense to optimize this so only the transactions trigger the `isValidProposal` validation (which may not be possible without refactoring, so this can wait). The most optimized version would have metadata && transaction data set separately on the proposal in different `useEffects` depending on which field type is being updated.

src/pages/ProposalCreate/index.tsx
```diff
@@ -157,6 +176,17 @@ function ProposalCreate() {
...
         <GridItem area="header">
+          {inputtedMetadata && (
+            <Text
+              textStyle="text-md-mono-regular"
+              color="gold.500"
+              cursor="pointer"
+              onClick={() => setInputtedMetadata(false)}
+              mb={4}
+            >
+              {'< Back'}
```
sethhrbek, line 187: does this need a translation?

src/pages/ProposalCreate/index.tsx
tbwebb22: Cool, simplified to this

src/pages/ProposalCreate/index.tsx
tbwebb22: Cool, simplified to this

src/components/Activity/ActivityDescriptionGovernance.tsx
tbwebb22: Yeah, agreed having the proposal number there makes sense. Just added it back in for metadata titles

tbwebb22:
> 

Working through these now. Can we handle the changes from "timelock" to "queue" in a separate PR since that is unrelated to this?

tbwebb22:
> > 
> 
> @herbig Working through these now. Can we handle the changes from "timelock" to "queue" in a separate PR since that is unrelated to this?

src/pages/ProposalCreate/index.tsx
tbwebb22: Good call, added translation for this

src/components/ProposalCreate/TransactionsAndSubmit.tsx
```diff
@@ -0,0 +1,107 @@
...
+    successToastMessage: string;
+    successCallback?: ((daoAddress: string) => void) | undefined;
+  }) => Promise<void>;
+  proposalData: ProposalExecuteData | undefined;
+  successCallback: () => void;
+  isCreateDisabled: boolean;
+  transactions: TransactionData[];
+  setTransactions: Dispatch<SetStateAction<TransactionData[]>>;
+  removeTransaction: (transactionNumber: number) => void;
+}) {
```
Da-Colon, line 43: Might be good to extract to an interface above for readability

src/components/ProposalCreate/UsulMetadata.tsx
```diff
@@ -0,0 +1,125 @@
...
+    documentationUrl: string;
+  };
+  setMetadata: Dispatch<
+    SetStateAction<{
+      title: string;
+      description: string;
+      documentationUrl: string;
+    }>
+  >;
+}) {
```
Da-Colon, line 26: same comment as above

Da-Colon (commented):
Will need to get that new SVG in the design system but we can worry about that later.

one other comment but LGTM

herbig (changes requested):
@tbwebb22 two things:

- The description placeholder is larger than the others, maybe that has to do with the larger height? but is there a way to make the placeholder text size the same?
- Metadata title should be on the second screeen, if it exists:
Screenshot 2023-01-12 at 4 15 16 PM

src/components/ProposalCreate/TransactionsAndSubmit.tsx
tbwebb22: Good call, pushed this up

src/components/ProposalCreate/UsulMetadata.tsx
tbwebb22: Good call, pushed this up

herbig (approved):
🚀🚀🚀

Da-Colon (approved):
LGTM

Issue #910, opened 2023-01-11
Bug: new proposal requires refresh


herbig:
### Overview

> After creating a new proposal, it doesn’t show up on the Proposals page unless a page refresh happens.

### Steps to reproduce

n/a

### Version

Dev (app.dev.fractalframework.xyz/)

### Browser

Chrome (or Chromium based)

### Operating System

None

### web3 wallet

None

tbwebb22:
I'll be tackling this one. Looks like this is only an issue for Multisig DAOs. On a Usul DAO, proposals page seems to get updated automatically with the new proposal

PR #922, opened 2023-01-12
Create proposal refresh

tbwebb22:
## Description
This fixes the bug with a newly created multisig proposal not being automatically added to the proposal list. Previously, a refresh was required to get the new proposal.

## Notes
The fix implemented is to add a 1 second timeout after the proposal is created before requesting the updated proposal list from the Gnosis API.

https://github.com/decent-dao/fractal-interface/issues/910

## Testing
1) In both a multisig DAO and a token voting DAO, create new proposals and verify that they are automatically added to the proposal list screen without needing to refresh the app.


src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -111,6 +111,7 @@ export default function useSubmitProposal() {
               }
             )
           );
+          await new Promise(resolve => setTimeout(resolve, 1000));
```
sethhrbek, line 114: One thing could be worth looking at is the `successCallback` functions --

`useCreateSubDAOProposal.ts` uses:
```
  const successCallback = async (daoAddress: string) => {
    await refreshSafeData();
    navigate(DAO_ROUTES.dao.relative(daoAddress));
  };
```

In this instance, it looks like the `await refreshSafeData();` would get called twice for multisig proposals -- once on line 115 here, and once when the `successCallback` runs.

Thinking we should homogenize how these are setup -- either all the success callback functions should be responsible for refreshing the safe data, or none should. If `successCallback` is responsible for refreshing the safe data, then it could also have a 1 second delay before navigating.

tbwebb22: Hmm yeah agreed, we should make these successCallbacks more consistent. Should we address that in this PR?

sethhrbek: We can kick that down the road for now, since this solution works.

sethhrbek (approved):

herbig (approved):

PR #941, opened 2023-01-25
BugFix: Search Bar Issues

### LINKED ISSUE #902: Bug: Search bar issues
### Overview

We should address the following Search bar issues:

- Allow selecting text within the input, e.g. click to a specific index in the string
- Address the empty results / error box (I can't replicate on Chrome or Firefox, this feedback was on Chrome)

### Steps to reproduce

> The search bar becomes hard to interact with at this point.
> 
> The cursor stays as a `pointer` when hovering over the text I typed in (i’d think it should be a `text` cursor), and it becomes impossible to select or click on. I can use the arrow keys on my keyboard to “move around” in the textbox, but not the mouse. seems weird. i wonder what this experience is like on mobile…
> 
> When I do delete all the text from the textbox, it results in an empty “results / error box” being displayed.

### Screenshots / Video

_No response_

### Version

### Browser

Chrome (or Chromium based)

### Operating System

None

### web3 wallet

None

Da-Colon:
## Description
Switched over the component to use `Popover` instead of the `Menu` Component

Closes #902

image

image

### DIFF
--- a/src/components/ui/menus/DAOSearch/SearchDisplay.tsx
+++ b/src/components/ui/menus/DAOSearch/SearchDisplay.tsx
@@ -1,4 +1,4 @@
-import { Box, Flex, MenuItem, Text, Button } from '@chakra-ui/react';
+import { Box, Flex, Text, Button } from '@chakra-ui/react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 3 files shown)


herbig (approved):
Nice, looks great

PR #947, opened 2023-01-27
Enchancement  | Add Polygon network

Da-Colon:


src/index.tsx
```diff
@@ -39,6 +39,7 @@ root.render(
             <FractalErrorBoundary fallback={<ErrorFallback />}>
               <WagmiConfig client={wagmiClient}>
                 <RainbowKitProvider
+                  initialChain={chains[0]}
```
Da-Colon, line 42: Intentional sets the fallback chain here. This might not be important now, but in the future we may need to be mindful of which chain is default. (say if we let the user 'choose' what network they are on via a dropdown

src/providers/NetworkConfig/networks/polygon.ts
```diff
@@ -0,0 +1,33 @@
...
+    fractalUsulMasterCopy: FractalUsul.address,
+    fractalModuleMasterCopy: FractalModule.address,
+    fractalRegistry: FractalRegistry.address,
+    votesTokenMasterCopy: VotesToken.address,
+    claimingMasterCopy: TokenClaim.address,
+    usulVetoGuardMasterCopy: UsulVetoGuard.address,
+    vetoMultisigVotingMasterCopy: VetoMultisigVoting.address,
+    vetoERC20VotingMasterCopy: VetoERC20Voting.address,
+    gnosisVetoGuardMasterCopy: VetoGuard.address,
+    gnosisSafe: '0x3E5c63644E683549055b9Be8653de26E0B4CD36E',
```
Da-Colon, line 27: For Polygon we are using https://github.com/safe-global/safe-deployments/blob/main/src/assets/v1.3.0/gnosis_safe_l2.json. Both contracts work but The regular one shows a warning in Gnosis. @tbwebb22 Do you see any reason this affects anything? I was able to create a DAO just fine with this L2 contract.

src/providers/NetworkConfig/rainbow-kit.config.ts
```diff
@@ -2,23 +2,24 @@ import '@rainbow-me/rainbowkit/styles.css';
...
 }
 
 export const { chains, provider } = configureChains(chainsArr, [
-  infuraProvider({ priority: 0, apiKey: process.env.REACT_APP_INFURA_API_KEY! }),
-  alchemyProvider({ priority: 1, apiKey: process.env.REACT_APP_ALCHEMY_API_KEY! }),
-  publicProvider({ priority: 2 }),
+  infuraProvider({ apiKey: process.env.REACT_APP_INFURA_API_KEY! }),
+  alchemyProvider({ apiKey: process.env.REACT_APP_ALCHEMY_API_KEY! }),
+  publicProvider(),
 ]);
```
Da-Colon, line 23: I removed the priority as they were unneeded, but another place to be mindful if we are hitting our limits, right now Alchemy is the only provider that we have setup that supports Polygon.

src/hooks/utils/useChainData.ts
```diff
@@ -31,6 +38,12 @@ const chainsMetaData: EVMChainMetaData = {
   },
 };
 
+export function useSupportedENS() {
+  const provider = useProvider();
+  const networkId = provider.network.chainId;
+  return [137].includes(networkId) ? chainsArr[0].id : networkId;
+}
+
```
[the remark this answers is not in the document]
herbig: what do you mean by not ENS ready?

as far as I'm aware, ENS is only supported on Eth mainnet, and testnets like Georli

Polygon has no concept of ENS.  Some web3 services will support showing your ENS even on Polygon, but I think that's kind of dangerous, since a smart contract with an ENS won't exist on other chains, unlike a personal wallet

herbig (approved):
looks goood

src/providers/NetworkConfig/networks/polygon.ts
tbwebb22: Can you post the warning you are seeing in Gnosis, or how to re-create it?

src/hooks/utils/useChainData.ts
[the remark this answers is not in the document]
Da-Colon: The latter is the assumption I was going with. good call. probably should just remove this and not support ENS on Polygon. I'll get this code updated.

src/providers/NetworkConfig/networks/polygon.ts
Da-Colon: ah yessir. in the network files you would replace the Gnosis Safe contract address with the gnosis_safe.json (same address as goerli). Then deploy a DAO. If you then visit Safe's app and go to your safe address it will say its base contract is wrong.

Da-Colon: > Just to verify - have you selected chain in the safe app itself? I mean not in the wallet, but in the app
yessir it does switch the contract addresses to polygon's network file. When the app changes networks it would update the selected config file.

src/index.tsx
Da-Colon: yeah removed this ha. was causing problems. really the fallback is decided by the `chainArr` in the `rainbow-kit.config.ts` file. whatever network is the first index is the fallback

src/components/ui/EtherscanLinkAddress.tsx
```diff
@@ -10,8 +10,8 @@ function EtherscanLinkAddress({
   address?: string;
   children: React.ReactNode;
 }) {
-  const subdomain = useSubDomain();
-  const href = address ? `https://${subdomain}etherscan.io/${path}/${address}` : undefined;
+  const { etherscanBaseURL } = useNetworkConfg();
```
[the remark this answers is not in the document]
herbig: I think there are EVM chains that don't have Etherscan based "scan" sites, so we may eventually need to refactor this to support those.

In that case we'll need to do more than just tack in the base URL though, since the sites operate differently

src/hooks/utils/useChainData.ts
```diff
@@ -35,3 +43,29 @@ export function useChainData(chainId: number) {
...
 }
+
+export function useNativeSymbol() {
+  const provider = useProvider();
+  const networkId = provider.network.chainId;
+  switch (networkId) {
+    case 137:
+      return 'MATIC';
+    case 1:
+    case 5:
```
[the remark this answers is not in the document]
herbig: yeah good call, @Da-Colon might as well change that, it's another useful check to know you're on a testnet rather than main, just in case

src/pages/Treasury/hooks/useFormatTransfers.tsx
```diff
@@ -27,6 +27,8 @@ export function useFormatTransfers(
   safeAddress: string
 ): TransferDisplayData[] {
   let displayData: TransferDisplayData[] = new Array(transfers.length);
+  const nativeSymbol = useNativeSymbol();
+  const naviveIcon = useNativeIcon();
```
herbig, line 31: another typo!

src/providers/NetworkConfig/networks/polygon.ts
tbwebb22: Oh ok I get it. Yeah should be fine since Polygon is just for testing purposes

tbwebb22 (approved):

PR #948, opened 2023-01-27
Refactor | SafeContracts

Da-Colon:
## Description
Testing Polygon network, showed that more thought will need to be put into how we are using `queryFilter` getting the events. With how fast polygon's block time is, the public RPC that most folks will is not able to query the Data as we were requesting it.

Temporary Solution: is to connect to the contract via our provider. Querying through our Alchemy fixes this issue for now.

When going through and cleaning up this implementation, I noticed alot of places where code was being duplicated to connect to the contract instances. There is a handy method `attach()` that accepts a contract address as argument and will do the same thing.


src/types/contract.ts
```diff
@@ -1,3 +1,40 @@
...
+  gnosisSafeSingletonContract: ContractConnection<GnosisSafe>;
+  zodiacModuleProxyFactoryContract: ContractConnection<ModuleProxyFactory>;
+  fractalModuleMasterCopyContract: ContractConnection<FractalModule>;
+  fractalRegistryContract: ContractConnection<FractalRegistry>;
+  gnosisVetoGuardMasterCopyContract: ContractConnection<VetoGuard>;
+  usulVetoGuardMasterCopyContract: ContractConnection<UsulVetoGuard>;
+  vetoMultisigVotingMasterCopyContract: ContractConnection<VetoMultisigVoting>;
+  vetoERC20VotingMasterCopyContract: ContractConnection<VetoERC20Voting>;
+  votesTokenMasterCopyContract: ContractConnection<VotesToken>;
+}
```
Da-Colon, line 40: Refactored the contract type. Anywhere were we are setting up a an instance of the contract that will all fit this pattern. 

`asSigner` = is basically connected as Signer or Provider when provider isn't connecting.

`asProvider` - will allow us control over which provider is used when a wallet is connected

src/providers/Fractal/utils/api.ts
```diff
@@ -15,6 +15,10 @@ const CHAINS: ChainMetadata[] = [
     name: 'Goerli',
     id: 5,
   },
+  {
+    name: 'Polygon',
+    id: 137,
+  },
```
Da-Colon, line 21: Probably should of made it in first PR, oops

src/providers/Fractal/hooks/useNodes.ts
```diff
@@ -33,9 +29,9 @@ export default function useNodes({
...
   const getDAOOwner = useCallback(
     async (safeInfo?: Partial<SafeInfoResponseWithGuard>) => {
-      if (safeInfo && safeInfo.guard && signerOrProvider) {
+      if (safeInfo && safeInfo.guard && gnosisVetoGuardMasterCopyContract) {
         if (safeInfo.guard !== ethers.constants.AddressZero) {
-          const guard = VetoGuard__factory.connect(safeInfo.guard, signerOrProvider);
+          const guard = gnosisVetoGuardMasterCopyContract.asSigner.attach(safeInfo.guard);
           const guardOwner = await guard.owner();
           if (guardOwner !== safeInfo.address) {
             return guardOwner;
```
Da-Colon, line 37: Good example of another refactor that been made. instead of reconnecting as `SignerOrProvider` here we only need to `attach` the address of the instance we want. reduces a little bit of the dependencies of the useEffects and removes some properties that are being passed around

src/pages/DaoCreate/index.tsx
```diff
@@ -19,6 +19,7 @@ function DaoCreate() {
 
   const successCallback = useCallback(
     async (daoAddress: string) => {
+      console.log('🚀 ~ file: index.tsx:22 ~ daoAddress', daoAddress);
```
herbig, line 22: probably don't need to keep this!

herbig (approved):
Looks good.

tbwebb22:
When initially connecting to the app, I get the metamask message to switch to Goerli like below. Is anyone else getting this behavior?
image

sethhrbek:
@Da-Colon -- I'm getting the same issue as @tbwebb22 when trying to connect to Polygon network on metamask.

sethhrbek:
My concern with this PR is it is adding some overhead / complexity to handle an edge case we are experiencing on polygon. 

I assumed we would target explicit event calls that could be costly and then use a custom RPC in only those cases (by doing some custom code **just** for those specific event lookups) , rather than needing to explicitly define every contract call as `asSigner` or `asProvider`. Some of the factory contracts that fractal deploys will probably not have that many events on them and probably could still use the signer. 

I say we still move forward using this pattern b/c it does give us flexibility going forward, just wanted to get these thoughts down.

Da-Colon:
@tbwebb22 @sethhrbek I'lll look into this today

PR #951, opened 2023-01-30
Support Polygon in UI

Da-Colon:
## Description
Updates formatters to support a second native coin (Polygon)


herbig (approved):
Makes sense, good catch

tbwebb22:
I'm getting the Metamask message to change networks from Polygon to Goerli on this PR as well when I initially connect my wallet.

herbig (changes requested):
@Da-Colon one issue: it still displays my Goerli ENS name, even though I'm on Polygon

Screenshot 2023-01-31 at 9 23 54 PM

PR #965, opened 2023-02-04
move everything network related into the network config files

herbig:
## Description

Moves all of the various network related configs into the network config files, to make it easier to add networks in the future.


.env
```diff
@@ -1,5 +1,4 @@
 # Wallet Provider
-REACT_APP_SUPPORTED_CHAIN_IDS="5, 137"
```
herbig, line 2: don't need this anymore, our supported chains is defined by what we define in the wagmi setup

sethhrbek: @herbig -- are we planning on having different environments (`app.dev.fra..`, `app.fractal`) support different chains (testnet, mainet)? Or will the app always support all chains supported by wagmi? If we need differing chain support by environment, then may need to keep these env vars.

herbig: I think ultimately what I'd like to have is 3 environments: `dev, staging, prod`, and the app should support all chains we support in every environment, including testnets

the only case prod wouldn't support a chain is when it is newly supported, and in the pipeline of `dev -> staging`

Da-Colon (approved):

tests/models/mock/data/creation.ts
```diff
@@ -1,11 +1,9 @@
-import { localhostConfig } from '../../../../src/providers/NetworkConfig/networks/localhost';
-
 export const SINGLE_OWNER_MULTISIG_CREATION = {
   created: '2022-12-13T20:32:48Z',
   creator: '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
   transactionHash: '0x932725fc8787f0694f4cda7cdf4913275d4a10792e7e93b24b84024df46e04a5',
-  factoryAddress: localhostConfig.contracts.gnosisSafeFactory,
-  masterCopy: localhostConfig.contracts.gnosisSafe,
+  factoryAddress: '0xa6B71E26C5e0845f74c812102Ca7114b6a896AB2',
+  masterCopy: '0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552',
```
sethhrbek, line 6: Should these be coming from the npm package (are these master copies there?), or maybe just setup in their own file in tests in case we need to reuse these? If we expand the tests to include subDAOs, we are going to need guard addresses as well.

This could get handled in a separate task, so this is non-blocking.

sethhrbek (approved):

tests/models/mock/data/creation.ts
herbig: @Da-Colon made this change, and the data mocks I haven't fully looked at

are these being hardcoded an issue?

PR #970, opened 2023-02-07
Lint precommit

sethhrbek:
Adds pre-commit script for running linter with husky.

Run `npm run prepare`, and then that should enable the precommit hook.


herbig:
I think @mudrila set something like this up before, but we ended up reverting it, but I don't remember why..  @mudrila do you remember?

herbig (approved):

herbig:
@sethhrbek I'm going to merge this in, just to move it along!

@decent-dao/fractal see the PR description for Husky usage!

Issue #989, opened 2023-02-20
ENS Request Caching


herbig:
### Overview

We need to cache whatever requests we can in preparation for Eth Denver (requests from our internal test day below).

Let's start with ENS requests.

![Screen Shot 2023-02-19 at 10 41 34 PM](https://user-images.githubusercontent.com/384559/220004510-c651b4ce-efc7-46c5-9862-7ca0af9e3789.png)

PR #990, opened 2023-02-20
Expirable caching

herbig:
## Description

- Adds a general method for getting / setting cached values, dependent on chainId
- Caches ENS resolving

closes https://github.com/decent-dao/fractal-interface/issues/989

## Testing
TBD..


src/hooks/utils/useAddress.ts
```diff
@@ -53,9 +72,13 @@ const useAddress = (addressInput: string | undefined) => {
       .resolveName(addressInput)
       .then(resolvedAddress => {
         if (!resolvedAddress) {
+          // cache an unresolved address as 'none' for 20 minutes
```
herbig, line 75: might be able to just use `undefined` here

src/providers/Fractal/hooks/account/useLocalStorage.ts
```diff
@@ -1,8 +1,13 @@
...
+import { useCallback, useEffect } from 'react';
+import { useNetworkConfg } from '../../../NetworkConfig/NetworkConfigProvider';
 
 export enum CacheKeys {
   FAVORITES = 'favorites',
   AUDIT = 'not_audited_acceptance',
+  // name.eth -> 0x0 caching
+  ENS_RESOLVE = 'ens_resolve_',
+  // 0x0 -> name.eth caching
+  ENS_LOOKUP = 'ens_lookup_',
```
herbig, line 10: this will be used for the reverse, which probably does way more requests

src/providers/Fractal/hooks/account/useLocalStorage.ts
```diff
@@ -34,3 +39,47 @@ export const useLocalStorage = () => {
...
+}
+
+export const useFractalStorage = () => {
+  const { chainId } = useNetworkConfg();
+
+  const setValue = useCallback(
+    (key: string, value: any, expirationMinutes?: number) => {
+      const val: FractalVal = {
+        value: value,
+        expiration: expirationMinutes ? Date.now() + expirationMinutes * 60000 : undefined,
```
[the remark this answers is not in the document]
herbig: this defaults to infinity if the user doesn't give an expiration, is that what you mean?

src/providers/Fractal/hooks/account/useLocalStorage.ts
```diff
@@ -34,3 +39,47 @@ export const useLocalStorage = () => {
     return;
   }, []);
 };
+
+interface FractalVal {
```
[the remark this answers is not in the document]
herbig: updated!

src/providers/Fractal/hooks/account/useLocalStorage.ts
```diff
@@ -34,3 +39,47 @@ export const useLocalStorage = () => {
...
+  value: any;
+  expiration?: number;
+}
+
+export const useFractalStorage = () => {
```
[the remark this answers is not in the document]
herbig: updated!

src/providers/Fractal/hooks/account/useLocalStorage.ts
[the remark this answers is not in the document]
herbig: ah, that doesn't work with things like Favorites or the audit prompt, which should be indefinite.

maybe we should have a default, like one week, but make indefinite caching explicitely stated?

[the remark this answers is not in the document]
herbig: updated! Thanks for the feedback, I think this is pretty simple and works great

src/providers/Fractal/FractalProvider.tsx
```diff
@@ -52,8 +51,6 @@ export function FractalProvider({ children }: { children: ReactNode }) {
     initializeConnectedAccount
   );
 
-  useLocalStorage();
```
herbig, line 55: this should be unnecessary, since cache defaults will be set on first call elsewhere

src/providers/Fractal/hooks/account/useLocalStorage.ts
```diff
@@ -21,16 +26,71 @@ const CACHE_MAPPING = [
...
+  const { chainId } = useNetworkConfg();
+
   /**
-   * adds empty objects for setting up caching
+   * Sets cache default values, if we have not already done so.
    */
   useEffect(() => {
-    CACHE_MAPPING.forEach(({ key, defaultValue }) => {
+    CACHE_DEFAULTS.forEach(({ key, defaultValue }) => {
       if (!localStorage.getItem(key)) {
```
herbig, line 52: it feels like maybe we would want to `return` if this not the case, there's no point in iterating the whole array if we know we've already set the defaults.

It doesn't matter much now but as we get more defaults and use local storage more, it's an optimization

src/providers/Fractal/hooks/account/useLocalStorage.ts
```diff
@@ -21,16 +26,71 @@ const CACHE_MAPPING = [
...
+  const getValue = useCallback(
+    (key: string) => {
+      const rawVal = localStorage.getItem('fractal_' + chainId + '_' + key);
+      if (rawVal) {
+        const parsed: IStorageValue = JSON.parse(rawVal);
+        if (parsed.expiration === CacheExpiry.NEVER) {
+          return parsed.value;
+        } else {
+          if (parsed.expiration < Date.now()) {
+            localStorage.removeItem('fractal_' + chainId + '_' + key);
```
herbig, line 82: do we need to remove the item, or is it fine to just let an expired value sit in local storage?

@mudrila @Da-Colon I'm not sure if removeItem has any kind of performance hit vs. the extra local storage use, maybe I'm overthinking it

src/providers/Fractal/FractalProvider.tsx
herbig: update: cache defaults are returned dynamically when there's no value to give

sethhrbek (approved):

PR #1002, opened 2023-02-21
single character cache keys

herbig:
## Description

a small tweak to reduce total cache size, as these keys will be used constantly in future caching


src/hooks/DAO/useCreateSubDAOProposal.ts
```diff
@@ -46,7 +46,7 @@ export const useCreateSubDAOProposal = () => {
               predictedGnosisSafeAddress,
             ]),
           ],
-          title: 'Create SubDAO',
+          title: t('Create a subDAO', { ns: 'proposalMetadata' }),
```
herbig, line 49: also threw in this hardcoded metadata fix

PR #1006, opened 2023-02-22
Cache ENS and DAO name lookups

herbig:
## Description

Cache requests to look up ENS names, as well as our own DAO name contract.

We could refactor ENS resolving to use `useEnsResolver`, similar to what we're doing with `useEnsName` and get their built in caching, but I decided to leave as is for now.


Da-Colon (approved):
💯

Issue #1010, opened 2023-02-22
Bug: Remove Quorum bar


herbig:
### Overview

The second "Quorum" bar doesn't work properly, and is the same info as the line indicator on the Support bar above it, so let's remove it from the UI.

### Steps to reproduce

on Azorius proposals

### Screenshots / Video

<img width="335" alt="Screenshot 2023-02-22 at 12 41 37 PM" src="https://user-images.githubusercontent.com/384559/220715823-16c96729-7e3d-4ab0-9703-2264006e441e.png">

### Version

Dev (app.dev.fractalframework.xyz/)

### Browser

All

### Operating System

None

### web3 wallet

None

PR #1015, opened 2023-02-22
remove quorum bar and tweak quorum indicator

herbig:
## Description

Removes the redundant quorum bar and shifts the indicator line slightly so it's touching the progress indicator

closes https://github.com/decent-dao/fractal-interface/issues/1010

### DIFF
--- a/src/components/ui/utils/ProgressBar.tsx
+++ b/src/components/ui/utils/ProgressBar.tsx
@@ -36,7 +36,7 @@ export default function ProgressBar({
         <Box
           position="absolute"
           top="-10px"
-          left={`${requiredValue}%`}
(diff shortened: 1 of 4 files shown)


src/components/ui/utils/ProgressBar.tsx
```diff
@@ -36,7 +36,7 @@ export default function ProgressBar({
         <Box
           position="absolute"
           top="-10px"
-          left={`${requiredValue}%`}
+          left={`${requiredValue - 0.9}%`}
```
herbig, line 39: this line is a few pixels too far, so it doesn't appear as though you've hit quorum

this shifts it slightly to show that you have

PR #1030, opened 2023-02-24
Cache failed proposal states, to remove RPC errors

herbig:
## Description

Caches Usul TxProposalState.Rejected and TxProposalState.Failed state lookups indefinitely.  The Usul state method throws an error in those cases, rather than return a real state.

This will not be necessary with Azorius, which fixes this issue.

## Notes

You should now only see a single RPC error occur, further lookups will use the cache.


src/providers/Fractal/hooks/account/useLocalStorage.ts
```diff
@@ -49,6 +50,43 @@ function keyInternal(chainId: number, key: string): string {
   return 'fract_' + chainId + '_' + key;
 }
 
+export const setValue = (
```
herbig, line 53: I moved these to regular static functions, which take a chainId, so you can use them in contexts where you can't use hooks.

you can still get them along with the chainId with the useLocalStorage hook

src/providers/Fractal/utils/usul.ts
```diff
@@ -26,8 +27,16 @@ import { Providers } from '../types/ethers';
...
-  proposalId: BigNumber
+  proposalId: BigNumber,
+  chainId: number
 ): Promise<TxProposalState> => {
+  const cache: TxProposalState = getValue(
+    CacheKeys.PROPOSAL_STATE_PREFIX + linearVotingMasterCopyContract.asSigner.address + proposalId,
+    chainId
+  );
+  if (cache) {
+    return cache;
```
herbig, line 38: returns the cached state rather than doing an RPC call, if we have it

src/providers/Fractal/utils/usul.ts
```diff
@@ -43,8 +52,24 @@ export const getTxProposalState = async (
         return TxProposalState.Queueable;
       } catch (e: any) {
         if (e.message.match(ProposalIsPassedError.MAJORITY_YES_VOTES_NOT_REACHED)) {
+          setValue(
```
herbig, line 55: these can be cached forever

Da-Colon (approved):
Works wonderfully, console is readable again :D

PR #1037, opened 2023-02-25
Bugfix rpc requests reduction

Da-Colon:
## Description
LFG. This PR makes 3 changes
- removes duplicate use of `useTokenData` hook. replaces with data that is already loaded in state
- removes unneeded dependencies in `useTokenData` useEffects
- adds local caching to `useTokenData` hook

## Notes
- This reduces `eth_call` requests by 60-80%!!
- To see the `eth_call` requests you can add the following above in Fractal Provider
```ts
  useEffect(() => {
    const listener = (ev: any) => {
      if (ev.data.data?.data?.method === 'eth_call') {
        console.count('ETH_CALL COUNT');
      }
    };
    window.addEventListener('message', listener);
    return () => {
      window.removeEventListener('message', listener);
    };
  }, []);
  ```

### DIFF
--- a/src/components/Proposals/UsulDetails/index.tsx
+++ b/src/components/Proposals/UsulDetails/index.tsx
@@ -6,7 +6,6 @@ import { ProposalAction } from '../../../components/Proposals/ProposalActions/Pr
 import ProposalSummary from '../../../components/Proposals/ProposalSummary';
 import ProposalVotes from '../../../components/Proposals/ProposalVotes';
 import { BACKGROUND_SEMI_TRANSPARENT } from '../../../constants/common';
-import useTokenData from '../../../providers/Fractal/governance/hooks/useGovernanceTokenData';
 import useUpdateProposalState from '../../../providers/Fractal/governance/hooks/useUpdateProposalState';
 import { useFractal } from '../../../providers/Fractal/hooks/useFractal';
 import { TxProposalState, UsulProposal } from '../../../providers/Fractal/types';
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/components/Proposals/UsulDetails/index.tsx [hunk 1]
```diff
@@ -21,13 +20,16 @@ export function UsulProposalDetails({ proposal }: { proposal: UsulProposal }) {
...
   const { chainId } = useNetworkConfg();
   const updateProposalState = useUpdateProposalState({ governance, governanceDispatch, chainId });
 
   const { address: account } = useAccount();
 
   useEffect(() => {
+    const timeLockPeriod = governance.governanceToken?.timeLockPeriod;
+    if (!timeLockPeriod) {
+      return;
+    }
```
herbig, line 32: minor point but these can be simplified to `if (!governance.governanceToken?.timeLockPeriod) {`

src/components/ui/proposal/ProposalTime.tsx
```diff
@@ -39,6 +38,11 @@ function ProposalTime({ proposal }: { proposal: TxProposal }) {
   const usulProposal = proposal as UsulProposal;
 
   useEffect(() => {
+    const timeLockPeriod = governance.governanceToken?.timeLockPeriod;
```
herbig, line 41: same here I believe

herbig (approved):
looks good, we should hammer it over the weekend / on Monday to make sure this hasn't borked anything..

Issue #1035, opened 2023-02-25
Bug: Extend clickability of Proposals / Treasury info to the whole card


herbig:
### Overview

This is on me, I made these cards clickable at some point and didn't realize it's only the content that is clickable, not the full height of the card.

### Steps to reproduce

click the cards below the content

### Screenshots / Video

<img width="473" alt="Screenshot 2023-02-25 at 11 18 54 AM" src="https://user-images.githubusercontent.com/384559/221367779-664dbe42-3a9b-43d8-b2a3-cf33630cf29b.png">

### Version

Dev (app.dev.fractalframework.xyz/)

### Browser

Firefox

### Operating System

None

### web3 wallet

None

PR #1041, opened 2023-02-26
make full infobox clickable in DAO home

herbig:

fixes https://github.com/decent-dao/fractal-interface/issues/1035


Da-Colon (commented):
Governance appears clickable, but doesn't do anything. probably should have cursor be default if no link?

herbig:
> Governance appears clickable, but doesn't do anything. probably should have cursor be default if no link?

oh right, good call. fixed!

Da-Colon (approved):

PR #1048, opened 2023-03-01
Feature  | Parent Token Allocation

Da-Colon:
## Description
Adds Parent Token Claiming

- Adds Parent Token UI Component to Create SubDAO workflow
- Updates DaoBuilder and UsulBuilder to build and deploy claim contract
- Add UI to DAO Dashboard to claim tokens

## Testing
- Create a Parent Token DAO
--Allocate tokens to more than one account (double check that allocations are being spilt correctly)
- Create a SubDAO Proposal and allocate an amount to parent Allocation
- After Proposal Passes, visit new DAO

localhost_3000_(Desktop) (42)

*Amount is different as I have made an update to this screenshot (spacing)
localhost_3000_ (46)


src/pages/DaoDashboard/TokenClaim.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  useEffect(() => {
+    loadClaim();
+  }, [loadClaim]);
+
+  const claimToken = async () => {
+    if (!tokenClaimContract || !governanceToken || !account) {
+      return;
+    }
+    const claimableString = utils.formatUnits(userClaimable, governanceToken.decimals);
```
herbig, line 38: we have `numberFormats.formatCoin` which will do this, handle trailing zeros, and the symbol

src/components/DaoCreator/UsulParentTokenAllocation.tsx
```diff
@@ -0,0 +1,5 @@
+import { ICreationStepProps } from './types';
+
+export function UsulTokenAllocations({}: ICreationStepProps) {
+  return <div></div>;
```
herbig, line 4: this looks like it was stubbed but never finished or used!

src/models/DaoTxBuilder.ts
```diff
@@ -96,13 +96,26 @@ export class DaoTxBuilder extends BaseTxBuilder {
...
+    }
+
     // If subDAO, deploy Fractal Module
     if (this.parentDAOAddress) {
       txs.push(this.deployFractalModuleTx!);
     }
 
     txs.push(this.buildExecInternalSafeTx(usulTxBuilder.signatures()));
 
+    // token claim deployment tx is pushed at the end to ensure approval is executed first
```
herbig, line 115: can't the `tokenClaimTx` claim block above just be at the end as well?

src/pages/DaoDashboard/TokenClaim.tsx
```diff
@@ -0,0 +1,89 @@
...
+    contractCall({
+      contractFn: () => tokenClaimContract?.claimToken(account),
+      pendingMessage: t('pendingTokenClaim', { symbol: governanceToken.symbol }),
+      failedMessage: t('failedTokenClaim', { symbol: governanceToken.symbol }),
+      successMessage: t('successTokenClaim', {
```
herbig, line 43: we have other pending / failed / success messages in `transaction.json`, not sure it's the best system, but would probably be good to stick to it for now

src/pages/DaoDashboard/TokenClaim.tsx
```diff
@@ -0,0 +1,89 @@
...
+        symbol: governanceToken.symbol,
+      }),
+      successCallback: async () => {
+        await loadClaim();
+      },
+    });
+  };
+
+  if (!governanceToken || userClaimable.isZero()) return null;
+  const claimableString = utils.formatUnits(userClaimable, governanceToken.decimals);
```
herbig, line 54: this could be defined above `claimToken` and used in the success message as well

herbig (changes requested):
- [x] ` {{parentName}}` in the helper text should be filled with the parent's name, or change to just "claimable by all parent token holders"
- [x] The error logic seems a bit off.  It seems like I can't have claimable tokens without at least one address allocated above, I don't think that should be the case.
- [ ] The Advanced component is a lot different from the designs, but honestly I'm not sure what the logic for the original designs should be.. We should discuss with @xraystyle1980 and make sure we're not losing something with this simpler input field.
- [x] The claim component on the DAO homepage has an info i icon, rather than the ! icon

src/components/DaoCreator/formComponents/UsulTokenAllocations.tsx
```diff
@@ -76,6 +93,58 @@ export function UsulTokenAllocations(props: ICreationStepProps) {
...
+                        <LabelComponent
+                          label={t('labelParentAllocation')}
+                          helper={t('helperParentAllocation')}
+                          isRequired={false}
+                        >
+                          <LabelWrapper
+                            errorMessage={
+                              (
+                                values.govToken.parentAllocationAmount?.bigNumberValue &&
+                                (errors.govToken?.parentAllocationAmount as any)
```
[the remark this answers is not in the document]
Da-Colon: yessir since this is a required value. It will show "This value is required" on load or if the field is empty.  since validation is ran on mount. currently we weren't showing 'required' errors. We could switch this up, would just need to create some logic to diasable the button on each screen on intial by whether things were touched* and then create error string for each of these.

src/components/DaoCreator/UsulParentTokenAllocation.tsx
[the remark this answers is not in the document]
Da-Colon: Ah @herbig, I'll answer here as you asked same question. Yeah I was going to extract out at the beginning and stubbed it, got left in. I'll remove.

src/models/UsulTxBuilder.ts
```diff
@@ -224,10 +306,9 @@ export class UsulTxBuilder extends BaseTxBuilder {
         encodedStrategyInitParams,
       ]);
 
-    const strategyByteCodeLinear =
-      '0x602d8060093d393df3363d3d373d3d3d363d73' +
-      this.usulContracts!.linearVotingMasterCopyContract.address.slice(2) +
-      '5af43d82803e903d91602b57fd5bf3';
```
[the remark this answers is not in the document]
Da-Colon: @sethhrbek might be able to explain it better. but this is how the signature must be structured based on the code from contract.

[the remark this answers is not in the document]
sethhrbek: yeah these are the strings that are in gnosis' module proxy factory -- the strings differ slightly between the gnosis module in the npm package and the latest contracts they have on their github. 

When we deploy the newest contract changes, we should also deploy the module proxy factory on goerli (and update the proxy master copy address) to ensure 100% these strings match the ones in gnosis mainnet module proxy factory

herbig:
> > The Advanced component is a lot different from the designs, but honestly I'm not sure what the logic for the original designs should be.. We should discuss with @xraystyle1980 and make sure we're not losing something with this simpler input field.
> 
> image
> 
> O I must be blind did not see this block here when I did this? is this new? this can be a future PR

yeah, we can do that in a followup for sure

🤷 maybe it was added later on?  @xraystyle1980 ?

src/models/DaoTxBuilder.ts
Da-Colon: ah no sir, this transaction must happen after the internal txs which is pushed in on line 113. So that approval can take place.

src/pages/DaoDashboard/TokenClaim.tsx
Da-Colon: ? governanceToken.decimals prevents it from being used above here

src/pages/DaoDashboard/TokenClaim.tsx
```diff
@@ -0,0 +1,87 @@
...
+        justifyContent="space-between"
+        alignItems="center"
+      >
+        <AlertIcon boxSize="24px" />
+        <AlertTitle>
+          <Text
+            textStyle="text-lg-mono-medium"
+            whiteSpace="pre-wrap"
+          >
+            {t('alertTokenClaim', {
```
herbig, line 71: `formatCoin` takes the coin symbol as a parameter, and if it's not provided, formats it as ETH, so I think this will show e.g. "1000 ETH SYMBOL"

`alertTokenClaim` only needs to be  `"You have {{amount}} tokens to claim!"`, without the symbol, and the symbol you provide to `formatCoin`

herbig (approved):

tbwebb22 (approved):
Tested a few different combos of allocation amounts, worked perfectly for me, nice work

PR #1050, opened 2023-03-02
Bugfix proposal state not updating consistently

Da-Colon:
## Testing
- Create/Use a Token DAO with 0 queue period
- Create a proposal and go through the motions
- After Queuing should see it update to execute button correctly

### DIFF
--- a/src/components/ui/proposal/ProposalTime.tsx
+++ b/src/components/ui/proposal/ProposalTime.tsx
@@ -1,7 +1,7 @@
 import { Flex, Text, Tooltip } from '@chakra-ui/react';
 import { Vote, Execute, Lock } from '@decent-org/fractal-ui';
 import { VetoGuard } from '@fractal-framework/fractal-contracts';
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { getTxQueuedTimestamp } from '../../../hooks/utils/useSafeActivitiesWithState';
 import { useFractal } from '../../../providers/Fractal/hooks/useFractal';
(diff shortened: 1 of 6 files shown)


src/components/ui/proposal/ProposalTime.tsx
```diff
@@ -29,11 +29,14 @@ function ProposalTime({ proposal }: { proposal: TxProposal }) {
...
-  const isExecutable = proposal.state === TxProposalState.Executing;
-  const showCountdown = isActive || isTimeLocked || isExecutable || isQueued;
+  const isActive = useMemo(() => proposal.state === TxProposalState.Active, [proposal]);
+  const isTimeLocked = useMemo(() => proposal.state === TxProposalState.TimeLocked, [proposal]);
+  const isQueued = useMemo(() => proposal.state === TxProposalState.Queued, [proposal]);
+  const isExecutable = useMemo(() => proposal.state === TxProposalState.Executing, [proposal]);
+  const showCountdown = useMemo(
+    () => isActive || isTimeLocked || isExecutable || isQueued,
+    [isActive, isTimeLocked, isExecutable, isQueued]
+  );
```
Da-Colon, line 39: Normally these values would be fine to use outside of `useMemo` but because these values are being used within the `useEffect` below, its not updating the state of the component properly when proposal is updated.

src/components/ui/proposal/ProposalTime.tsx
```diff
@@ -51,24 +54,19 @@ function ProposalTime({ proposal }: { proposal: TxProposal }) {
...
+          setCountdown(usulProposal.deadline * 1000 - Date.now());
         }, 1000);
         setCountdownInterval(interval);
       } else if (isTimeLocked && usulProposal.deadline && timeLockPeriod) {
         const interval = setInterval(() => {
-          const now = new Date();
-          setCountdown(
-            (usulProposal.deadline + Number(timeLockPeriod.value)) * 1000 - now.getTime()
-          );
+          setCountdown((usulProposal.deadline + Number(timeLockPeriod.value)) * 1000 - Date.now());
```
Da-Colon, line 62: `new Date().getTime()` and `Date.now()` return the same value. but the former is calculated when the variable is defined and in the time of computing can become inaccurate.

herbig: yeah good call

we should maybe do an audit of Date.now() / getTime() and see if there's any other places we might be doing this

herbig (approved):

PR #1060, opened 2023-03-10
BugFix | Duplicate intervals setting

Da-Colon:
## Description
@herbig  Maybe we can combine this with the work you were doing adding the idle update

These changes prevent `updateTimer` hook from setting more than one interval or the same method. helps save some of those idle calls.

This update changes the timers[] into an object that contains the method 'signature' as a string that can be compared to see if a timer has already been set. if it has clear old timer set new one (this helps make sure we aren't stopping anything to be called when DAOs are changed.

There is still one more extra call being made on first render. might be saved in data layer refactor


herbig:
@Da-Colon I'll try to look at this later today and implement the interval fix I was working on

herbig (approved):

PR #1083, opened 2023-03-23
Refactor - DL | FractalNode hook

Da-Colon:
## Description

The first series of PRs will be code review only, once the main app state has been restored the next series will update the app back to working state. 

This PR covers loading a DAO into state. 

Couple beginning notes here:
- Except for the static items (baseContracts | thridPartyClients) the `AppProvider` (FractalProvider) will no longer have custom hooks that update state. Partly for context seperation, partly to not have the entire tree update as state is worked out. 

This Job will handed off to the useDAOController(). This hook now will call all the nesscary hooks to validate and update the app state. (see `useFractalNode`). 

The last in this first series will have the `useDAOController`, which will begin the failing of tests. As well as replace `FractalProvider`. 

## Notes
- Also for ease and quickness, These PRs only require a single approval, as they won't be merged into develop into the app is back together.


package-lock.json
Da-Colon, line 1: make sure we are all using `nvm use` prior to installing packages and commiting

src/hooks/DAO/loaders/useDAOName.ts
Da-Colon, line 1: This implementation of useDAOName custom hook removes `useENSName` implementation so that it can be called as a callback function versus being passed in the hook (not available to do it this way). I don't believe this implementation will be around long once @mudrila merges in the subGraph worked (I'll update the store accordingly).

src/hooks/DAO/loaders/useFractalModules.ts
Da-Colon, line 1: `useGnosisModuleTypes` refactored to use `baseContracts` from fractalModules and only return the modules instead of setting state

src/hooks/DAO/loaders/useFractalNode.ts
Da-Colon, line 1: This file takes on two different roles, it now validates the DAO from url params (redirecting if not valid). but also loads the `FractalNode` into state. This includes the `getSafeInfo`, `FractalModules` as well as the DAO's name. I believe this is where we would also update to call out to the graph for all FractalRegistry info for the DAO.

src/providers/App/AppProvider.tsx
```diff
@@ -1,29 +1,32 @@
...
 export const FractalContext = createContext<FractalStore | null>(null);
 
 export const useFractal = (): FractalStore => useContext(FractalContext as Context<FractalStore>);
 
 // @RENAME to FractalProvider
-export function AppProvider() {
-  // loads base Fractal contracts with provider into state
-  const baseContracts = useSafeContracts();
-  // loads safe service into state;
-  const safeService = useSafeService();
```
Da-Colon, line 26: moved these below the reducers

src/providers/App/AppProvider.tsx
```diff
@@ -69,6 +77,15 @@ export function AppProvider() {
         governanceContracts: governanceContractsDispatch,
         guardContracts: guardContractsDispatch,
         nodeHierarchy: nodeHierarchyDispatch,
+        resetDAO: () => {
```
Da-Colon, line 80: Single method that can be called to reset all DAO state

package-lock.json
herbig: @decent-dao/fractal

herbig (approved):

PR #1090, opened 2023-03-26
Refactor DataLayer | Freeze Guard and Contracts

Da-Colon:

### DIFF
--- a/src/hooks/DAO/loaders/useFractalGuardContracts.ts
+++ b/src/hooks/DAO/loaders/useFractalGuardContracts.ts
@@ -72,35 +72,34 @@ export const useFractalGuardContracts = () => {
       }
     }
 
-    if (!vetoGuardContract) {
-      return;
+    if (!!vetoGuardContract) {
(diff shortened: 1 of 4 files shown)


src/hooks/DAO/loaders/useFractalGuardContracts.ts
```diff
@@ -0,0 +1,122 @@
...
+          asSigner: gnosisVetoGuardMasterCopyContract.asSigner.attach(guard),
+          asProvider: gnosisVetoGuardMasterCopyContract.asProvider.attach(guard),
+        };
+        vetoGuardType = VetoGuardType.MULTISIG;
+      }
+    }
+
+    if (!vetoGuardContract) {
+      return;
+    }
```
herbig, line 77: it might be better to do `if (vetoGuardContract) { setup stuff for it }` rather than if ! return

that way if we have further things to add later on, we don't make the mistake of throwing them into the end of this code and assuming it's going to work

herbig (approved):

PR #1101, opened 2023-03-31
Don't init Sentry on prod

herbig:
## Description

Prevents even initializing or calling any Sentry related functions on production builds.


Da-Colon (approved):
LGTM.

PR #1106, opened 2023-04-02
Refactor Bugfix | erc20 account update

Da-Colon:
## Description
Noticed that the logic I had here was a little off and was preventing updates when DAOs are changed.

This also updates the `showDelegate` variable on the proposals page. it would sometimes error out trying to calculate token balance on first load.

### DIFF
--- a/app/daos/[daoAddress]/proposals/page.tsx
+++ b/app/daos/[daoAddress]/proposals/page.tsx
@@ -3,6 +3,7 @@
 import { Box, Button, Show } from '@chakra-ui/react';
 import { AddPlus } from '@decent-org/fractal-ui';
 import Link from 'next/link';
+import { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useAccount } from 'wagmi';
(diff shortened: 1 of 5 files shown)


app/daos/[daoAddress]/proposals/page.tsx
```diff
@@ -21,12 +22,18 @@ export default function ProposalsPage() {
   } = useFractal();
   const { type } = governance;
   const { address: account } = useAccount();
-  const azoriusGovernance = governance as AzoriusGovernance;
   const delegate = useFractalModal(ModalType.DELEGATE);
-  const showDelegate =
-    type === StrategyType.GNOSIS_SAFE_USUL &&
-    !!azoriusGovernance.votesToken.balance &&
-    azoriusGovernance.votesToken.balance.gt(0);
+  const showDelegate = useMemo(() => {
```
Comment (herbig) [inline app/daos/[daoAddress]/proposals/page.tsx:26]: is there any thought to putting this in the data provider? so it's just a parameter you can use anywhere

herbig (approved):

PR #1114, opened 2023-04-06
Refactor | Combine Reducers

Da-Colon:
## Description
This update takes care of a few things
- Improves the loading of Nodes, and removes more unnecessary requests
- This pattern improves the ability to unload state when DAOs are changed or invalidated

## Notes
- Fixes one of the gnosis multsig create form bugs. Left it in a commit.


src/components/DaoCreator/formComponents/GnosisMultisig.tsx
```diff
@@ -55,7 +55,7 @@ export function GnosisMultisig(props: ICreationStepProps) {
         setFieldValue('gnosis.trustedAddresses', gnosisAddresses);
       }
     }
-    setFieldValue('gnosis.numOfSigners', inputValue);
+    setFieldValue('gnosis.numOfSigners', numOfSigners);
```
Da-Colon, line 58: When I pivoted to work on this update. I left in the changes for another bug here. This update prevents the input on Gnosis Signers. from being over the inputted amount

src/providers/App/combinedReducer.ts
Da-Colon, line 1: This combines all the reducers into a single reducer function. the main goal of this is to let `resetDAO` to effect the entire state at once. Originally this was impossible do to the state being spread across many reducers. This setup also exposed a few other issues with loading that I was able to fix as well.

src/providers/App/governance/action.ts
```diff
@@ -11,19 +11,18 @@ import {
...
+  UPDATE_NEW_USUL_VOTE = 'UPDATE_NEW_USUL_VOTE',
+  UPDATE_PROPOSAL_STATE = 'UPDATE_PROPOSAL_STATE',
+  UPDATE_VOTING_PERIOD = 'UPDATE_VOTING_PERIOD',
+  UPDATE_VOTING_QUORUM = 'UPDATE_VOTING_QUORUM',
+  UPDATE_TIMELOCK_PERIOD = 'UPDATE_TIMELOCK_PERIOD',
+  SET_TOKEN_DATA = 'SET_TOKEN_DATA',
+  SET_TOKEN_ACCOUNT_DATA = 'SET_TOKEN_ACCOUNT_DATA',
+  SET_CLAIMING_CONTRACT = 'SET_CLAIMING_CONTRACT',
+  RESET_TOKEN_ACCOUNT_DATA = 'RESET_TOKEN_ACCOUNT_DATA',
 }
```
Da-Colon, line 26: So the combined reducer setup calls for these actions to be `string` enums. This ensure that when actions are dispatched they are caught by the appropriate reducer

src/providers/App/guard/reducer.ts
```diff
@@ -31,8 +31,6 @@ export function guardReducer(state: FreezeGuard, action: FractalGuardActions) {
         userHasVotes,
       };
     }
-    case FractalGuardAction.RESET:
-      return initialGuardState;
     default:
       return state;
   }
```
Da-Colon, line 36: these now become important to have. need to ensure that each reducer returns current state if action is not updating that state

src/utils/guard.ts
```diff
@@ -25,7 +25,7 @@ export async function getTxQueuedTimestamp(activity: Activity, vetoGuard: VetoGu
         [
           multiSigTransaction.to,
           multiSigTransaction.value,
-          ethers.utils.solidityKeccak256(['bytes'], [multiSigTransaction.data]),
+          ethers.utils.solidityKeccak256(['bytes'], [multiSigTransaction.data || '0x']),
```
Da-Colon, line 28: So I found sometimes during loading I was getting an error on 'Send Eth' transactions as I believe the data for these are null. Probably need to follow up with a bug report on this.

herbig (approved):
A+

PR #1121, opened 2023-04-12
Refactor |  Next set of various bug fixes

Da-Colon:
## Description
Some more bugs from the list.

- [x] Clicking Token Voting Reject (No vote) doesn't work.
- [x] (MAJOR) Token Voting parent with multisig child -> View Hierarchy of subDAO, click into the parent DAO, governance details and transactions of the child are incorrectly displayed.
- [x] Token Voting Governance card shows Quorum as "less than 5 seconds" (screenshot).
- [x] Address the issue where the favorite state in DAO Hierarchy doesn't update for subDAOs.
- [x] Clicking favorite on a subDAO from the hierarchy screen seems to favorite it, but doesn't change the star icon.
- [x] ReAdd ENS to DAOName loading
- [x] ENS doesn't work in Send Assets modal.

## Notes
On top the above, saw some issues with the listeners when voting. I separated into separate `useEffect` and seem to work smoother

### DIFF
--- a/src/components/pages/DaoHierarchy/DaoNode.tsx
+++ b/src/components/pages/DaoHierarchy/DaoNode.tsx
@@ -69,7 +69,7 @@ export function DaoNode({
         trueDepth={trueDepth}
         numberOfSiblings={numberOfSiblings}
         numberOfChildren={fractalNode?.nodeHierarchy.childNodes.length}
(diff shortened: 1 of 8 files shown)


herbig (approved):
Looks good, aside from @mudrila's question about logic

PR #1130, opened 2023-04-17
Add remove owners

herbig:
## Description
This PR:
1) Adds a "Manage Signers" option on the "Manage DAO" menu option
2) Adds a page listing the current safe signers, and buttons for adding and removing signers
3) Adds modals for both adding and removing signers, which allows a user to create a proposal to do either action

image
image
image
image

## Testing
1) Create a proposal to add a signer, execute the proposal and verify that the signer was added to the safe
2) Create a proposal to remove a signer, execute the proposal and verify that the signer was removed from the safe


herbig:
Fixed the original PR, so closing this

Issue #1135, opened 2023-04-18
Execute Proposal Permissions


sethhrbek:
**Bug**
As a non-signer on a Multisig, it shows me the "Execute Proposal" button and allows me to click "Execute Proposal". The proposal will fail.

**Expected Behavior**
As a non-signer on a Multisig, it does not show the "Execute Proposal" button.

**Steps to Recreate**
1. Create a proposal on a multisig as one of the signers
2. View the proposal as a non-signer, and see that you can see the execute proposal button.

PR #1140, opened 2023-04-19
Readonly state variables

herbig:
## Description

Adds "read only" state variables to make common calculations more consistent.

Currently only implementing a few, but more will be added in future refactors.

fixes https://github.com/decent-dao/fractal-interface/issues/1135

## Testing

This impacts all aspects of the app, but most notably switching between wallets.  Please confirm this works seemlessly.


src/providers/App/useReadOnlyValues.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const useReadOnlyValues = () => {
+  const readOnlyValues = useCallback((_state: Fractal, _account: string | undefined) => {
+    const { node, governance } = _state;
+
+    const isSigner = _account && node.safe?.owners.includes(_account);
+    const tokenWeight =
+      (governance as AzoriusGovernance).votesToken?.votingWeight || BigNumber.from(0);
+
+    const votingWeight = () => {
```
herbig, line 13: @Da-Colon I realized this was really what we need, rather than `isOwnerOrDelegate`, just the user's total voting power in whatever DAO is connected

Da-Colon (approved):

PR #1141, opened 2023-04-20
Usul to Azorius Rename

herbig:
## Description

Replaces all instances in the codebase of "Usul" with "Azorius".

This should help with the upcoming refactor by eliminating unnecessary line changes, and help code review.

## Notes

- Typechain types still reference the Usul contracts (FractalUsul, UsulVetoGuard)
- Usul is also mentioned in `fractal.ts` in reference to the Proposal states, but that will be revamped with the full refactor.

## Testing

This should have no front end impact, so the app should all work as it did before, with no visible changes.


src/models/AzoriusTxBuilder.ts
```diff
@@ -86,28 +86,28 @@ export class UsulTxBuilder extends BaseTxBuilder {
   public buildLinearVotingContractSetupTx(): SafeTransaction {
     return buildContractCall(
       this.linearVotingContract!,
-      'setUsul',
-      [this.usulContract!.address],
+      'setAzorius',
```
herbig, line 89: @decent-dao/fractal this is the one change I wasn't entirely sure of

what is this `setUsul` string here doing, and will that break by changing it to Azorius?

tbwebb22: So looks like the `setUsul` string is the function name, so yeah if we change this string to `setAzorius`, it will break things since the FE is still using the old  contracts where the function name is `setUsul`.

herbig: awesome, ok I'll switch that back and we should be good

everything else is just a variable, string, or import name

Da-Colon (approved):
Merged into my branch no problems. LGTM

tbwebb22 (approved):
LGTM

PR #1142, opened 2023-04-21
Feature |  Import Existing Token (Create Workflow)

Da-Colon:
## Description
This PR adds the ability to import an existing ERC20 token and wrap it for use as a Governance token in a DAO. It covers the first part of deploying a new DAO with this new feature and updates the UI during the token configuration step.

## Main changes
- Modified `AzoriusTxBuilder.ts` to handle the import and wrapping of existing ERC20 tokens.
- Updated `AzoriusTokenDetails.tsx` to include new UI elements for the token configuration step.

## Additional Notes
- Fixed an issue with Toasts not properly closing and a missing translation.
- This PR also covers checking for tokens that are already presumed to be ERC20 Votes tokens.

## Testing
1. Create a new Token DAO using the following methods:
   - As normal
   - With existing VotesToken
   - With existing ERC20 token
2. Ensure that the DAO deploys correctly and shows up as a Token Voting DAO. Note that there is no other way to ensure proper deployment until the other end of this feature is worked on.

localhost_3000_create


src/components/DaoCreator/formComponents/AzoriusTokenDetails.tsx
```diff
@@ -1,82 +1,183 @@
...
+      setFieldTouched('govToken.tokenName', true, true);
+    }, 0);
+  }, [
+    checkVotesToken,
+    errors?.govToken?.tokenImportAddress,
+    setFieldValue,
+    setFieldTouched,
+    provider,
+    values.govToken.tokenImportAddress,
+  ]);
```
Da-Colon, line 75: handles the logic to update the tokenName/tokenSymbol. for display purposes

src/components/DaoCreator/formComponents/AzoriusTokenDetails.tsx
```diff
@@ -1,82 +1,183 @@
...
+              <RadioWithText
+                label={t('radioLabelExistingToken')}
+                description={t('helperExistingToken')}
+                testId="choose-existingToken"
+                value={TokenCreationType.IMPORTED}
+                onClick={() => {
+                  setFieldValue('govToken.tokenName', '');
+                  setFieldValue('govToken.tokenSymbol', '');
+                }}
+              />
```
Da-Colon, line 133: These `onClick` makes sure to empty the fields when radio is switched

src/components/DaoCreator/formComponents/AzoriusTokenDetails.tsx
```diff
@@ -1,82 +1,183 @@
...
+                isDisabled={values.govToken.tokenCreationType === TokenCreationType.NEW}
+              >
+                <Input
+                  name="govToken.tokenImportAddress"
+                  onChange={handleChange}
+                  value={values.govToken.tokenImportAddress}
+                  placeholder={createAccountSubstring(constants.AddressZero)}
+                  isDisabled={values.govToken.tokenCreationType === TokenCreationType.NEW}
+                />
+              </LabelWrapper>
```
Da-Colon, line 149: The error message flickers very briefly, I've tried to see if its the useEffect or the test itself but no luck. I'm not sure the cause.

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -43,7 +43,7 @@ export const useFractalNode = ({
     action,
   } = useFractal();
   const { getDaoName } = useLazyDAOName();
-  const { t } = useTranslation();
+  const { t } = useTranslation('dashboard');
```
Da-Colon, line 46: Noticed this was missing the library

src/components/DaoCreator/hooks/usePrepareFormData.ts
```diff
@@ -1,14 +1,31 @@
...
+      try {
+        const votesContract = new ethers.Contract(address, ERC20Votes__factory.abi, provider);
+        await votesContract.estimateGas.delegate('0x0000000000000000000000000000000000000001');
+        return true;
+      } catch (error) {
+        return false;
+      }
+    },
+    [provider]
+  );
```
Da-Colon, line 28: checks if `delegate` function exists. linking*

herbig (changes requested):
Hey @Da-Colon this works flawlessly! I was able to use Wrapped Ether to govern my DAO ha.

A few things to discuss / fix:

- [x] We should use the same margin between "Select Token" and the component that we use between "Token Metadata" and its component below it
- [ ] Can we instead have the Existing Token address input and blue text appear / disappear depending on if it's selected?
- [x] does the blue color make it look like a hyperlink? not sure if we should change it, just wondering
- [ ] we should decide on a tooltip max width and apply it throughout the app (anything works, just the full line is too long)
- [ ] This is a weird bug, but the click space to change back to the first "New Token" option is basically the whole dark box except for the Existing Token radio option.  Try selecting the second option then clicking the blue text

tbwebb22:
When selecting a token, when I input an existing token that does have the ERC20Votes functionality, it still seems that a new wrapped token is being deployed for the DAO. In this case, we wouldn't want to re-deploy a token, and just let the existing VotesToken be used, right?

image

herbig:
> When selecting a token, when I input an existing token that does have the ERC20Votes functionality, it still seems that a new wrapped token is being deployed for the DAO. In this case, we wouldn't want to re-deploy a token, and just let the existing VotesToken be used, right?
> 
> image

Oh true good point,

It's not actually being deployed as wrapped, but the logic should be:

No blue text until the address if legit

Then blue text appears based on whether it's being wrapped or not

tbwebb22 (approved):
Fix for my comment was implemented on 1144, so LGTM now

src/components/DaoCreator/formComponents/AzoriusTokenDetails.tsx
herbig: strange, yeah I think we could use an app-wide audit of error handling at some point

PR #1145, opened 2023-04-27
BugFix | Tooltips don't appear in Modals

### LINKED ISSUE #1127: Bug: Tooltips in Modals don't appear
### Overview

We have two tooltips in modals that don't appear:

1. Multisig Send Assets modal
2. Add / remove signers modal

This likely has something to do with the z index of the tooltip and modal

### Steps to reproduce

View the Send Assets modal of a multisig DAO

### Screenshots / Video

_No response_

### Version

### Browser

All

### Operating System

None

### web3 wallet

None

Da-Colon:
## Description
That was a tough one to solve. Basically found that there were nested portals going on and the Tooltips parent ref got attached to a parent portal. By declaring the ref and setting to the immediate parent, the tooltip now shows up properly

## Notes
Also removed `autoClose` off the main `ToastContainer` component this doesn't needed to be set here

Closes #1127


tbwebb22 (commented):
LFG - nice work on this one

I think there are still a few other tooltips that need to be updated - like the update required signers, and there is another one I think on either delegate or send assets modal

herbig:
> LFG - nice work on this one
> 
> I think there are still a few other tooltips that need to be updated - like the update required signers, and there is another one I think on either delegate or send assets modal

@tbwebb22 this issue is fixed now! give it a relook when you get a chance

tbwebb22 (approved):
Noice

PR #1161, opened 2023-05-04
BugFix | remove `useDefaultNonce` hook use `safe.nonce` instead

Da-Colon:
## Description

Since this involved changing a few different files broke it out into its own PR. This addresses @herbig 's comment ` Create proposal doesn't work on a multisig DAO, button enables but clicking it doesn't open Metamask, notification briefly pops up. It works fine on Token voting however.`

I realized we didn't need the `useDefaultNonce` hook at all. and replaced with `safe.nonce` instead. 

## Notes
- You will notice also that I added nonce to Yup and updated it with the validation.

## Testing

Create a Proposal in a Multsig DAO, and any where else where nonce may be used.


src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -131,6 +131,7 @@ export default function useSubmitProposal() {
 
       try {
         const gnosisContract = GnosisSafe__factory.connect(safeAddress, signerOrProvider);
+        console.log('🚀 ~ file: useSubmitProposal.ts:134 ~ gnosisContract:', gnosisContract);
```
herbig, line 134: don't need this anymore!

herbig (approved):
nice

PR #1162, opened 2023-05-08
cleans up useDecodedTransaction to cache requests

Da-Colon:
## Description
Uses The browser's indexedDB to cache some of these calls. I believe this hook can be used to cache data that will not change over time. There is a lot of opportunity here. I modeled it using `useLocalStorage` setup as a model. and this pretty much can be used in the same way. except by passing the Object's store key into the hook as a param

### DIFF
--- a/src/hooks/utils/useDecodeTransaction.ts
+++ b/src/hooks/utils/useDecodeTransaction.ts
@@ -1,17 +1,18 @@
 import axios from 'axios';
-import { useCallback, useRef } from 'react';
+import { useCallback } from 'react';
 import { useNetworkConfg } from '../../providers/NetworkConfig/NetworkConfigProvider';
 import { DecodedTransaction, DecodedTxParam, MetaTransaction } from '../../types';
 import { buildGnosisApiUrl, parseMultiSendTransactions } from '../../utils';
+import { DBCacheKeys, useIndexedDB } from './useLocalDB';
 
 export const useDecodeTransaction = () => {
   const { safeBaseURL } = useNetworkConfg();
-  const metaDataMapping = useRef<Map<string, DecodedTransaction[]>>(new Map());
+  const [setValue, getValue] = useIndexedDB(DBCacheKeys.DECODED_TRANSACTIONS);
 
   const decodeTransactions = useCallback(
-    async (proposalId: string, transactions: MetaTransaction[]) => {
+    async (strategyAddress: string, proposalId: string, transactions: MetaTransaction[]) => {
       const apiUrl = buildGnosisApiUrl(safeBaseURL, '/data-decoder/');
-      const cachedTransactions = metaDataMapping.current.get(proposalId.toString());
+      const cachedTransactions = await getValue(strategyAddress + '_' + proposalId.toString());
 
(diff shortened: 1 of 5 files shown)


src/utils/azorius.ts
Da-Colon, line 1: This looks like it snuck in unrelated to this PR. This removes an extra request that was happening

src/hooks/utils/useLocalDB.ts
```diff
@@ -0,0 +1,163 @@
...
+  });
+};
+
+export const getIndexedDBValue = async (
+  storeKey: string,
+  key: string,
+  chainId: number
+): Promise<any> => {
+  return withIndexedDB(async db => {
+    const transaction = db.transaction(storeKey, 'readonly');
```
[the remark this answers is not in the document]
herbig: ah good call, yeah I hadn't considered security settings disabling this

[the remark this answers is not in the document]
herbig: also @Da-Colon it would be useful to have some documentation on what the best use case is for this over caching to local storage

src/hooks/utils/useDecodeTransaction.ts
```diff
@@ -62,13 +64,17 @@ export const useDecodeTransaction = () => {
             }
           })
         );
-        metaDataMapping.current.set(proposalId.toString(), decodedTransactions.flat());
-        return decodedTransactions.flat();
+        const flattened = decodedTransactions.flat();
+        await setValue(
+          CacheKeys.DECODED_TRANSACTION_PREFIX + strategyAddress + '_' + proposalId.toString(),
```
herbig, line 69: @Da-Colon I added a prefix here, just on the off chance we cache something else in the future with "strategyAddress + proposalId" we'd get bugs or weirdness and it would be difficult to figure out why

src/hooks/utils/cache/useLocalStorage.ts
```diff
@@ -1,54 +1,6 @@
...
-interface IStorageValue {
-  // the value to store, 1 character to minimize cache size
-  v: any;
-  // the expiration, as a UTC timestamp
-  e: number;
-}
-
-function keyInternal(chainId: number, key: string): string {
-  return 'fract_' + chainId + '_' + key;
-}
```
herbig, line 51: I moved all this stuff to a common file for both caches to use, rather than duplicating in both

src/hooks/utils/cache/useLocalDB.ts
```diff
@@ -0,0 +1,149 @@
...
+    let transaction: IDBTransaction, store: IDBObjectStore, request: IDBRequest<any>;
+
+    try {
+      transaction = db.transaction(storeKey, 'readonly');
+      store = transaction.objectStore(storeKey);
+      request = store.get(keyInternal(chainId, key));
+    } catch (e) {
+      logError(e);
+      return Promise.reject(e);
+    }
```
herbig, line 99: @Da-Colon @mudrila added try / catch here and in the set function for the initial db setup

herbig (approved):

Issue #1154, opened 2023-05-03
Snapshot Integration


herbig:
Integrating Snapshot spaces into the UI.

Notion Doc:
https://www.notion.so/hulking-cosmos-3e8/Snapshot-Simple-Integration-d7c07bb6528548c4812d8bad0ecc3987

Design feedback:
https://www.figma.com/file/cO1nBoNZCmMKCwCCpyTxfx/branch/FNlE2DBziIcVmU6IsYpk8M/%F0%9F%A6%84-Fractal-%7C-Pages-%26-Flows-Master?node-id=4934-31575&t=e4LxaW2LW6p8kyLr-0

Final Designs:
https://www.figma.com/file/cO1nBoNZCmMKCwCCpyTxfx/%F0%9F%A6%84-Fractal-%7C-Pages-%26-Flows-Master?node-id=4204-28213&t=6fIw2nP9PF131wre-0

PR #1166, opened 2023-05-15
Snapshot integration

tbwebb22:
## Description
1) This PR adds Snapshot integration for DAOs. When creating a DAO, users can configure a Snapshot ENS name for the DAO, and any proposals created within Snapshot will then be displayed in the DAO proposal list.

https://github.com/decent-dao/fractal-interface/issues/1154

## Testing
1) Go to [Snapshot ](https://demo.snapshot.org/#/) and set up a space with an ENS name. 
2) Within Fractal, create a DAO and set the ENS name used in step 1
3) Create some proposals in the Snapshot space and verify they show up in the DAO proposals list

image


herbig (changes requested):
Hey @tbwebb22 a couple bugs, I wasn't able to fully test this before getting 500 errors on everything. I don't get the same errors on dev for the DAO I created, so I assume something is up because of the SnapShot changes.

- [x] let's change the hit from fractaldao.eth to decent-dao.eth for now. Hopefully we can scoop that up when it expires, but until we own it I don't want to give the impression that it's our address
- [x] ENS validation seems a bit weird, but I can't pinpoint how. It seems easy to get an invalid address, e.g. asdf.eth worked for me, but asdff.eth didn't
- [x] the SnapShot button / icon should be much smaller and right aligned, as in the designs
- [x] the SnapShot button in DAO homepage links to https://demo.snapshot.org, not the SnapShot space
- [ ] I keep getting "500 Internal Server Error" on pages e.g. https://deploy-preview-1166--fractal-framework-interface-dev.netlify.app/daos/0x059Cb714b9dF7110b40116F5Ec6Cc4DCa8dD8219
- [ ] I set the above DAO to use Aave's SnapShot, but didn't see any of their proposals, before getting the permanent 500 error.

tbwebb22:
@herbig Regarding the ENS name validation, it looks to me like asdf.eth is a registered ENS name, while asdff.eth is not. So the behavior you're describing seems right to me.

herbig:
> @herbig Regarding the ENS name validation, it looks to me like asdf.eth is a registered ENS name, while asdff.eth is not. So the behavior you're describing seems right to me.

Oh shoot, yeah.

src/components/Activity/ActivityCard.tsx
```diff
@@ -34,6 +36,13 @@ export function ActivityCard({
             mb="1rem"
           >
             {Badge}
+            {isSnapshot && (
+              <Image
+                src="/images/snapshot-icon.svg"
+                alt="snapshot icon"
```
herbig, line 42: we haven't been great about having alt text in the app, but if we do add it we should move this to the translation json files.  Also fine to just remove this for now..

src/components/DaoCreator/formComponents/EstablishEssentials.tsx
```diff
@@ -66,6 +68,27 @@ export function EstablishEssentials(props: ICreationStepProps) {
...
         mb={4}
       />
+      <LabelComponent
+        label={t('snapshot')}
+        helper={t('snapshotHelper')}
+        isRequired={false}
+      >
+        <LabelWrapper errorMessage={errors?.essentials?.snapshotURL}>
+          <Input
+            id="searchEssentials-snapshotURL"
```
herbig, line 78: we have an id for the DAO name, but I'm not really sure why, don't think we need this here

src/components/Proposals/index.tsx
```diff
@@ -37,6 +39,7 @@ const FILTERS_MULTISIG_CHILD = [
 
   FractalProposalState.REJECTED,
   FractalProposalState.EXPIRED,
+  FractalProposalState.CLOSED,
```
herbig, line 42: @tbwebb22 should `PENDING` be in these as well?

src/hooks/schemas/common/useValidationAddress.tsx
```diff
@@ -7,6 +7,25 @@ import { useFractal } from '../../../providers/App/AppProvider';
 import { AddressValidationMap } from '../../../types';
 import { Providers } from '../../../types/network';
 
+export async function validateENSName({ ensName }: { ensName: string }) {
+  if (!!ensName && ensName.trim() && ensName.endsWith('.eth') && ensName.length >= 7) {
```
herbig, line 11: emojis count as 2 characters I think, so I'm pretty sure `🚀🚀.eth` would pass this.  We might need something to handle emoji counting

herbig: maybe ` _.toArray('😁').length` here instead

src/hooks/schemas/common/useValidationAddress.tsx
```diff
@@ -7,6 +7,25 @@ import { useFractal } from '../../../providers/App/AppProvider';
 import { AddressValidationMap } from '../../../types';
 import { Providers } from '../../../types/network';
 
+export async function validateENSName({ ensName }: { ensName: string }) {
```
herbig, line 10: minor thing but I don't think this needs to be async

src/i18n/locales/en/daoCreate.json
```diff
@@ -84,5 +84,7 @@
     "helperNewToken": "Create a new ERC-20 token to vote with",
     "helperExistingToken": "Use an existing ERC-20 for voting",
     "warningExistingToken": "This token must be wrapped in order to be used as a voting token",
-    "warningExistingTokenTooltip": "Fractal will deploy a new ERC-20 governance token to wrap this one. Holders can deposit their tokens to vote with or withdraw them at any time."
+    "warningExistingTokenTooltip": "Fractal will deploy a new ERC-20 governance token to wrap this one. Holders can deposit their tokens to vote with or withdraw them at any time.",
+    "snapshot": "SnapShot",
```
herbig, line 88: the designs had `SnapShot`, but we should lowercase the `shot` for all these strings since that's the proper styling

src/i18n/locales/en/proposal.json
```diff
@@ -26,6 +26,10 @@
     "stateExecutedTip": "The proposal has been executed.",
     "stateFailed": "Failed",
     "stateFailedTip": "The proposal failed to achieve quorum or did not get enough votes to pass.",
+    "stateClosed": "Closed",
+    "stateClosedTip": "The snapshot proposal voting has been closed",
+    "statePending": "Pending",
+    "statePendingTip": "The snapshot proposal voting has not begun yet",
```
herbig, line 32: but uppercase Snap for these two usages

herbig (changes requested):
Looks gooood, a few minor things to check out and we're good to go

src/components/DaoCreator/formComponents/EstablishEssentials.tsx
tbwebb22: Cool, removed this

src/components/Proposals/index.tsx
tbwebb22: Ah yes, just added

src/hooks/schemas/common/useValidationAddress.tsx
tbwebb22: Cool, just pushed up fix for this

src/hooks/schemas/common/useValidationAddress.tsx
tbwebb22: Ah good catch, pushed this up

src/i18n/locales/en/proposal.json
tbwebb22: done

herbig (approved):
⚡️⚡️⚡️

PR #1179, opened 2023-05-25
Remove ipfs-http-client

mudrila:
## Description
This PR removes `ipfs-http-client` and implements basic client over `axios` and HTTP API that Infura provides.

## Notes
Eventually, the idea is to fix 500 error that happens when visiting DAO dashboard page by direct link


app/layout.tsx
```diff
@@ -16,15 +16,15 @@ import Layout from '../src/components/ui/page/Layout';
...
 import { NetworkConfigProvider } from '../src/providers/NetworkConfig/NetworkConfigProvider';
 import { chains, wagmiClient } from '../src/providers/NetworkConfig/rainbow-kit.config';
 
 export default function RootLayout({ children }: { children: ReactNode }) {
-  useEffect(() => {
-    initErrorLogging();
-  }, []);
+  // useEffect(() => {
+  //   initErrorLogging();
+  // }, []);
```
herbig, line 27: we should turn Sentry back on while we have it.

Although I'm not sure if we even need Sentry at this point honestly...

PR #1180, opened 2023-05-26
500 Error fix testing

herbig:
## Description

Reverts next library version to what we had prior to https://github.com/decent-dao/fractal-interface/pull/1171/files, which is where it seems the 500 error was introduced.


herbig:
Have reverted everything manually and nothing seems to work.. closing this.

PR #1181, opened 2023-05-26
Revert 1171

herbig:


herbig:
this reverted the version in package.json, but package lock is out of sync with the proper version, so it didn't work properly..

PR #1182, opened 2023-05-30
Downgrade next version to 13.3.0

herbig:
## Description

Downgrades the next version to 13.3.0, which does not seem to have the 500 error we're seeing on 13.4.3.


herbig:
this didn't work either, and neither does a fresh package lock generation, closing this..

PR #1204, opened 2023-06-06
Bugfix variety pack child DAOs

Da-Colon:
## Description
This PR covers a few subDAO related loading issues. 

- 2 Additional Requests were found in `useClawback` and in the `DAOInfoCard` that were calling additional api requests and re-renders. 

- I found that the conditionals for the loading of the `FreezeGuard` had a flaw that was causing the guard not to load.

## Notes
After fixing these, I went to test the issue that Tom had saw and haven't been able to reproduce. 

## Testing
Should see smooth loading between DAOs. If you have it I would test all combos of loading and ensuring that all data loads and unloads properly. P-Multi -> C-Multi, C-Multi -> P-Token, etc....


Da-Colon:
@herbig Yeaaaah buddy I got it fixed. `useFractalNode`. sorry this wasn't caught before. the reintroduction of the subGraph query as useQuery hook getting fired on mount was causing child info to be pushed into the App Store.

I've extracted the code for `loadDAO` into its own hook and remove the `onMount` stuff from `useFractalNode` as its only used in the `useDAOController` now.

herbig (approved):
all good, fixes a lot of weirdness, thanks @Da-Colon

PR #1205, opened 2023-06-08
Deploy Azorius

mudrila:
## Description
This PR implements Azorius deployment for existing Multisig safe - either one created through Fractal or any random Safe.

## Testing
1. Modify governance of existing Multisig DAO or just Safe where you're signer
2. Make sure proposal created properly
3. Execute proposal
4. Verify that DAO now is displayed as Azorius Token Voting DAO with proper gov module settings passed during "Modify Governance" form fill


herbig (changes requested):
Hey @mudrila !

I have a proposal to convert this Safe to token voting, and the transaction fails:

https://deploy-preview-1205--fractal-framework-interface-dev.netlify.app/daos/0x4F7A6f79BB743c12B35492C5cea9e3648ecAF574/proposals/0xbf8fdd4202b3529c51f8e68d65777ee726af9f5dfc7d2f686255e3ead0490e32

You should be a signer there as well, it's the old moneydao.eth

src/hooks/DAO/useDeployAzorius.ts
```diff
@@ -0,0 +1,139 @@
...
+      const daoTxBuilder = txBuilderFactory.createDaoTxBuilder();
+      const safeTx = await daoTxBuilder.buildAzoriusTx(shouldSetName, shouldSetSnapshot, false);
+
+      const proposalData: ProposalExecuteData = {
+        targets: [multiSendContract.asSigner.address],
+        values: [BigNumber.from('0')],
+        calldatas: [multiSendContract.asSigner.interface.encodeFunctionData('multiSend', [safeTx])],
+        title: t('Change governance module to Azorius', { ns: 'proposalMetadata' }),
+        description: '',
+        documentationUrl: '',
```
herbig, line 96: @mudrila do we need these when converting from multisig to token voting?  This metadata is only for creating an existing token voting proposal right?

herbig (approved):
nice, works like charm

PR #1221, opened 2023-06-21
Add missing graphql assets building

mudrila:
## Description
This PR adds missing step of building GraphQL assets during IPFS release


herbig (approved):
Oh right, we need that now, ha

PR #1227, opened 2023-06-28
clean up decoding and caching transactions

herbig:
## Description

Adds better transaction decoding for multisig proposals by calling the Safe API directly if it hasn't already been decoded.

Also handles better caching of transaction data.


src/hooks/utils/useSafeTransactions.ts
```diff
@@ -281,7 +283,13 @@ export const useSafeTransactions = () => {
...
                   ),
                 }
-              : undefined;
+              : {
+                  decodedTransactions: await decode(
+                    multiSigTransaction.value,
+                    multiSigTransaction.to,
+                    multiSigTransaction.data
+                  ),
+                };
```
herbig, line 292: if the Safe API doesn't have decoded data, we'll try the decode API directly.

This causes the Safe's proposals list API to then have the data properly in the future, as we've called to decode it.

src/hooks/DAO/loaders/governance/useAzoriusProposals.ts
```diff
@@ -25,7 +25,18 @@ export const useAzoriusProposals = () => {
...
+    async (transactions: MetaTransaction[]) => {
+      const decodedTransactions = await Promise.all(
+        transactions.map(async tx => {
+          return decode(tx.value.toString(), tx.to, tx.data);
+        })
+      );
+      return decodedTransactions.flat();
+    },
+    [decode]
+  );
```
herbig, line 39: after refactoring `useDecodeTransaction` to be more generic to any transaction, this was all that was left so I just moved it into here directly

`useParseSafeTxs` was unused, so I removed it

PR #1235, opened 2023-07-05
Bugfix | submitting proposal not working on token DAOs

Da-Colon:
## Description

So I noticed that the logic inside of `useSubmitProposal` was incorrect and wasn't correctly using the `globalAzoriusContract` (current DAO). To keep a long story short the only place that should be passing `safeAddress` into the function call `submitProposal` should be `useClawback` as this allows creating the proposal on the parent DAO.

I've fixed this logic and removed the prop on all the other spots. 

## Testing
I've tested (so Far):
Parent Multisig -> Create SubDAO
Child DAO -> SendAssets
Parent Token -> SendAssets, Create SubDAO, create Proposal Template
Child Token -> create Proposal Template


herbig (approved):
Looks good, tested on azorius template creation and running a transaction from one

PR #1244, opened 2023-07-07
Strings cleanup

herbig:
## Description

Remove unused strings and a few modifications of existing ones.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fractal-interface",
-  "version": "0.1.3",
+  "version": "0.1.4",
```
herbig, line 3: upped the version since I'm going to merge into production shortly

src/components/pages/DaoSettings/components/Metadata/index.tsx
```diff
@@ -101,7 +101,7 @@ export default function MetadataContainer() {
 
   return (
     <SettingsSection
-      contentTitle={t('daoMetadataTitle')}
+      contentTitle={''}
       contentHeader={
```
herbig, line 105: the logic of SettingsSection is `contentHeader | contentTitle`, so this was never used here

Issue #1371, opened 2024-02-21
DAO Proposal state re-gathered during proposal creation


adamgall:
### Overview

From @Da-Colon:

> The current workflow for `submitProposal` has a conditional path that was setup to I beleive to allow for `createProposal` to be called from another DAO by just passing in the safeAddress Param (the conditional). For some reason, The Create Proposal Page is using this path instead of using state. So the hanging is caused by all the DAO data trying to be regathered again.
>
> Removing This value should just use state (currently loaded DAO) to execute the create proposal, but I paused because I added this 10 months ago, knowing this.

### Additional Info

Seeing this issue on Sepolia.

Connect to the App using a non-Metamask wallet. I use Frame, and there are two different RPC providers set up in Frame. One is their default one called "Pylon", the other is a public node: https://ethereum-sepolia-rpc.publicnode.com

Attempt to create a proposal on a Fractal DAO. Upon clicking the "Create Proposal" button, nothing happens visually. I see one of two errors in the console, depending on which RPC node my Frame wallet is connected to:

Pylon:
```
{
    "message": "Method eth_getLogs is not supported",
    "code": -32601
}
```

PublicNode:
```
{
    "message": "exceed maximum block range: 50000",
    "code": -32701
}
```

### Version

Dev (app.dev.fractalframework.xyz/)

PR #1373, opened 2024-02-23
Voting Strategies Upgrade + Master Copy Caching

Da-Colon:
## Description

This PR strives to remove some of the loading errors from the 429 request error seen in the app. 

First I saw an opportunity to save on the strain of Event filtering by replacing the queryFilter for `EnabledStrategies()` with  just a call on the contract `getStrategies()` to retrieve the voting contract address (as its labeled in the code). 

The second part of the fix is to be considered a bandaid. After discussing with @adamgall There is a better and more efficient pattern for retrieving the master copy instead of relying on the event `queryFilter`. This will be detailed later in a card in `Fractal Maintenance` Project Board. 

For now I'm using the `useLocalStoage` hook that we have in place to store `master_copy_of`. to save `proxyAddress: masterCopyAddress`. So once requested and found we now cache these values so that we don't have to relook up these values don't change.

## Notes
Additional to Adam Suggested fix. I've noticed that current patterns in `useLocalStorage` hasn't scaled well and could use an relook. it works for now so not a priority for V1 at the moment.

### DIFF
--- a/src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
+++ b/src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
@@ -17,6 +17,7 @@ import {
 import useSubmitProposal from '../../../../hooks/DAO/proposal/useSubmitProposal';
 import useUserERC721VotingTokens from '../../../../hooks/DAO/proposal/useUserERC721VotingTokens';
 import useClawBack from '../../../../hooks/DAO/useClawBack';
+import { CacheKeys } from '../../../../hooks/utils/cache/cacheDefaults';
 import { useLocalStorage } from '../../../../hooks/utils/cache/useLocalStorage';
 import useBlockTimestamp from '../../../../hooks/utils/useBlockTimestamp';
 import { useFractal } from '../../../../providers/App/AppProvider';
@@ -116,7 +117,9 @@ export function ManageDAOMenu({
                 0
               )
             )[1];
-            const cachedMasterCopyAddress = getValue('master_copy_of_' + votingContractAddress);
+            const cachedMasterCopyAddress = getValue(
+              CacheKeys.MASTER_COPY_PREFIX + votingContractAddress
+            );
             let votingContractMasterCopyAddress = cachedMasterCopyAddress;
             if (!votingContractMasterCopyAddress) {
               const rpc = getEventRPC<ModuleProxyFactory>(zodiacModuleProxyFactoryContract);
@@ -126,7 +129,10 @@ export function ManageDAOMenu({
                 .then(proxiesCreated => {
                   return proxiesCreated[0].args.masterCopy;
                 });
-              setValue('master_copy_of_' + votingContractAddress, votingContractMasterCopyAddress);
+              setValue(
+                CacheKeys.MASTER_COPY_PREFIX + votingContractAddress,
+                votingContractMasterCopyAddress
+              );
             }
 
             if (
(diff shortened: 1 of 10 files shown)


src/hooks/DAO/loaders/useGovernanceContracts.ts
```diff
@@ -59,38 +60,38 @@ export const useGovernanceContracts = () => {
...
+            0
+          )
+        )[1];
 
         if (!votingContractMasterCopyAddress) {
-          const rpc = getEventRPC<ModuleProxyFactory>(zodiacModuleProxyFactoryContract);
-          const filter = rpc.filters.ModuleProxyCreation(votingContractAddress, null);
-          votingContractMasterCopyAddress = await rpc.queryFilter(filter).then(proxiesCreated => {
-            return proxiesCreated[0].args.masterCopy;
-          });
```
Da-Colon, line 93: Here I removed the previous way we were caching this could cause problems down the road as it was caching values that may change. updated to only cache proxy/master values

src/utils/azorius.ts
```diff
@@ -104,7 +104,7 @@ export const getProposalVotes = async (
       ...rest,
       voter,
       choice: VOTE_CHOICES[voteType],
-    };
+    } as ProposalVote; // This bypasses the type check, but it's fine
```
Da-Colon, line 107: I was getting an off and on error in my IDE. But for some reason it was not causing a problem during build. For this function this fix should be fine since the typing is casted when used

tomstuart123:
I believe I am to test here the issue of 'create a proposal button not clicking' + 'add a sub-safe button not appearing'. 

If correct, I've tried on this PR with WalletConnect<>Safe and then Metamask on Brave. 

Both could do both which is good news Saying that, I'm remembering this bug was really hard to replicate and appeared only in edge cases. 

Did we have a browser or wallet setup that always made it rear its head so I can 100% confirm?

tomstuart123 (approved):

src/hooks/DAO/loaders/useGovernanceContracts.ts
adamgall: Is there a concern of browsers out there in the wild that currently have these potentially variable values cached?

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
```diff
@@ -107,18 +109,31 @@ export function ManageDAOMenu({
...
+              votingContractMasterCopyAddress = await rpc
+                .queryFilter(filter)
+                .then(proxiesCreated => {
+                  return proxiesCreated[0].args.masterCopy;
+                });
+              setValue(
+                CacheKeys.MASTER_COPY_PREFIX + votingContractAddress,
+                votingContractMasterCopyAddress
+              );
+            }
```
adamgall, line 136: There's some code that looks similar to this 2 or 3 times total in this PR. Can/should it be extracted into it's own function to DRY up a bit?

adamgall (commented):
Left a couple of comments.

Da-Colon:
@adamgall
| Is there a concern of browsers out there in the wild that currently have these potentially variable values cached?

No Concern. I can make it clear this key is no longer used, I doubt it will be anyway. between the v2 and the refactor needed for localstorage that needs to happen anyway for v1.

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
Da-Colon: 100% agreed. I can do that pretty quickly

adamgall:
I have confirmed that clicking that "Create proposal" button for me now works as expected, when connected with Frame which is using my local RPC node, as documented in #1371

adamgall (approved):

PR #1374, opened 2024-02-23
Fix Multiple Re-render caused by `useEffect`

Da-Colon:
## Description

Fixed continuous re-rendering issue on Hierarchy Page

This patch addresses a performance issue on the Hierarchy Page observed while navigating between DAOs. A persistent error was logged in the console, and the performance monitor in the development tools never stabilized. Additionally, navigating to a new DAO from this page was hindered due to this problem.

The issue was traced back to a `useEffect` hook that was continuously re-executing after component mounting, causing unnecessary re-renders. To resolve this, a `useRef` hook has been added to prevent the `useEffect` from re-executing, thus improving the page's performance and navigation stability.

## Notes
I separated this fix from #1373 so that it can be merged separately, But this should be merged first.

### DIFF
--- a/src/components/pages/DaoHierarchy/DaoNode.tsx
+++ b/src/components/pages/DaoHierarchy/DaoNode.tsx
@@ -1,6 +1,6 @@
 import { Box } from '@chakra-ui/react';
 import { utils } from 'ethers';
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useLoadDAONode } from '../../../hooks/DAO/loaders/useLoadDAONode';
 import { useLoadDAOData } from '../../../hooks/DAO/useDAOData';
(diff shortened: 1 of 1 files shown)


tomstuart123:
Just played with jumping between DAOs/subDAOs. It seems very smooth from a user standpoint. 

When opening the dev tools I mostly don't see any errors but I did notice, that this one would occasionally arise. 

Screenshot 2024-02-26 at 10 42 12

Was this the error you were looking to avoid?

tomstuart123 (approved):

adamgall (approved):
These feels kinda hacky, but it's better than nothing.

@Da-Colon do you know the underlying reason why this `useEffect` keeps firing? Which piece of state inside of it changes which triggers it to re-run?

PR #1375, opened 2024-02-26
Align deployment of Safe to be of same version as Safe UI-deployed and having fallback handler attached

### LINKED ISSUE #1361: Fractal UI deploys out-of-spec Safes
- Upgrade the actual Safe contract version from 1.3.0 to 1.3.0+L2.
- Make sure the Safe contract is deployed with a CompatibilityFallbackHandler contract attached to it.

mudrila:
## Description
This PR swaps all the `GnosisSafe` usages to `GnosisSafeL2` and adds `fallbackHandler` to the create Safe flow, right into `setup` function call.

## Issue

Closes #1361 

## Testing
1. Deploy brand new Safe via Fractal
2. Create new Safe via Safe UI, go to Safe's Settings page
3. Go to Safe UI and find Fractal-created Safe, go to Safe's Settings page
4. Compare Safe created via Fractal and Safe created via Safe UI - their version should be exactly the same (`1.3.0+L2`) and attached Fallback Handler (could be found under "Modules" tab) also should be exactly the same. See screenshots below for references on how it should look like

Знімок екрана 2024-02-26 о 19 06 43
Знімок екрана 2024-02-26 о 19 43 23

### DIFF
--- a/src/providers/NetworkConfig/networks/mainnet.ts
+++ b/src/providers/NetworkConfig/networks/mainnet.ts
@@ -12,9 +12,10 @@ import MultisigFreezeVoting from '@fractal-framework/fractal-contracts/deploymen
 import VotesERC20 from '@fractal-framework/fractal-contracts/deployments/mainnet/VotesERC20.json';
 import VotesERC20Wrapper from '@fractal-framework/fractal-contracts/deployments/mainnet/VotesERC20Wrapper.json';
 import {
-  getSafeSingletonDeployment,
   getProxyFactoryDeployment,
   getMultiSendCallOnlyDeployment,
+  getSafeL2SingletonDeployment,
+  getCompatibilityFallbackHandlerDeployment,
 } from '@safe-global/safe-deployments';
 import { mainnet } from 'wagmi/chains';
 import { GovernanceType } from '../../../types';
@@ -47,7 +48,11 @@ export const mainnetConfig: NetworkConfig = {
     erc20FreezeVotingMasterCopy: ERC20FreezeVoting.address,
     erc721FreezeVotingMasterCopy: '', // TODO - Add actual address once contract is deployed on mainnet
     multisigFreezeGuardMasterCopy: MultisigFreezeGuard.address,
-    safe: getSafeSingletonDeployment({ version: SAFE_VERSION, network: CHAIN_ID.toString() })
+    fallbackHandler: getCompatibilityFallbackHandlerDeployment({
+      version: SAFE_VERSION,
+      network: CHAIN_ID.toString(),
+    })?.defaultAddress!,
+    safe: getSafeL2SingletonDeployment({ version: SAFE_VERSION, network: CHAIN_ID.toString() })
       ?.defaultAddress!,
     safeFactory: getProxyFactoryDeployment({
       version: SAFE_VERSION,
--- a/src/providers/NetworkConfig/networks/mainnet.ts
+++ b/src/providers/NetworkConfig/networks/mainnet.ts
@@ -51,19 +51,19 @@ export const mainnetConfig: NetworkConfig = {
     fallbackHandler: getCompatibilityFallbackHandlerDeployment({
       version: SAFE_VERSION,
       network: CHAIN_ID.toString(),
-    })?.defaultAddress!,
+    })?.networkAddresses[CHAIN_ID.toString()]!,
     safe: getSafeL2SingletonDeployment({ version: SAFE_VERSION, network: CHAIN_ID.toString() })
-      ?.defaultAddress!,
+      ?.networkAddresses[CHAIN_ID.toString()]!,
     safeFactory: getProxyFactoryDeployment({
       version: SAFE_VERSION,
       network: CHAIN_ID.toString(),
-    })?.defaultAddress!,
+    })?.networkAddresses[CHAIN_ID.toString()]!,
     zodiacModuleProxyFactory: ModuleProxyFactory.address,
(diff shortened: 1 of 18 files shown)


src/providers/NetworkConfig/networks/mainnet.ts
```diff
@@ -47,7 +48,11 @@ export const mainnetConfig: NetworkConfig = {
     erc20FreezeVotingMasterCopy: ERC20FreezeVoting.address,
     erc721FreezeVotingMasterCopy: '', // TODO - Add actual address once contract is deployed on mainnet
     multisigFreezeGuardMasterCopy: MultisigFreezeGuard.address,
-    safe: getSafeSingletonDeployment({ version: SAFE_VERSION, network: CHAIN_ID.toString() })
+    fallbackHandler: getCompatibilityFallbackHandlerDeployment({
+      version: SAFE_VERSION,
+      network: CHAIN_ID.toString(),
+    })?.defaultAddress!,
```
adamgall, line 54: I don't think we should be using the `defaultAddress` here... that's why you're seeing the "wrong address" for Sepolia ("11155111"). See: https://github.com/safe-global/safe-deployments/blob/main/src/assets/v1.3.0/compatibility_fallback_handler.json

I think you need to dig into the `networkAddresses` object to pull out the proper address based on the network id.

Da-Colon (approved):
Surprised we still have to rely on the typechain files for these. But tested and it indeed works. I see that the Contract+L2 is correctly displayed in the settings and the CompatibilityFallbackHandler has been set up as a module.

src/providers/NetworkConfig/networks/mainnet.ts
```diff
@@ -47,7 +48,11 @@ export const mainnetConfig: NetworkConfig = {
...
+    safe: getSafeL2SingletonDeployment({ version: SAFE_VERSION, network: CHAIN_ID.toString() })
       ?.defaultAddress!,
```
adamgall, line 56: Same issue here. Don't rely on this `defaultAddress`, but dig into the `networkAddresses` object to find the correct address.

src/providers/NetworkConfig/networks/polygon.ts
```diff
@@ -47,7 +48,11 @@ export const polygonConfig: NetworkConfig = {
     erc20FreezeVotingMasterCopy: ERC20FreezeVoting.address,
     erc721FreezeVotingMasterCopy: '',
     multisigFreezeGuardMasterCopy: MultisigFreezeGuard.address,
-    safe: getSafeSingletonDeployment({ version: SAFE_VERSION, network: CHAIN_ID.toString() })
+    fallbackHandler: getCompatibilityFallbackHandlerDeployment({
+      version: SAFE_VERSION,
+      network: CHAIN_ID.toString(),
+    })?.networkAddresses[CHAIN_ID.toString()]!,
```
adamgall, line 54: Correct!

src/providers/NetworkConfig/networks/polygon.ts
```diff
@@ -47,7 +48,11 @@ export const polygonConfig: NetworkConfig = {
...
+    safe: getSafeL2SingletonDeployment({ version: SAFE_VERSION, network: CHAIN_ID.toString() })
       ?.defaultAddress!,
```
adamgall, line 56: Fix please!

src/providers/NetworkConfig/networks/sepolia.ts
```diff
@@ -48,7 +49,11 @@ export const sepoliaConfig: NetworkConfig = {
     erc20FreezeVotingMasterCopy: ERC20FreezeVoting.address,
     erc721FreezeVotingMasterCopy: ERC721FreezeVoting.address,
     multisigFreezeGuardMasterCopy: MultisigFreezeGuard.address,
-    safe: getSafeSingletonDeployment({ version: SAFE_VERSION, network: CHAIN_ID.toString() })
+    fallbackHandler: getCompatibilityFallbackHandlerDeployment({
+      version: SAFE_VERSION,
+      network: CHAIN_ID.toString(),
+    })?.networkAddresses[CHAIN_ID.toString()]!,
```
adamgall, line 55: Correct!

src/providers/NetworkConfig/networks/sepolia.ts
```diff
@@ -48,7 +49,11 @@ export const sepoliaConfig: NetworkConfig = {
...
+    safe: getSafeL2SingletonDeployment({ version: SAFE_VERSION, network: CHAIN_ID.toString() })
       ?.defaultAddress!,
```
adamgall, line 57: Fix please!

adamgall (changes requested):
Whenever using the `safe-deployments` package to `get` singleton addresses, the filter simply confirms that the contract has been deployed on the given network.

It's still up to the caller to dig into the `networkAddresses` object of the result to find the _official_ address for a given chain, do not rely on the `defaultAddress`!

This applies to the new `CompatibilityFallbackHandler` code (which seems good for Sepolia and Polygon, but might as well do it for Mainnet too), as well as the actual Master Safe Singleton deployment address as well (for ALL THREE networks).

adamgall:
> Surprised we still have to rely on the typechain files for these. But tested and it indeed works. I see that the Contract+L2 is correctly displayed in the settings and the CompatibilityFallbackHandler has been set up as a module.

Pretty lucky that we had those `GnosisSafeL2` types in that `usul` folder, huh! It looks like the only reason we had those types in the project, before, this PR, was to get the Multisend typing.

adamgall (approved):
LGTM!

PR #1377, opened 2024-02-26
Release 0.1.6

adamgall:
I'm attempting to follow the "git flow" workflow here.

https://nvie.com/posts/a-successful-git-branching-model/#creating-a-release-branch

I've branched off of `develop`, made a change (bumped the version number in `package.json`, and now attempting to get this branch merged into `main`.

After this is merged, I'll tag the commit.

Then we also need to get this branched merged into `develop`, which should happen after it's merged into `main`.

I'm not sure if there should be another PR for that, or if I just do that locally and force push through the Github restrictions.

Open for comments on this process fellas!

PR #1379, opened 2024-02-27
Bumped version number to 0.1.6

adamgall:
Finish the git flow, started via release process in #1377


Da-Colon (approved):

adamgall:
> You probably also need to run npm install - this should update package-lock as well

Oh snap you're right! I didn't see this comment before I merged. Oops, oh well, we'll get it in the next release branch.

PR #1382, opened 2024-02-28
Show token prices

### LINKED ISSUE #1362: Asset Prices not showing - Design and Build Pricing Service
https://ethereum.stackexchange.com/questions/156437/safe-transaction-api-balance-returns-0-for-fiatconversion-for-all-tokens. 

Apparently they had to shut it down as it was getting hard to keep up. They suggest integrating with coingecko until a new service is being built out (currently don't see it).

---

After discussion with the team, this won't be solvable with a pure client-side CoinGecko API integration, unfortunately.

We need to build a backend price supplier API.

Here are some requirements:

- It should include IP-based rate limiting.
- It should include CORS-based limiting, same origin.
- It uses CoinGecko's Free Tier API key, which has a limit of 30 requests per minute.
- Service is stateless with regards to Safes, it doesn't know anything about Safe addresses.
- Service accepts a list of token addresses (and/or ETH), and responds back with USD value for each token address (and/or ETH).

I see the hard part being satisfying these three things at the same time. As long as number 3 is a _should_, not a _must_, we'll be gucci:

1. Service _must_ always respond to a request in a timely manner (no slower than however much longer a CoinGecko request takes, plus some time for our service to build a response).
2. Service _must_, globally, never hit the CoinGecko API more than 30 times in a 1 minute window.
3. Service _should not_ return stale price data ("staleness" to be tweaked later, lets call it 30 mins for now).

mudrila:
## Description

This PR implements fetching coin prices from CoinGecko.
The implementation is as dumb as you could only imagine. Probably we could save some requests to putting returned value of `useFormatCoins` into global context - but that might lead us into unnecessary implementation over-engineering (given the fact that we could be migrating to the Safe API Kit once they'll release pricing service anyway).

## Issue
Closes #1362 

## Testing
1. Visit some mainnet DAO with bunch of assets
2. Verify that prices for token and total USD is calculated properly


Da-Colon:
Not quite working. Went to Decent DAO on mainnet and it is showing only $9,708.66

adamgall:
Seeing that this implementation contained a CoinGecko API key surprised me, frankly.

I suppose it's been a minute (~2 years) since I last implemented a CoinGecko powered pricing mechanism in a frontend app. Back then, the API was _keyless_ -- totally open without any need for an API key, and rate limiting was simply enforced via IP address. Even still, those limits were quite generous. I believe they allowed something like up to 30 requests per minute. As long as I wrote some smart code to minimize API requests, there was no way any one user was going to hit those limits unless they refreshed the page over and over and over.

I see that this isn't the case any longer, and I wonder if this implementation as it currently stands will work for us.

Here are the problems I see:

- Open API full deprecation
  - According to the [official CoinGecko Reddit account](https://www.reddit.com/r/coingecko/comments/16bc259/comment/kok615p/?utm_source=share&utm_medium=web2x&context=3), the open keyless API should be fully deprecated _as of right now_.
  - In practice this isn't the case as far as I can tell. See: https://api.coingecko.com/api/v3/simple/price?ids=ethereum&vs_currencies=usd
  - There are still IP-based rate limits with this open API. Hit refresh a few times on that above API call and you'll hit it pretty quickly. It's certainly not the "30 per minute" that it used to be.
  - I think it'd be foolish to rely on this open API (without needing a key) being available indefinitely.

- Free API key limits
  - They have a free API tier, which includes a key that [rate limits at 30 calls per minute](https://docs.coingecko.com/v3.0.1/reference/common-errors-rate-limit#rate-limit).
  - This sounds good initially, but important to note that the limits are enforced _at the API key, not the IP address_. So if we bundle this API key in the app, and all Fractal users are relying on this same API key, there's no way in heck this will work for us.

- Paid tier
  - So now we're looking at [CoinGecko's paid tiers](https://www.coingecko.com/en/api/pricing). The cheapest tier still has a rate limit of 500 calls per minute, and I don't think that would hold up either during periods of a couple of DAOs with a couple of proposals active at the same time.
  - If there are a few dozen users hitting the site at the same time, I could quickly see those per-minute rates being hit, because each site load is going to hit a handful of endpoints.
  - Even still, it's just a bad pattern to include this API key in the frontend codebase. Anyone can sniff it and start using it. CoinGecko doesn't allow URL Referral enforcements either.
  - CoinGecko strongly recommends we store this key on a backend server, and proxy requests through it. ([See the Notes section on this page](https://docs.coingecko.com/reference/authentication)).

So where does that leave us? I'm not sure.

- Are there any other open pricing APIs we could use?
- If not
  - Do we simply start paying for a high-volume CoinGecko API tier, stick that API key in our frontend bundle, and hope for the best?
  - Or do we build a backend service (probably a Netlify function cause why not that seems easy enough) to proxy calls through, and importantly _build some persistence into that service_ so we don't need to hammer CoinGecko's API.

adamgall (changes requested):

netlify/functions/tokensPrices.mts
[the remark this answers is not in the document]
adamgall: Ok I learned what `.mts` is 👍 
https://docs.netlify.com/functions/get-started/?fn-language=ts#create-function-file

netlify/functions/tokenPrices.mts
```diff
@@ -0,0 +1,94 @@
...
+  if (!tokensString) {
+    return Response.json({ error: 'Tokens to request were not provided' });
+  }
+  const tokens = tokensString.split(',');
+  try {
+    const now = Date.now();
+    const cachedPrices = await Promise.all(
+      tokens.map(
+        tokenAddress =>
+          store.getWithMetadata(tokenAddress, {
```
adamgall, line 26: Why `getWithMetadata`, and not simply `get`?

netlify/functions/tokenPrices.mts
```diff
@@ -0,0 +1,94 @@
...
+};
+
+export default async function getTokenPrices(request: Request) {
+  const store = getStore('token-prices');
+  const tokensString = new URL(request.url).searchParams.get('tokens');
+
+  if (!tokensString) {
+    return Response.json({ error: 'Tokens to request were not provided' });
+  }
+  const tokens = tokensString.split(',');
```
adamgall, line 20: We should do some validation that these addresses are valid (using ethers.js or viem `isAddress` is fine), before calling into our store and/or calling out to CoinGecko.

adamgall (changes requested):
What's the easiest way for me to test this code out locally?

netlify/functions/tokenPrices.mts
adamgall: Oh I get it, the expiration value is stored in `metadata`.

adamgall:
I've just pushed up a pretty major refactor. Some things to note:

- Need two new local environment variables:
  - `TOKEN_PRICE_VALID_CACHE_MINUTES="5" # or however long you want`
  - `TOKEN_PRICE_INVALID_CACHE_MINUTES="10" # or however long you want`
- Heavily optimized to not bombard CoinGecko
  - Caches valid prices in blob storage (see `TOKEN_PRICE_VALID_CACHE_MINUTES`)
  - Don't throw error when query param is not a valid address, just skip it.
  - Don't send "ethereum" string as a token to CG token API endpoint.
  - Return ethereum and token price in response immediately if all data can be found in cache.
  - Add entries to blob store (with long expiration, see `TOKEN_PRICE_INVALID_CACHE_MINUTES`) for token addresses that CoinGecko doesn't track, or doesn't have USD price for, to avoid constantly asking about these tokens
  - Netlify function will always return response consisting of requested token addresses, even if all of the prices are "0" (in worst case)
- Frontend Treasury page use some more optimizing for when CoinGecko fails / hits rate limits. The backend currently works well and as expected (by returning cached data, even if it might be stale), but frontend throws an error toast. I think maybe it shouldn't throw an error toast.

tomstuart123 (approved):

adamgall (changes requested):
One more thing I just thought of -- we should namespace these blob keys to include the network.

netlify/functions/tokenPrices.mts
```diff
@@ -0,0 +1,279 @@
...
+  if (!tokensStringParam) {
+    return Response.json({ error: 'Tokens missing from request' }, { status: 400 });
+  }
+
+  const networkParam = requestSearchParams.get('network') as SupportedNetworks;
+  if (!networkParam || !SUPPORTED_NETWORKS.includes(networkParam)) {
+    return Response.json({ error: 'Requested network is not supported' }, { status: 400 });
+  }
+
+  const store = getStore(`${networkParam}-token-prices`);
```
adamgall, line 203: I don't think we should be introducing multiple stores here. Please just keep one single `token-prices` store.

Each KV pair should have that chainspacing on it: `${networkId}/${tokenAddress}`

adamgall (changes requested):

src/providers/App/hooks/usePriceAPI.ts
```diff
@@ -0,0 +1,52 @@
...
+        }
+        if (tokensAddresses.length > 0) {
+          const pricesResponse = await fetch(
+            `/.netlify/functions/tokenPrices?tokens=${tokensAddresses.join(',')}&network=ethereum`
+          );
+
+          const pricesResponseBody = await pricesResponse.json();
+          if (pricesResponseBody.error) {
+            // We don't need to log error here as it is supposed to be logged through Netlify function anyway
+            toast.warning(t('tokenPriceFetchingError'));
```
adamgall, line 35: Let's not show a toast error _if we receive a response from the API_.

We _should_ show a toast error _if the API service itself is down_. Would that happen down between lines 45-46?

adamgall (changes requested):

netlify/functions/tokenPrices.mts
```diff
@@ -62,7 +61,7 @@ async function splitData(
 
   // Let's pull out all of the expired addresses from our cache.
   const expiredCachedTokenAddresses = cachedTokenPrices
-    .filter(tokenPrice => tokenPrice.metadata.expiration < config.now)
+    .filter(tokenPrice => tokenPrice.metadata.fetched > config.now + config.cacheTime)
```
adamgall, line 64: To get expired addresses, wouldn't we want it to be "fetched + cache < now"?

adamgall (commented):
Just one thing that seems off to me.

netlify/functions/tokenPrices.mts
```diff
@@ -0,0 +1,268 @@
...
+  // Filter out the null values, leaving us with an array of
+  // TokenPricesWithMetadata. All of these TokenPrices will be either
+  // expired or unexpired.
+  const cachedTokenPrices = possibleCachedTokenPrices.filter(
+    (possible): possible is TokenPriceWithMetadata => possible !== null
+  );
+
+  // Let's pull out all of the expired addresses from our cache.
+  const expiredCachedTokenAddresses = cachedTokenPrices
+    .filter(tokenPrice => tokenPrice.metadata.fetched + config.cacheTime < config.now)
```
adamgall, line 64: Perhaps we should be converting this `cacheTime` from "minutes" to "seconds" heh

adamgall (changes requested):
I'm testing this locally, and it's not working. I have a value of "5" (minutes) set in my env var, but putting in a console.log right before the fetch to coingecko shows that it's making that request every time the frontend calls the API (every ~20 seconds).

netlify/functions/tokenPrices.mts
adamgall: `config.now` gives seconds (https://github.com/decent-dao/fractal-interface/pull/1382/files#diff-284b3930e569c4259e85b685d36c139226ef2a210eaeb25f614b752fd1b2e910R197)

The env var is called `TOKEN_PRICE_CACHE_INTERVAL_MINUTES`, which specifies that it should be `MINUTES`. Minutes are a lot easier to reason about that milliseconds, when setting configs in an env var.

I think the cleaner solution is to convert the value to seconds when setting the config property here: https://github.com/decent-dao/fractal-interface/pull/1382/files#diff-284b3930e569c4259e85b685d36c139226ef2a210eaeb25f614b752fd1b2e910R198
```ts
const cacheTime = parseInt(process.env.TOKEN_PRICE_CACHE_INTERVAL_MINUTES) * 60
```

adamgall: @mudrila i just pushed up the fix, no worries

adamgall (approved):
Let's go!

PR #1383, opened 2024-02-28
Caching Cleanup + Timer clearing update

Da-Colon:
## Description
This took me longer than I expected. The original goal of this PR was to clean up some of the redundancy in the master copy lookup and caching. While doing this I uncovered a loading issue. I'll go through a code review and talk about the specific changes as I only have an assumption on why my changes caused this issue to pop up.

### Issue Arises
So after create `useMasterCopy` and updating all the places that use the same pattern. I started noticing that the original decrease was erased and requests increased. After several console logs through out the loading places that were loading either out of sequence or now causing re-renders that were not being revealed before as the previous RPC request held these things up.

### The Fix

- `useUpdateTimer` in `useProposals`
When loading a Azorius DAO (pre-fix) it would for a moment load as a Multisig DAO, which caused me to notice that the setMethodInterval was firing in the background. This was also happening after switching from viewing one to another via links (Hierarchy Page and parent-safe link).

- `useGovernanceContract` logic cleanup
This seemed in the end to be causing extra re-renders. I cleaned up the logic a bit and seems to be working now as expected.

## Notes
Every once in a while, if clicking between DAOs very quickly I would run into the ErrorBoundaryScreen and would need to reload but was pressing around really quickly.

## Testing
To test that everything is working as expected. I'd would main checking this PR by loading different DAO in different ways. 
- via Search
- via Hierarchy Page
- via Reload

Then just do a few reloads on the deeper pages to make sure state is still loading correctly.


app/layout.tsx
```diff
@@ -1,19 +1,12 @@
...
 import 'react-toastify/dist/ReactToastify.min.css';
 import { APP_NAME } from '../src/constants/common';
-import { initErrorLogging } from '../src/helpers/errorLogging';
 import Providers from '../src/providers/Providers';
 
 export default function RootLayout({ children }: { children: ReactNode }) {
-  useEffect(() => {
-    initErrorLogging();
-  }, []);
-
```
Da-Colon, line 16: Left over from the Providers move refactor. This was being init twice.

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
Da-Colon, line 1: Swap out for `useMasterCopy` implementation

src/hooks/DAO/loaders/useFractalGuardContracts.ts
Da-Colon, line 1: Swap out for `useMasterCopy` implementation

src/hooks/DAO/loaders/useFractalModules.ts
Da-Colon, line 1: Swap out for `useMasterCopy` implementation

src/hooks/DAO/loaders/useGovernanceContracts.ts
```diff
@@ -1,264 +1,211 @@
...
   ]);
 
   useEffect(() => {
-    if (
-      (!!node.daoAddress &&
-        node.isModulesLoaded &&
-        node.daoAddress !== currentValidAddress.current) ||
-      lastChainId !== chainId
-    ) {
-      setLastChainId(chainId);
```
Da-Colon, line 260: Wasn't really sure what setLastChainId was doing here. I cleaned up this state and seems to be working as expected

src/hooks/DAO/loaders/useGovernanceContracts.ts
```diff
@@ -1,264 +1,211 @@
...
+    if (currentValidAddress.current !== daoAddress && isModulesLoaded) {
       loadGovernanceContracts();
+      currentValidAddress.current = daoAddress;
```
Da-Colon, line 208: Moving this value outside of being set inside of the `loadGovernanceContracts()` callback all seem to help the multiple loads

src/hooks/DAO/loaders/useGovernanceContracts.ts
adamgall: Was this perhaps being used to invalidate state when chains are switched?

Da-Colon: maybe but when chain is changed the entire DAO is unloaded anyway and the "This DAO doesn't exist on [Network]" with a reload button shows up

adamgall: Word, that makes sense.

adamgall:
My testing:

- Attempting to load Azorius DAO
  - No wallet connected
    - [x] via Search
    - [x] via Reload
    - [ ] via Hierarchy Page
  - Frame wallet connected via Browser Wallet
    - [x] via Search
    - [x] via Reload
    - [ ] via Hierarchy Page
  - Safe connected via WalletConnect
    - [x] via Search
    - [x] via Reload
    - [ ] via Hierarchy Page
- Attempting to load Multisig DAO
  - No wallet connected
    - [x] via Search
    - [x] via Reload
    - [ ] via Hierarchy Page
  - Frame wallet connected via Browser Wallet
    - [x] via Search
    - [x] via Reload
    - [ ] via Hierarchy Page
  - Safe connected via WalletConnect
    - [x] via Search
    - [x] via Reload
    - [ ] via Hierarchy Page

@Da-Colon @tomstuart123 do you have any sepolia daos offhand that have some hierarchy set up that I can use for testing?

adamgall (changes requested):
I'm seeing some strange behavior when clicking in and out of the parent/child DAOs. Some state is hanging around between Safes. Sometimes the Child (Multisig) presents itself as Token Voting. Sometimes the Parent shows as having the Execution time of the Child. Talking with David now, he's on it 💪

Da-Colon:
@adamgall looks like we are good to go for another round of testing. Sorry for this misfire. The 'reload' screen issue that pops up every once in a while seems to be a problem with some state somewhere in `VoteContext`. Gonna see if I can locate it. 

This same component seems to show up alot with the 429 errors too so gonna take a closer look.

Issue #1384, opened 2024-02-28
Run ESLint/Prettier as a Github Action


adamgall:
### Overview

We should be doing all of the Linting and Prettifying as jobs that run on PRs, which must pass in order for the PR to be mergable.

### Version

Dev (app.dev.fractalframework.xyz/)

PR #1385, opened 2024-02-28
Remove husky

adamgall:
## Description

I didn't like how ESLint was running on every commit. Really slows things down.

Methinks that ESLint should be part of a Github Action job that is required to pass for PRs to be mergable (#1384). Devs should have a proper development environment set up anyway, which Lints and Prettifies on file save regardless.

## Notes

Dear Devs: be sure to check out your .git/config and grep for "husky". You'll probably want to delete the line in there that sets the (now removed) `.husky` directory as the place to look for hooks.

## Issue

n/a

## Testing

n/a

n/a


Da-Colon (approved):
cool with me

PR #1386, opened 2024-02-28
Remove Issue and PR templates

adamgall:
There was some stale information in the Issue template config (linking to the Fractal Discord, which is deprecated).

There was some duplicated information in the issue template page (two links to the security policy).

The two "types" of issues that could be opened were too restrictive (only bugs and feature requests, what about other QoL updates, or other maintenance tasks).

I think at this stage in the project, trying to template out all of the ways that people interact with this repo is a bit premature.

So, I've removed it in this PR.


Da-Colon (approved):
👍

PR #1389, opened 2024-02-28
Include a "MAINNET" identifier in the mainnet Alchemy env var

adamgall:
To bring some consistency to the environment variable names.

I've already _added_ the new environment variable in Netlify, so this should work in the preview deployments and once merged.

After this code merges into `develop`, it'll be safe to delete `NEXT_PUBLIC_ALCHEMY_API_KEY` from the dev site in Netlify.
After this code merges into `main`, it'll be safe to delete `NEXT_PUBLIC_ALCHEMY_API_KEY` from the prod site in Netlify.

As a side note, I strongly encourage all engineers on the project (@Da-Colon, @mudrila) to set up their own free-tier accounts on Etherscan, Alchemy, Infura, and WalletConnect Cloud, and use those keys in their `.env.local` files.

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,5 @@
-NEXT_PUBLIC_ALCHEMY_API_KEY=""
+# Alchemy provider API key, used on Mainnet
+NEXT_PUBLIC_ALCHEMY_MAINNET_API_KEY=""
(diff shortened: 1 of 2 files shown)


Da-Colon (approved):
🚢

PR #1397, opened 2024-03-01
Fix Nested Anchor Tags

### LINKED ISSUE #1396: Browser complaining about nested anchor tags on Proposal List page
When loading a DAO, the browser complains about nested anchor tags.

This was a bug introduced during #1380/#1381. The existing buttons on the cards are links, and so making the entire card a link made these existing button links nested.

adamgall:
Closes #1396 

The `ProposalAction` buttons, which show up in proposal list cards, no longer needs to be a link.

Testing:

On dev load up any DAO that has proposals. Notice that when the proposal list populates, there's an error in the console talking about how nested anchor tags are not allowed.

Then do the same on this branch, and notice how that error does not appear.

@mudrila I would love your eyes on this, specifically. There are three files where the `ProposalAction` component is instantiated:
- src/components/Activity/ActivityGovernance.tsx
- src/components/Proposals/AzoriusDetails/index.tsx
- src/components/Proposals/SnapshotProposalDetails/index.tsx

The first file, `ActivityGovernance.tsx` is the proposal card that shows up on the DAO Home and Proposals pages. That seems good.

But I'm not sure about these other two files. It seems like `AzoriusDetails` and `SnapshotProposalDetails` are both the actual details pages for the various types of proposals. The `ProposalAction` button, in these pages, is being used to display the various "Vote" buttons, when those are relevant? So I probably don't want to just straight up disable their links, as I'm doing in this PR?

I'm thinking that's the case, just want to get your take on it.

### DIFF
--- a/src/components/Proposals/ProposalActions/ProposalAction.tsx
+++ b/src/components/Proposals/ProposalActions/ProposalAction.tsx
@@ -1,9 +1,7 @@
 import { Button, Flex, Text } from '@chakra-ui/react';
-import Link from 'next/link';
 import { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { BACKGROUND_SEMI_TRANSPARENT } from '../../../constants/common';
-import { DAO_ROUTES } from '../../../constants/routes';
 import useSnapshotProposal from '../../../hooks/DAO/loaders/snapshot/useSnapshotProposal';
 import { useFractal } from '../../../providers/App/AppProvider';
 import { ExtendedSnapshotProposal, FractalProposal, FractalProposalState } from '../../../types';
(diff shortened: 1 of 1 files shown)


adamgall:
Thank you for the review @mudrila !

@tomstuart123 @Da-Colon when you get a chance, please :)

Given that @mudrila said this implementation looks good, I just went ahead and tested more thoroughly. The issues that I _though_ would pop up, didn't! So yeah, this seems good to go from my end as well.

The things to test:
- No console errors on the Proposals List page complaining of nested anchor tags.
- Clicking the various Vote buttons within the Proposal Details page do what they're supposed to do.

Da-Colon (approved):
Nicely done 👍

PR #1406, opened 2024-03-04
[BUGFIX] Stale WalletConnect Session

Da-Colon:
## Description
Looking into MPCVault Wallet Connections, I noticed than after disconnecting from Wallets connected via WalletConnect. I am unable to reopen the wallet connect menu and the Console shows errors related to the session key not matching and then showing `core/relayer` errors. 

After looking at the setup. It appears that what is happening is that for some reason after clicking `disconnect` that the menu closes and causes the func to end before updating local storage that the session has ended. This causes WAGMI and Walletconnect to desync. 

~Adding a small delay (`setTimeout`) after execution of disconnect fixes this issue.~

Instead of using AppState to determine connection status, as this seems to be updated quicker than wagmi's state. switching out to use `useWalletClient` to determine connection status seems to be a better fix.

## Testing
First see the error on dev:
- Connect any wallet via WalletConnect and disconnect.
- Reopen Connect Modal and Select Wallet Connect again. Should do nothing.

On deploy preview do the same thing and you should notice that you are able to click into the WalletConnect QR code again. (You may need to refresh browser after testing the error)

### DIFF
--- a/src/components/ui/menus/AccountDisplay/MenuItems.tsx
+++ b/src/components/ui/menus/AccountDisplay/MenuItems.tsx
@@ -2,17 +2,14 @@ import { MenuList } from '@chakra-ui/react';
 import { Connect, Disconnect } from '@decent-org/fractal-ui';
 import { useConnectModal } from '@rainbow-me/rainbowkit';
 import { useTranslation } from 'react-i18next';
-import { useDisconnect } from 'wagmi';
-import { useFractal } from '../../../../providers/App/AppProvider';
+import { useDisconnect, useWalletClient } from 'wagmi';
 import { MenuItemButton } from './MenuItemButton';
 import { MenuItemNetwork } from './MenuItemNetwork';
 import { MenuItemWallet } from './MenuItemWallet';
 
 export function MenuItems() {
-  const {
-    readOnly: { user },
-  } = useFractal();
   const { disconnect } = useDisconnect();
+  const { data: isConnected } = useWalletClient();
   const { openConnectModal } = useConnectModal();
   const { t } = useTranslation('menu');
   return (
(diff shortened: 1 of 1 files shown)


adamgall (commented):
I wouldn't think that this code actually delays anything from happening... the `onClick` handler is synchronous and so execution won't wait for the timer to complete.

Da-Colon:
> I wouldn't think that this code actually delays anything from happening... the `onClick` handler is synchronous and so execution won't wait for the timer to complete.

Yeah it was delaying the function from finishing. But this delay wasn't fixing the problem. I edited the PR description but.

| Instead of using AppState to determine connection status, as this seems to be updated quicker than wagmi's state. switching out to use useWalletClient to determine connection status seems to be a better fix.

adamgall:
@Da-Colon can you please open an issue for this PR, add it to the new Fractal project (yeah yeah i know, going back to just a single project board), and fill in the appropriate custom data?

Da-Colon:
@adamgall no worries can do. can you link the project to fractal-interface repo for easy access whenever you get a chance

adamgall:
@Da-Colon 

> First see the error on dev:
> 1. Connect any wallet via WalletConnect and disconnect.
> 2. Reopen Connect Modal and Select Wallet Connect again. Should do nothing.

This is not my experience on the dev site. I'm able to connect via WC, disconnect using the Fractal app, then immediately connect again with WC without a refresh.

Da-Colon:
@adamgall hmmm you are right, I'm not able to reproduce on dev...weird. it was very consistent and was easy to reproduce yesterday

Da-Colon:
Closing PR for now, currently not able to reproduce on dev. Will keep an eye out if it becomes an issue

PR #1408, opened 2024-03-04
Refactor ethers provider and signer

### LINKED ISSUE #1411: Refactor useEthersProvider and useEthersSigner
Currently, we're creating new `provider` instance while invoking `useEthersProvider`, same with `useEthersSigner`. 
We can optimize that via creating global context provider and having just 1 instance of `signer` and 1 instance of `provider`, then `useEthersSigner` and `useEthersProvider` would simply return instances from context

mudrila:
Closes #1411


adamgall:
@mudrila can you please create an issue to go along with this PR?

src/hooks/DAO/loaders/useFractalGuardContracts.ts
```diff
@@ -27,9 +27,8 @@ export const useFractalGuardContracts = ({ loadOnMount = true }: { loadOnMount?:
     action,
   } = useFractal();
 
-  const {
-    network: { chainId },
-  } = useEthersProvider();
+  const provider = useEthersProvider();
+  const chainId = provider?.network.chainId;
```
adamgall, line 31: `chainId` might be `undefined`, and later we use it to check and set keys for local storage (at least I believe that's what's happening). It might be possible for that key to include the string "undefined". Can we do an early exit if `chainId` is `undefined`?

adamgall:
@mudrila does it make sense / can this be rebased on `develop` and target `develop`?

adamgall:
Going to close this PR and re-open it as a new one, so Netlify picks it up.

PR #1417, opened 2024-03-07
Refactor/ethers provider and signer

### LINKED ISSUE #1411: Refactor useEthersProvider and useEthersSigner

adamgall:
This PR replaces #1408, which I closed because we changed the target branch to `develop` yet Netlify wasn't picking it up for a build preview.

Closes #1411


adamgall (commented):
I can't leave an official approval because I opened the PR (should have had @mudrila do it... oops), but I do approve this. I was able to get through the Kleros workflow without any issues in multiple browsers.

Da-Colon (approved):

PR #1430, opened 2024-03-11
Bug/#1402 proposal time

### LINKED ISSUE #1402: Timer in Fractal - Change in time format at the end of a 24hr period
A shutter dao member just raised this: 

Image

Kirill and I noted this in prior conversations 

Image

Image

Image

Image

Image

In short, as the clock gets to the end of a day remaining it removes the hours label from the clock (e.g. we see a transition FROM 3Days: 1hour: 1 minutes TO 3days:minutes

As this can make it seem like the proposal has hours instead of days to execute, this can be very jarring for the reader and they have to click the proposal and it's end time to confirm all is okay

adamgall:
Closes #1402 

As assumed in the issue, we were hiding time "segments" (days, hours, minutes), if they were 0, regardless of whether or not there were "larger" segments that were non-zero.

The code here now forces "zero" segments to display if there is a larger time segment that is still non-zero.

This code was extensively manually tested (by forcing "remaining times" to be specific numbers, via manipulating the code), but it'll be hard to test in realtime given the nature of the issue.

### DIFF
--- a/src/components/ui/proposal/ProposalCountdown.tsx
+++ b/src/components/ui/proposal/ProposalCountdown.tsx
@@ -64,6 +64,11 @@ export function ProposalCountdown({
   const hoursLeft = Math.floor((secondsLeft! / (60 * 60)) % 24);
   const minutesLeft = Math.floor((secondsLeft! / 60) % 60);
 
+  const showDays = daysLeft > 0;
+  const showHours = showDays || hoursLeft > 0;
+  const showMinutes = showHours || minutesLeft > 0;
+  const showSeconds = secondsLeft >= 0;
+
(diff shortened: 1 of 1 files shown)


Da-Colon (approved):
Code changes make sense to me. 

I assume, you'll be be taking a dive down component unit testing setup next? (or create an issue for it)

adamgall:
> Code changes make sense to me.
> 
> I assume, you'll be be taking a dive down component unit testing setup next? (or create an issue for it)

@Da-Colon yeah, I'm not going to rock the boat by introducing or implementing any new testing patterns or libraries _in this PR_.

PR #1432, opened 2024-03-11
Implement markdown support for proposal templates and proposal view

### LINKED ISSUE #1399: Proposal Description - Enlarge + Markdown
Notes: This work should happen in the "create proposal template" components, which will soon replace the original "create proposal" components (#1363).

---

As you can see by this request, a Shutter member requested a larger field for proposal creation. 

Image

You can see here just how small the description field is here

Image

They also requested usage of markdown to help with more customisation in creation. 

Note, they might have come to this expectation as you can see below that snapshot offers markdown 

Image

mudrila:
Closes #1399 

## Changes:
- Changes to `ProposalTemplates` to display markdown from description. See `/daos/[daoAddress]/proposal-templates/new` page
- Rendering proposals in the list with stripped markdown styling. See `/daos/[daoAddress]/proposals`(Proposals list page) and `/daos/[daoAddress]` (DAO Dashboard - Activity list rendering affected)
- Rendering Markdown in proposal details. See `/daos/[daoAddress]/proposals/[proposalId]`

We certainly will need to revisit this once we adopt `ProposalTemplates` components. Probably we want to implement description preview / editor tab, so instead of visualizing description preview in the sidebar - we should be doing that via switching markdown input "mode" from edit to preview (or somehow displaying both of them).

I also tried to incorporate https://www.npmjs.com/package/@uiw/react-markdown-editor and https://github.com/OpusCapita/react-markdown - yet, I found them quite opinionated and thought it would be better to iterate over editor experience, so pure `TextArea` is the best option as of now, from my opinion

Знімок екрана 2024-03-11 о 23 36 47
Знімок екрана 2024-03-14 о 13 08 37

## Testing:
1. Create Azorius or Multisig DAO
2. Create proposal with markdown in description. Make sure to put some very long strings to test edge cases
3. Make sure it properly appears in the proposals list, yet it should be truncated. In the list it should be as plain text without markdown styling applied, but also markdown symbols should be removed
4. Make sure markdown properly displayed on proposal details page


adamgall (commented):
@mudrila, will you please
- Leave some comments here in Github on the various code changes of this PR.
- Describe in full where in the UI is affected by these changes.
- Explain the best way to test these changes.
Thanks!

xraystyle1980:
I love the idea of supporting markdown and making the proposals more robust. I would like to spend some design time exploring how this change would effect the preview. Can we revisit this as design works through the proposal creation flow? I think making "Markdown support" a requirement of the prop creation flow redesign makes sense.

adamgall (changes requested):
# Issue 1

On the Proposal Creation Description textarea, can you please enable resizing that textarea? Is there a reason that's been disabled?

Screenshot 2024-03-14 at 2 54 57 PM

---

# Issue 2

When displaying the Proposal Description after a proposal has been created, the description is showing up twice.

_Note: I did not add a title for this proposal. This seems to be the problem._

See the description that was created onchain:

https://sepolia.etherscan.io/tx/0x4a691de95190c84e852d858718ebe9229b6a7842d6ab6b1cad17005f3eafdb48

```
"description":"# This is my Proposal\n\n## This a sub heading\n\n_these are italics_"
```

This is what it looks like in the UI, both in Proposal List and Proposal Details:

Screenshot 2024-03-14 at 3 01 21 PM

Screenshot 2024-03-14 at 2 59 38 PM

---

# Issue 3

When clicking on a Proposal Template (I do see the Markdown in the "Proposal Template Card", which is good!), the app crashes.

Da-Colon (commented):
I have no additional code questions.

@xraystyle1980 @mudrila I guess we will look into adjust the preview in a future PR.

I'll do some additional testing once @adamgall questions/comments have been resolved.

Da-Colon (approved):
Looks good

We need to update the preview section to support the new Markdown Syntax. Not alot of space to show a good description. Proposal Template has a character limit of 300 I believe. But the proposal creation has more. I'm not sure why the limit on template?

Image

Not too bad on the other end.

Image

(Unrelated) Did notice that when I click on 'show more' on the template it expanded on the main screen but also popped it up in the modal

Overall no glaring issues and works end-to-end

adamgall (changes requested):
> On the Proposal Creation Description textarea, can you please enable resizing that textarea? Is there a reason that's been disabled?

This is still unresizable. Is that intentional?

src/components/ui/modals/ProposalTemplateModal.tsx
```diff
@@ -134,12 +134,7 @@ export default function ProposalTemplateModal({
...
   return (
     <Box>
-      <LineBreakBlock
-        textStyle="text-base-mono-regular"
-        color="grayscale.100"
-        marginBottom="1.5rem"
-        text={description}
-      />
+      {description && <Markdown content={description} />}
       <Divider color="chocolate.700" />
```
adamgall, line 138: Nitpick, but we'd probably want this `<Divider>` to render when _in_ the conditional with the `<Markdown>` component, right? Please test to make sure this looks correct, either way. We shouldn't have a divider rendering with nothing above it.

src/components/ui/proposal/Markdown.tsx
```diff
@@ -67,7 +67,7 @@ export default function SnapshotProposalDescription({
         setTotalLinesError(false);
       }
     }
-  }, []);
+  }, [content]);
```
adamgall, line 70: Why add this as a dependency to a `useEffect` that doesn't rely on it?

src/router.tsx
```diff
@@ -19,7 +20,11 @@ import Treasury from './pages/daos/[daoAddress]/treasury';
 export const router = createBrowserRouter([
   {
     path: '/',
-    element: <Layout />,
+    element: (
+      <ModalProvider>
+        <Layout />
+      </ModalProvider>
+    ),
```
adamgall, line 27: What was the reason for moving this out of `Providers` and into the `Router`?

src/components/Activity/ActivityDescriptionGovernance.tsx
```diff
@@ -60,7 +60,7 @@ export function ProposalTitle({ activity }: { activity: Activity }) {
   return (
     <>
       <Text>{formatId((activity as GovernanceActivity).proposalId)}</Text>
-      <Text>{metaData.title || metaData.description}</Text>
+      {metaData.title ? <Text>{metaData.title}</Text> : null}
```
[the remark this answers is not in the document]
adamgall: Agree, this makes sense to me. Thank you!

src/components/ui/forms/InputComponent.tsx
```diff
@@ -142,15 +144,24 @@ export function EthAddressComponent(props: EthAddressProps) {
...
+  } = props;
   return (
     <LabelComponent
       {...props}
       disabled={disabled}
     >
       <Textarea
         id={id}
-        resize="none"
+        resize={resize as any}
```
adamgall, line 164: Please no `any` if we can avoid it. This `resize` property should be one of four strings: `"vertical" | "horizontal" | "both" | "none"`

Defaulting to `vertical` is fine, because it seems like that's the behavior when this `resize` prop is left off of a `textarea` anyway.

src/router.tsx
adamgall: Ok. I agree that placing `ModalProvider` inside of `Layout` doesn't seem right, but I also think it doesn't seem right inside of the `router` generator either. I just played around with this on my local machine, and sure enough it breaks unless that `ModalProvider` is inside of the `Router` object.  I understand why (we navigate from within a Modal). Can you please leave a comment in the code here explaining why this `Provider` isn't with the rest of them?

adamgall (approved):
SHIPIT

PR #1439, opened 2024-03-12
Feedback/#1404 timezones

### LINKED ISSUE #1404: Timer - Clarity on Timezone
There is some confusion here by a user on how our timezones work

Image

They were able to check etherscan to confirm the correct UTC time but preferably we should be able to give them guidance to avoid this

For the same proposal I get 12.34 which is the same as UTC time. 

Image

This makes me guess that we take the computer's IP address to work out our local time. If so, is there a way we can just clarify this in the copy or app somehow?

This feels important for our users who are possibly jumping timezones as they travel

adamgall:
Closes #1404 

### Changes

- Adds user's localized timezone (in the form of GMT offset) to Start and End dates of proposals
  - Works on Azorius, Multisig, and Snapshot proposals
  - Screenshot 2024-03-12 at 3 31 10 PM
- Adds tooltip displaying the associated GMT time, which will match what's on Etherscan
  - Works on Azorius, Multisig, and Snapshot proposals
  - Screenshot 2024-03-12 at 3 32 05 PM
  - In the above screenshot the tooltip happens to cover the End Date, fyi.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,6 +34,7 @@
         "axios": "^0.27.2",
         "classnames": "^2.3.1",
         "date-fns": "^2.29.3",
+        "date-fns-tz": "^2.0.1",
         "ethers": "^5.7.2",
         "evm-proxy-detection": "^1.1.0",
         "formik": "^2.2.9",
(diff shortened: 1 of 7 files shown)


Da-Colon (approved):
LGTM

PR #1476, opened 2024-03-20
Enhancement/#1414 unconnected network switch

### LINKED ISSUE #1414: Allow user to switch between networks without a connected wallet
Currently, the app has a default network when no wallet is connected (mainnet for prod, sepolia for dev), and that network is not changeable while no wallet is connected.

The network toggle which appears after a wallet is connected, should be available for use when a wallet is not connected as well.

I'd also like to see some UI in the header showing which network the app is currently using. @xraystyle1980 @nicolaus-sherrill do you have any input here?

adamgall:
Closes #1414 

1. Adds the network switcher component into the Wallet dropdown when a wallet is not connected. Selecting a chain in the dropdown calls `swtichChain` on `wagmi`'s `useSwitchChain` hook, so `wagmi` is aware of the newly selected chain.
2. Updates the `NetworkConfigContext`'s Context to use `wagmi`'s chain ID (via `const chainId = useChainId()`), instead of the wallet's (via `const { chain } = useAccount()`, which returns undefined when no wallet is connected), to set the Context's `NetworkConfig`.
3. Deletes the `disconnectedChain` helper property which was being sporadically used throughout the app to get information about "the chain" when a user's wallet wasn't connected.
  1. Replaces instances of this with simply calling `const network = useNetworkConfig()`, and using that returned `network`, which is correctly set to the connected chain, whether or not a wallet is connected, due to the updates described in 2 above.

Testing

Without a wallet connected:
- Change networks in the account dropdown.
- Search for DAOs on various networks using the search input, make sure the correct message appears.
- Load DAOs on different networks, making sure that they either correctly load or not, based on the network and where that DAO is deployed.


Da-Colon (approved):
🔥 Works and LGTM

PR #1488, opened 2024-03-24
BugFix Dev

### LINKED ISSUE #1487: Outdated Dependency

Da-Colon:
Closes #1487

~see issue

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "packages": {
     "": {
       "name": "fractal-interface",
+      "hasInstallScript": true,
       "dependencies": {
         "@apollo/client": "^3.7.10",
         "@chakra-ui/react": "^2.8.2",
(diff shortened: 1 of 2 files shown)


adamgall (commented):
I notice this removes the `graphql:build` step, which true, only needs to happen once on a dev's machine. 

It might make more sense for that to go into the `postinstall` script... What do you think?

https://docs.npmjs.com/cli/v10/using-npm/scripts#npm-install

Da-Colon:
Actually don't think I meant to leave in that removal. but yeah seemed weird to build if it wasn't updated

Da-Colon:
Yeah def makes sense. I've added the 'postinstall' script

| I'm just trying to figure out how to make that automatically happen without a (new) dev geting stuck or needing to think about it.

It runs directly after `npm install` is that what you mean? (I've just tested)

adamgall:
> It runs directly after `npm install` is that what you mean? (I've just tested)

Yes! This is what I was expecting. This satisfies what I was after, as far as I can tell.

adamgall (approved):

Issue #1457, opened 2024-03-14
Investigate Multiple Requests `useAzoriusProposals`


Da-Colon:
The useEffects that we have setup in `useAzoriousProposals` when a DAO's type is Azorius seems to be causing several repeated request when loading up a DAO.

Da-Colon:
Update: The #1489 PR was created from my effort to investigate all the 429 requests and find the sources of the problems and fix them if I could. While I was able to figure out a the issues with `useSubmitProposal` and `useERC20Claim`, I wasn't able to fix `useAzoriusProposals`. 

There are 3 queryFilters within the process of retrieving the proposals of a DAO

First one retrieves the `ProposalCreated`, this is what is looped through to gather each proposals to gather the rest of the data.

Within the loop, a function is returned `mapProposalCreatedEventToProposal`. Within this workflow there are two more queryFilters that occurs.

One for retrieving the proposals votes (`Voted`)

The other is for retrieving the Transaction Hash for an executed proposal (`Executed`)

PR #1486, opened 2024-03-22
Fractal Guard State update and SUB DAO Fixes

Da-Colon:

PR #1489, opened 2024-03-25
[Maintenance] `canUserSubmitProposal` extracted to own component

### LINKED ISSUE #1501: `canUserSubmitProposal` not working correctly everywhere
Noticed that there are several additional requests coming out of `useSubmitProposal`. On a closer look there is a useEffect firing on mount to request this information depending on which DAO type and user is connected. But this doesn't need to happen everywhere that `useSubmitProposal` is used. The conditional needs to be updated so that the information isn't requested every render.

Da-Colon:
Draft till merged; Branched from #1495 

## Investigate Additional Request in `useAzoriosProposals` hook

Ill update the card but wasn't able to improve anything in this workflow. 
Updated #1457

## No issue, Investigate additional Requests in `useSubmitProposals`

There was logic for `canUserSubmitProposal` that was firing in a useEffect when this hook was mounted. Several places through the app mounted this hook but didn't use it. So I removed the logic into its own hook and replaced it through out the app. 
Closes #1501 

This update reduces the a few extra requests and also fixes permissions not properly being set when refreshing on proposal templates

## Testing.

There isn't really anything much to directly to test for this update. But you can ensure that refreshing on a Proposals Details screen or when creating a proposal, that permissions still load properly (like being able to create a proposal or template, button not disabled.)


src/hooks/DAO/loaders/governance/useERC20Claim.ts
```diff
@@ -39,11 +45,20 @@ export function useERC20Claim() {
...
+      daoAddress + votesTokenContractAddress !== loadKey.current
+    ) {
+      loadKey.current = daoAddress + votesTokenContractAddress;
       loadTokenClaimContract();
     }
-  }, [loadTokenClaimContract, daoAddress]);
+    if (!daoAddress || !votesTokenContractAddress) {
+      loadKey.current = undefined;
+    }
+  }, [loadTokenClaimContract, daoAddress, votesTokenContractAddress]);
```
Da-Colon, line 62: added ref load key to prevent extra requests

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
```diff
@@ -61,23 +60,13 @@ export function ManageDAOMenu({
...
 
-  useEffect(() => {
-    const loadCanUserCreateProposal = async () => {
-      if (safeAddress) {
-        setCanUserCreateProposal(await getCanUserCreateProposal(safeAddress));
-      }
-    };
-
-    loadCanUserCreateProposal();
-  }, [safeAddress, getCanUserCreateProposal]);
```
Da-Colon, line 82: This logic already happens inside the hook

src/hooks/DAO/proposal/useSubmitProposal.ts
Da-Colon, line 1: This logic wasn't being used within the 'useSubmitProposal itself, but it was causing this 'useEffect' to execute every where it is mounted causing extra requests

src/hooks/utils/useCanUserSubmitProposal.ts
```diff
@@ -0,0 +1,105 @@
...
+    ],
+  );
+  useEffect(() => {
+    const loadCanUserCreateProposal = async () => {
+      setCanUserCreateProposal(await getCanUserCreateProposal());
+    };
+    if (canUserCreateProposal === undefined) {
+      loadCanUserCreateProposal();
+    }
+  }, [getCanUserCreateProposal, canUserCreateProposal, user]);
```
Da-Colon, line 102: Some changes to logic here. added a conditional so that once setup it doesn't continue to request this data.

I also updated the return above ensuring that only when it is actually trying to check the permission a boolean is returned. 

The falsey setup conditional will just return undefined.

adamgall:
@Da-Colon is the target branch for this PR correct? I see it's no longer in draft, but it doesn't target `develop`.

adamgall:
Maybe it's just me, but DAOs don't load for me on this branch.

Da-Colon:
> Maybe it's just me, but DAOs don't load for me on this branch.

weird. I just loaded one up on the preview.

adamgall:
> > Maybe it's just me, but DAOs don't load for me on this branch.
> 
> weird. I just loaded one up on the preview.

Hm yes, you're right, me too. Must be something weird with my local dev env. I'll debug more on my side.

adamgall:
@Da-Colon can you please create an issue and explain what needs fixin, to attach to this PR?

Da-Colon:
@adamgall For testing. This PR mostly eliminates several of the 429s we were seeing in the console but Also one area where a bug is shown from the speed up is. 

If you pull up #1486 Go to any DAO (I did this with a multisig) and go to create a new proposal Templates. Once on the create screen refresh.

When you fill out the form, the button to submit the form will remain disabled. (It may be flaky and work but I got a pretty high occurance) (so you don't have to fill out the form everytime, you can also fork a template then refresh on that screen)

When you do the same on this PR you won't have that problem.

src/hooks/utils/useCanUserSubmitProposal.ts
adamgall: Does `user` still need to be part of this dependency array?

src/hooks/DAO/loaders/governance/useERC20Claim.ts
adamgall: Looks like the changes to this file were reverted. Intentional?

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
adamgall: Yes, I see. Thank you

src/hooks/utils/useMasterCopy.ts
```diff
@@ -58,6 +58,11 @@ export function useMasterCopy() {
       const filter = contract.filters.ModuleProxyCreation(proxyAddress, null);
       return contract.queryFilter(filter).then(proxiesCreated => {
         if (proxiesCreated.length === 0) {
+          setValue(
+            CacheKeys.MASTER_COPY_PREFIX + proxyAddress,
+            constants.AddressZero,
+            CacheExpiry.ONE_WEEK,
+          );
```
adamgall, line 65: Would you mind commenting on these additions. What benefits do we get by adding this data to local storage cache?

adamgall (commented):
Looking great @Da-Colon , i just have two comments/questions before giving an approval.

src/hooks/utils/useMasterCopy.ts
Da-Colon: Add a comment to the code

src/hooks/utils/useCanUserSubmitProposal.ts
adamgall: I see it has been removed, cool

adamgall (approved):

PR #1493, opened 2024-03-26
Fix BigNumber Typo in templates

Da-Colon:
While exploring the myosin DAO's template execution error is was discovered that the reason why the template would not execution was caused by an error in the template's creation. The json contains a typo in the `ethValue` property. newer templates doesn't have this typo. 

I did however notice that one more typo still exists though it doesn't seem to affect new templates.

Edit:
This was in fact effecting creation of forked templates.


adamgall (approved):
Lot going on here, too much to leave a proper review on. Going to Approve it anyway because I don't want to hold things up.

PR #1495, opened 2024-03-26
[Maintenance] Investigate extra requests `useERC20Claim`

### LINKED ISSUE #1456: Investigate Multiple Requests `useERC20Claim`
There seems to be several additional "too many requests" coming out of ERC20Claim hook when loading a Azorius DAO. We should see what is causing this.

Da-Colon:
Branched from #1486 

## Investigate Additional Requests in `useERC20Claim` hook
Multiple re-renders was causing this to run this request more than once. Added a conditional to only allow this to fire once. 
Closes #1456

### DIFF
--- a/src/hooks/DAO/loaders/governance/useERC20Claim.ts
+++ b/src/hooks/DAO/loaders/governance/useERC20Claim.ts
@@ -1,4 +1,4 @@
-import { useEffect, useCallback } from 'react';
+import { useEffect, useCallback, useRef } from 'react';
 import { useFractal } from '../../../../providers/App/AppProvider';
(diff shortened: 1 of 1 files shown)


src/hooks/DAO/loaders/governance/useERC20Claim.ts
Da-Colon, line 1: Most DAOs won't have a ERC20Claim contract. (assumingly). Maybe we need a better way to check for claiming contract?

But for now. It did not have anything preventing it from reattempting the requests as the app's initial render settles. This update adds a loadkey to prevent additional loads.

PR #1496, opened 2024-03-27
Enhancement/1483 chainspacing

### LINKED ISSUE #1336: Oops! Couldn't find a Safe...
Screenshot 2024-01-24 at 8 22 43 AM

Sometimes this message appears when attempting to load a Safe.

In the case of this screenshot, it appeared while on the `app.dev.fractalframework.xyz` domain. I did _not_ have any wallet connected to the site, so I assume that the default network for the `app.dev` framework is Goerli, which is referenced in the error message.

I'm opening this issue for two reasons.

1. Why am I being prompted to refresh the page? If a Safe on Goerli doesn't exist at this address, what will refreshing the page accomplish? (i.e., is "refreshing the page" a bandaid fix on top of a deeper data-availability issue?)
2. Why am I being prompted to change the networks in my wallet? I don't have a wallet connected, so I can't change networks.

### LINKED ISSUE #1483: Chainspacing the URL
Fractal URLs should include the network for a Safe. This will provide the following benefits:
- The app can know which network the user intends to load up the given Safe in.
- The app can either automatically switch to the intended network, or clearly tell the user that the app is on the wrong network.

Considerations:
- Search bar
- Loading up URLs that don't have network identifiers (for example, from bookmarks)

Workflows:
- Searching for a DAO from the search bar
  - should automatically populate the network prefix to the URL

adamgall:
Closes #1483 
Closes #1336 

This PR implements `chainspacing` in the URL.

All URLs that exist in the context of a Safe have a network ID (same ones that the Safe UI uses) prefixed to the address.

Additionally, error page logic has been separated to clearly identify the two scenarios that might occur for a Safe being unable to be loaded:
- The given address simply isn't a Safe on the expected (as per the network prefix) Safe.
- The app is connected to a different network than what the Safe is expecting (as per the network prefix).

Additionally, stronger typing has been enforced in the many places were navigation links exist. This stronger typing required early-exiting rendering attempts where either the network prefix or address are not available yet (e.g. as a Safe is loading), so special review and testing should be taken to ensure that the logical changes around rendering links was implemented correctly.

## Testing

Functional changes in this PR include:

- Safes will only load if they include a network prefix in the new query param-based address.
- automatically applying a network prefix to the address in the new query param-based url _only when redirecting from the old path-based URL_.
- fixing all internal links and navigation in the app to apply and utilize the new query param-based safe addresses.
- splitting the "can't load Safe" error screen into three distinct messages
  - for when there is a problem with the format of the address query param
  - for when the app is connected to a different network than the network prefix in the address query param
  - for when the syntax and network are all good, but the Safe can't load because of actual loading issues, like the Safe API is down or the address just isn't an actual Safe.


src/hooks/DAO/useDeployDAO.ts
```diff
@@ -13,10 +14,12 @@ const useDeployDAO = () => {
 
   const { t } = useTranslation('transaction');
 
+  const { addressPrefix } = useNetworkConfig();
```
adamgall, line 17: Unsure if this was the proper place to get the network prefix from. I don't think it's guaranteed that we'll have a loaded Safe when this hook is being used, so that's why I pulled the value from `useNetworkConfig`.

src/types/fractal.ts
```diff
@@ -234,6 +234,7 @@ export interface FractalGovernanceContracts {
 export interface FractalNode {
   daoName: string | null;
   daoAddress: string | null;
+  daoNetwork: string | null;
```
adamgall, line 237: Here's a big question that might have some big consequences for this PR... was it even appropriate to add this new property to the `FractalNode` state? Should I just be pulling the network for a given DAO from `useNetworkConfig`? Are there places in the app where the DAO loaded up in state belongs to a _different network_ than the one that the app is connected to (which is known in `useNetworkConfig`)?

src/hooks/DAO/useDAOController.ts
```diff
@@ -12,27 +12,40 @@ import { useGovernanceContracts } from './loaders/useGovernanceContracts';
...
-  const daoAddress = searchParams.get('dao');
+  const addressWithPrefix = searchParams.get('dao');
+
+  if (addressWithPrefix === null) {
+    throw new Error('address is null');
+  }
+
+  if (!addressWithPrefix.includes(':')) {
+    throw new Error("address doesn't include prefix");
+  }
```
adamgall, line 23: Unsure if this is the appropriate way to throw these critical errors... Maybe these need to be caught in an Error Boundary somewhere. Thoughts? Can test these out by manually removing sections of the address (like the network prefix).

src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -87,7 +87,7 @@ export default function useSubmitProposal() {
 
   const lookupModules = useFractalModules();
   const signerOrProvider = useSignerOrProvider();
-  const { chainId, safeBaseURL } = useNetworkConfig();
+  const { chainId, safeBaseURL, addressPrefix } = useNetworkConfig();
```
adamgall, line 90: Here's another place where I pulled the network prefix from `useNetworkConfig` instead of `useFractal`. Thoughts on which approach is more appropriate?

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -84,18 +95,24 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
   }, [safeAPI, daoAddress]);
 
   const setDAO = useCallback(
-    async (_chainId: number, _daoAddress: string) => {
+    async (_daoNetwork: string, _daoAddress: string, _connectedNetwork: string) => {
```
adamgall, line 98: This function now takes both the network for the given DAO, as well as the network we're currently connected to, in order to properly set the appropriate error state.

src/components/ui/modals/ConfirmModifyGovernanceModal.tsx
```diff
@@ -7,9 +7,13 @@ import { useFractal } from '../../../providers/App/AppProvider';
 export function ConfirmModifyGovernanceModal({ close }: { close: () => void }) {
   const { t } = useTranslation('modals');
   const {
-    node: { daoAddress },
+    node: { daoAddress, daoNetwork },
   } = useFractal();
 
+  if (!daoNetwork || !daoAddress) {
+    return <></>;
+  }
```
adamgall, line 15: Is this cool? I think and hope so.

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
```diff
@@ -124,19 +123,24 @@ export function ManageDAOMenu({
...
   const handleModifyGovernance = useFractalModal(ModalType.CONFIRM_MODIFY_GOVERNANCE);
 
   const options = useMemo(() => {
     const createSubDAOOption = {
       optionKey: 'optionCreateSubDAO',
-      onClick: () => navigate(DAO_ROUTES.newSubDao.relative(safeAddress)),
+      onClick: () => {
+        if (daoNetwork && daoAddress) {
+          navigate(DAO_ROUTES.newSubDao.relative(daoNetwork, daoAddress));
+        }
```
adamgall, line 142: Is this new conditional cool? I think and hope so.

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
```diff
@@ -124,19 +123,24 @@ export function ManageDAOMenu({
...
+  }, [fractalNode, safe, daoAddress, type, getZodiacModuleProxyMasterCopyData, baseContracts]);
 
-  const handleNavigateToSettings = useCallback(
-    () => navigate(DAO_ROUTES.settings.relative(safeAddress)),
-    [navigate, safeAddress],
-  );
+  const handleNavigateToSettings = useCallback(() => {
+    if (daoNetwork && daoAddress) {
+      navigate(DAO_ROUTES.settings.relative(daoNetwork, daoAddress));
+    }
```
adamgall, line 131: Is this new conditional cool? I think and hope so.

src/components/ui/menus/FavoritesMenu/FavoritesList.tsx
```diff
@@ -45,6 +47,7 @@ export function FavoritesList() {
             {favoritesList.map(favorite => (
               <Favorite
                 key={favorite}
+                network={addressPrefix}
```
adamgall, line 50: I don't love having to pull the "current network" off of `useNetworkConfig` to render these Favorite links. Wish I could have pulled these directly off the local storage information, but those are not very well reusable. So, since the `favoritesList` is only coming back with favorites for the given connected network anyway, this seemed safe enough for now.

src/components/ui/menus/DAOSearch/SearchDisplay.tsx
```diff
@@ -81,6 +75,12 @@ export function SearchDisplay({
           <Button
             alignSelf="center"
             data-testid="search-viewDAO"
+            onClick={() => {
+              onClickView();
+              if (closeDrawer) closeDrawer();
+              action.resetDAO();
+              navigate(DAO_ROUTES.dao.relative(addressPrefix, address));
+            }}
```
adamgall, line 83: Not super related to this PR, but I snuck this change in here anyway. Basically, made it so that only the button in the search result list is clickable, not the whole element.

src/components/pages/DaoSettings/components/Metadata/index.tsx
```diff
@@ -50,7 +50,11 @@ export default function MetadataContainer() {
 
   const userHasVotingWeight = votingWeight.gt(0);
 
-  const submitProposalSuccessCallback = () => navigate(DAO_ROUTES.proposals.relative(daoAddress));
+  const submitProposalSuccessCallback = () => {
+    if (daoNetwork && daoAddress) {
+      navigate(DAO_ROUTES.proposals.relative(daoNetwork, daoAddress));
+    }
```
adamgall, line 56: Is this new conditional cool? I think and hope so.

src/components/pages/DaoDashboard/Info/index.tsx
```diff
@@ -135,18 +135,22 @@ export function Info() {
...
               pt={{ sm: PAD, lg: NONE }}
             >
-              <InfoBox to={DAO_ROUTES.treasury.relative(daoAddress)}>
-                <InfoTreasury />
-              </InfoBox>
+              {daoNetwork && daoAddress && (
+                <InfoBox to={DAO_ROUTES.treasury.relative(daoNetwork, daoAddress)}>
+                  <InfoTreasury />
+                </InfoBox>
+              )}
```
adamgall, line 153: Is this new conditional cool? I think and hope so.

src/components/pages/DaoDashboard/Info/index.tsx
```diff
@@ -135,18 +135,22 @@ export function Info() {
...
               pt={{ sm: PAD, lg: NONE }}
             >
-              <InfoBox to={DAO_ROUTES.proposals.relative(daoAddress)}>
-                <InfoProposals />
-              </InfoBox>
+              {daoNetwork && daoAddress && (
+                <InfoBox to={DAO_ROUTES.proposals.relative(daoNetwork, daoAddress)}>
+                  <InfoProposals />
+                </InfoBox>
+              )}
```
adamgall, line 142: Is this new conditional cool? I think and hope so.

src/components/Proposals/ProposalCard/ProposalCard.tsx
```diff
@@ -26,8 +26,12 @@ function ProposalCard({ proposal }: { proposal: FractalProposal }) {
 
   const isSnapshotProposal = !!(proposal as SnapshotProposal).snapshotProposalId;
 
+  if (!daoNetwork || !daoAddress) {
+    return <></>;
+  }
```
adamgall, line 31: Is this cool? I think and hope so.

adamgall (commented):
Left a handful of comments on some implementation details. Would love more eyes and brains on this.

tomstuart123 (approved):

tomstuart123:
@adamgall - approval given 

I did lots of testing of this PR with Decent, shutter, and myosin by pasting old URLs into this deploy link
Good news is they were all handled successfully :slightly_smiling_face:

I can also confirm your point above that linking to an old link to the treasury page, will now link back to the core homepage (note - this is absolutely fine)

src/components/Proposals/ProposalCard/ProposalCard.tsx
Da-Colon: Shouldn't by time this mounts. You can also just return `null`

adamgall: `return null;`! of course. i was just doing `return;` originally and then the compiler was yelling at me (because the implicit `undefined` there wasn't acceptable where the component was being instantiated). I forgot about returning null. i like that better, will go update now.

adamgall: updated and pushed

src/components/ui/menus/FavoritesMenu/FavoritesList.tsx
Da-Colon: We've talked before about revisiting LocalStorage setup. Should a card be made to start noting this?

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
```diff
@@ -56,34 +56,33 @@ export function ManageDAOMenu({
...
   const [governanceType, setGovernanceType] = useState(GovernanceType.MULTISIG);
   const {
-    node: { safe },
+    node: { safe, daoNetwork, daoAddress },
     governance: { type },
     baseContracts,
   } = useFractal();
   const currentTime = BigNumber.from(useBlockTimestamp());
   const navigate = useNavigate();
-  const safeAddress = fractalNode?.daoAddress;
```
Da-Colon, line 65: The way the safeAddress is used in this component needs to be the same. `ManageDAOMenu` it used to pull subDAO data as well on Hierarchy Page. 

The props should be what is used not the address from global state.

src/constants/routes.ts
```diff
@@ -3,7 +3,10 @@ export const BASE_ROUTES = {
   create: '/create',
 };
 
-type RouteInfo = { relative: (...args: any) => string; path: string };
+type RouteInfo = {
+  relative: (daoNetwork: string, daoAddress: string, ...args: string[]) => string;
```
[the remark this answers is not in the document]
Da-Colon: o I see em 👁️

src/hooks/DAO/proposal/useSubmitProposal.ts
Da-Colon: No thoughts, I don't see the harm in pulling from NetworkConfig provider

src/hooks/DAO/useDAOController.ts
Da-Colon: Haven't live tested. What happens? the ErrorBoundary? (I'll probably test it before you respond ha.)

src/main.tsx
```diff
@@ -10,12 +10,17 @@ import '@fontsource/ibm-plex-sans';
...
+  return <RouterProvider router={router(addressPrefix)} />;
+}
+
 const root = document.getElementById('root');
 if (root !== null) {
   ReactDOM.createRoot(root).render(
     <React.StrictMode>
       <Providers>
-        <RouterProvider router={router} />
+        <FractalRouterProvider />
```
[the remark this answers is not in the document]
Da-Colon: that makes sense

src/types/fractal.ts
Da-Colon: hmmm ah now I see the thought here (last comment I've read ha). 

Need to think more on this. At first I was going to say it doesn't matter but we probably don't need this in two places. as these should always be the same. The DAO won't work if the contracts are the right state. these things are tied together. So makes sense to keep this only read from a single place (IMO networkConfig)

Da-Colon (changes requested):
Nice! Answer to your big question and one requested change in `ManageDAOMenu`

src/hooks/DAO/useDAOController.ts
adamgall: I need to test this too

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
adamgall: Oh snap, great find. I didn't realize I was making an actual change Here. I didn't realize that `fractalNode` came from the constructor, I was incorrectly thinking it came out of `useFractal()` a few lines above.

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -84,18 +91,24 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
...
+    async (_addressPrefix: string, _daoAddress: string, _connectedAddressPrefix: string) => {
       setNodeLoading(true);
       setErrorLoading(false);
-      if (utils.isAddress(_daoAddress) && safeAPI) {
+      setWrongNetwork(false);
+
+      if (_connectedAddressPrefix !== _addressPrefix) {
+        currentValidSafe.current = undefined;
+        action.resetDAO();
+        setWrongNetwork(true);
```
adamgall, line 102: If the address prefix (from the URL) is different than the address prefix (of the network the app is connected to) reset the DAO and set this specific error state

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -120,20 +133,18 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
...
-    [action, safeAPI, lookupModules, fetchSafeInfo, requestWithRetries],
+    [safeAPI, action, requestWithRetries, fetchSafeInfo, lookupModules],
   );
 
-  const { chainId } = useNetworkConfig();
   useEffect(() => {
-    if (daoAddress && chainId + daoAddress !== currentValidSafe.current) {
+    if (
+      addressPrefix + daoAddress !== currentValidSafe.current ||
+      addressPrefix !== connectedAddressPrefix
```
adamgall, line 142: Two thing here:
1. Cleaned up the existing conditional -- "dao address" is always a non-empty string, so we don't need to check for its existence like we were doing originally.
2. added an "or" condition to check if the connected address prefix has changed from what's in the url

src/hooks/DAO/useDAOController.ts
adamgall: Just tested it, this does not exist in an Error Boundary

Screenshot 2024-03-28 at 7 15 27 PM

I'll need to fix this as part of this PR.

adamgall (commented):
Left a couple new comments as per my most recent pushed work.

In general, the majority of my _new_ updates tackle:
1. Fixing the unintentional bug I introduced in `ManageDAOMenu` (thanks @Da-Colon for finding that)
2. removing `daoNetwork` from Fractal state, and just using the value from `useNetworkConfig` instead, throughout the app.

src/pages/DAOController.tsx
```diff
@@ -1,40 +1,13 @@
...
-      </VStack>
-    </Center>
-  );
-}
+import LoadingProblem from './LoadingProblem';
 
 export default function DAOController() {
-  const { node } = useFractal();
-  const { nodeLoading, errorLoading } = useDAOController();
+  const { errorLoading, wrongNetwork, invalidQuery } = useDAOController();
```
adamgall, line 10: `useDAOController` will simply return back to us information for if any of the three error states have hit.

src/pages/DAOController.tsx
```diff
@@ -68,10 +40,15 @@ export default function DAOController() {
...
+    // the order of the if blocks of these next three error states matters
+  } else if (invalidQuery) {
+    display = <LoadingProblem type="badQueryParam" />;
+  } else if (wrongNetwork) {
+    display = <LoadingProblem type="wrongNetwork" />;
+  } else if (errorLoading) {
+    display = <LoadingProblem type="invalidSafe" />;
   } else {
-    display = <InvalidSafe />;
+    display = <Outlet />;
```
adamgall, line 51: If any of the three error states have hit, show the appropriate LoadingProblem, otherwise let that mfkr load

src/hooks/DAO/useDAOController.ts
```diff
@@ -12,27 +13,37 @@ import { useGovernanceContracts } from './loaders/useGovernanceContracts';
...
+  const invalidQuery = addressWithPrefix === null || !validDaoQueryString.test(addressWithPrefix);
+
+  const prefixAndAddress = addressWithPrefix?.split(':');
+  const addressPrefix = prefixAndAddress?.[0];
+  const daoAddress = prefixAndAddress?.[1];
+
+  const { addressPrefix: connectedAddressPrefix } = useNetworkConfig();
+  const wrongNetwork = addressPrefix !== connectedAddressPrefix;
+
+  const skip = invalidQuery || wrongNetwork;
```
adamgall: Just setting up some state here, to be used down below. This is where we check to see if the query param is formatted correctly, and that we're on the right network. Then we have a variable called `skip` which is needed to be passed into `useFractalNode`, which will, you guessed it, skip loading and reset state if true.

src/hooks/DAO/useDAOController.ts
```diff
@@ -12,27 +13,37 @@ import { useGovernanceContracts } from './loaders/useGovernanceContracts';
...
-      action.resetDAO();
-    }
-  }, [action, daoAddress]);
 
-  const { nodeLoading, errorLoading } = useFractalNode({ daoAddress: daoAddress || undefined });
+  const { errorLoading } = useFractalNode(skip, {
+    addressPrefix,
+    daoAddress,
+  });
+
```
adamgall, line 43: Here we are, let `useFractalNode` do it's thing. The only return value from here is `errorLoading` which signifies the error state that means that the given address just isn't a Safe.

src/hooks/DAO/useDAOController.ts
```diff
@@ -12,27 +13,37 @@ import { useGovernanceContracts } from './loaders/useGovernanceContracts';
...
   useGovernanceContracts();
   useFractalGuardContracts({});
   useFractalFreeze({ parentSafeAddress: parentAddress });
   useFractalGovernance();
   useFractalTreasury();
   useERC20Claim();
   useSnapshotProposals();
-  return { nodeLoading, errorLoading };
+
+  return { invalidQuery, wrongNetwork, errorLoading };
```
adamgall, line 52: Return `invalidQuery`, `wrongNetwork` (both of which were determined in this hook), and `errorLoading` (which will have come from the returned value of `useFractalNode`).

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -76,64 +84,73 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
...
-    }
-  }, [safeAPI, daoAddress]);
+  const reset = useCallback(
+    ({ error }: { error: boolean }) => {
+      currentValidSafe.current = undefined;
+      action.resetDAO();
+      setErrorLoading(error);
+    },
+    [action],
+  );
```
adamgall, line 94: Made a new little helper function to do this because it was repeated a few times.

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -76,64 +84,73 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
     pollInterval: ONE_MINUTE,
   });
 
-  const fetchSafeInfo = useCallback(async () => {
-    if (daoAddress && safeAPI) {
-      const safeInfo = await safeAPI.getSafeInfo(utils.getAddress(daoAddress));
-      return safeInfo;
-    }
-  }, [safeAPI, daoAddress]);
```
adamgall, line 84: Ended up just inlining this, it was a one-liner.

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -14,10 +14,18 @@ import { useFractalModules } from './useFractalModules';
...
+    addressPrefix,
+    daoAddress,
+  }: {
+    addressPrefix?: string;
+    daoAddress?: string;
+  },
+) => {
   // tracks the current valid Safe address and chain id; helps prevent unnecessary calls
   const currentValidSafe = useRef<string>();
-  const [nodeLoading, setNodeLoading] = useState<boolean>(true);
```
adamgall, line 20: We don't actually care about tracking whether or not the DAO is loading. 

Reason: From the point of view of the rest of the app, as long as there are no errors, we try to render. If there's no data in state, loading components are shown. As data finds its way into state, the components render that data.

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -76,64 +84,73 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
...
+      action.dispatch({
+        type: NodeAction.SET_FRACTAL_MODULES,
+        payload: await lookupModules(safeInfo.modules),
+      });
+
+      action.dispatch({
+        type: NodeAction.SET_SAFE_INFO,
+        payload: safeInfo,
+      });
     },
```
adamgall: This was mostly just flattening out the logic, putting all errors and exiting up top, and doing the happy path stuff at the bottom.

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -76,64 +84,73 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
...
   );
 
-  const { chainId } = useNetworkConfig();
   useEffect(() => {
-    if (daoAddress && chainId + daoAddress !== currentValidSafe.current) {
-      setNodeLoading(true);
-      setDAO(chainId, daoAddress);
+    if (skip) {
+      reset({ error: false });
+      return;
```
adamgall: Ok, into the main loop useEffect for this hook. If the passed-in `skip` is true (which means there was an error with the query param or we're on the wrong network), then clear state and GTFO.

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -76,64 +84,73 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
...
     }
-    if (!daoAddress) {
-      currentValidSafe.current = undefined;
-      setNodeLoading(false);
+
+    if (addressPrefix === undefined || daoAddress === undefined) {
+      reset({ error: false });
+      return;
+    }
```
adamgall: Otherwise, our other input params might be `undefined` (just because of the way things are), so same deal, clear state and GTFO.

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -76,64 +84,73 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
...
+
+    if (addressPrefix + daoAddress !== currentValidSafe.current) {
+      setDAO(addressPrefix, daoAddress);
     }
```
adamgall, line 143: And now, if we got here, we only want to call `setDAO` if we haven't done so yet for the given address

src/i18n/locales/en/common.json
```diff
@@ -98,7 +98,11 @@
   "isMeSuffix": " (you)",
   "refresh": "Refresh",
   "invalidSafe1": "We couldn't find a Safe at this address on {{chain}}.",
-  "invalidSafe2": "Try refreshing the page or changing networks in your wallet.",
+  "invalidSafe2": "Please double check the address then try again.",
+  "wrongNetwork1": "Fractal is currently connected to {{chain}}.",
+  "wrongNetwork2": "Try switching the app to the correct chain for your Safe.",
+  "badQueryParam1": "The Safe address in the URL isn't formatted correctly",
+  "badQueryParam2": "",
```
adamgall, line 105: Better error strings for the various kinds of initial loading issues that might occur

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -76,64 +84,64 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
...
+
+      action.dispatch({
+        type: NodeAction.SET_FRACTAL_MODULES,
+        payload: await lookupModules(safeInfo.modules),
+      });
+
+      action.dispatch({
+        type: NodeAction.SET_SAFE_INFO,
+        payload: safeInfo,
+      });
```
adamgall, line 130: This was mostly just a flattening of logic and moving all non-get-data stuff out of this file and back into useDAOController. This useFractalNode hook is now only used for fetching stuff, no more checking if input data is valid, if it's in here (and not undefined) we should use it to go start fetching.

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -76,64 +84,64 @@ export const useFractalNode = ({ daoAddress }: { daoAddress?: string }) => {
...
   );
 
-  const { chainId } = useNetworkConfig();
   useEffect(() => {
-    if (daoAddress && chainId + daoAddress !== currentValidSafe.current) {
-      setNodeLoading(true);
-      setDAO(chainId, daoAddress);
+    if (skip || addressPrefix === undefined || daoAddress === undefined) {
+      reset({ error: false });
+      return;
```
adamgall, line 138: If `skip` is true, that means there was an error with the address or formatting or we're on the wrong network. If addressPrefix of daoAddress is undefined, then damn the App is still standing up. In either case, just reset state and GTFO.

src/hooks/DAO/useDAOController.ts
```diff
@@ -12,27 +13,41 @@ import { useGovernanceContracts } from './loaders/useGovernanceContracts';
...
+
+  const invalidQuery =
+    addressWithPrefix === null ||
+    !validDaoQueryString.test(addressWithPrefix) ||
+    !utils.isAddress(daoAddress || '');
+
+  const { addressPrefix: connectedAddressPrefix } = useNetworkConfig();
+  const wrongNetwork = addressPrefix !== connectedAddressPrefix;
+
+  const skip = invalidQuery || wrongNetwork;
```
adamgall, line 31: This is fun. Do all the things needed to validate that:
1. we have a query string of the correct key
    - `dao`
2. the value of that key is formatted correctly
    - string-without-whitespace followed by ":" followed by string-without-whitespace
3. the ethereum address portion is formatted correctly
    - right length, checksum is valid if exists, etc
4. the network identified in the query param matches the network the app is currently on
    - so if the user is trying to load a `sep:0xabc` dao, the app must be on Sepolia

If any of those things fail, we don't try to load anything and show the appropriate error screen.

adamgall:
@Da-Colon @mudrila @tomstuart123 I have just pushed a bunch of fixes and nice QoL improvements, this is ready for review again.

src/types/fractal.ts
adamgall: Removed `daoNetwork` from `FractalNode`, now it only lives in Network state

src/hooks/DAO/useDAOController.ts
adamgall: Fixed

src/components/ui/menus/FavoritesMenu/FavoritesList.tsx
adamgall: Eh, not now. This seems like a big thing that I don't want us to tackle yet.

src/hooks/DAO/useDeployDAO.ts
adamgall: narrator: it was the proper place after all

src/components/pages/DaoDashboard/Info/index.tsx
adamgall: This one actually wasn't fine. It was stopping the InfoCard on the DAO Dashboard page from rendering with the Loading icon, during initial DAO load. This has been updated to `DAO_ROUTES.proposals.relative(daoNetwork, daoAddress || '')`

src/components/pages/DaoDashboard/Info/index.tsx
adamgall: This one actually wasn't fine. It was stopping the InfoCard on the DAO Dashboard page from rendering with the Loading icon, during initial DAO load. This has been updated to `DAO_ROUTES.proposals.relative(daoNetwork, daoAddress || '')`

src/hooks/DAO/loaders/useFractalNode.ts
adamgall: This got refactored again in a later commit, go check another comment down below about it.

src/hooks/DAO/loaders/useFractalNode.ts
adamgall: This got refactored again in a later commit, go check another comment down below about it.

tomstuart123:
Tried the new netlify link with 4 prior DAOs.

**TLDR**
- All mainnet prior DAO links worked when on the right chain already and were reformated correctly
- All sepolia prior dao links didn't work when on the right chain already and were given this
Screenshot 2024-03-29 at 13 08 53
- When on sepolia (different chain), all mainnet prior DAO links saw the below message which I believe is the intended error message 
Screenshot 2024-03-29 at 13 11 07

**Mainnet DAO links tried**
1) Decent multisig (mainnet) - https://app.dev.fractalframework.xyz/daos/0xD26c85D435F02DaB8B220cd4D2d398f6f646e235
2) Decent's tokenvoting (mainnet) - https://app.dev.fractalframework.xyz/daos/0xB98d45F9021D71E6Fc30b43FD37FB3b1Bf12c064
3) Shutter's token voting  (mainnet)- https://app.dev.fractalframework.xyz/daos/0x36bD3044ab68f600f6d3e081056F34f2a58432c4
4) AwakeVCs - https://app.dev.fractalframework.xyz/daos/0xdD6CeFA62239272f1eDf755ba6471eacb7DF2Fa5

**Sepolia DAO links tried**
1) Myosin multisig - https://app.dev.fractalframework.xyz/home?dao=0xdef90A94273a1A1A72B33D39129fa41E6C08Be3a
2) One of my sepolia test dao s- https://app.dev.fractalframework.xyz/home?dao=0xA287241240B5C8fce743Fb291D4F38cD669298d4

adamgall:
@tomstuart123 the Sepolia links you tried are indeed formatted incorrectly. Notice how they are the new query params, but don't include the network prefix. This is invalid as per this new PR.

tomstuart123 (approved):

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
adamgall: Fixed

Da-Colon (approved):
Looks good thanks for the changes!

PR #1500, opened 2024-03-27
Make displayed avatars consistent for same addresses

### LINKED ISSUE #1499: Avatars not consistent between Header and Proposal List
On dev, seeing inconsistencies with the avatars being displayed for a connected address in the header, and the address in the Proposal List.

If the address is the same, the avatars should be the same.

### No ENS name set:
Image

### ENS name set, but no custom PFP:
Image

### ENS name set, custom PFP:
Don't even see the custom PFP anywhere!
Image

adamgall:
Closes #1499 

See attached issue (#1499) for screenshots demonstrating the problem.

### No ENS name set:
Screenshot 2024-03-27 at 2 19 34 PM

### ENS name set, but no custom PFP:
Screenshot 2024-03-27 at 2 18 10 PM

### ENS name set, custom PFP:
Screenshot 2024-03-27 at 2 16 42 PM

### DIFF
--- a/src/components/Activity/ActivityDescriptionGovernance.tsx
+++ b/src/components/Activity/ActivityDescriptionGovernance.tsx
@@ -1,6 +1,7 @@
 import { Box, Flex, Text } from '@chakra-ui/react';
 import { useTranslation } from 'react-i18next';
 import { useGetMetadata } from '../../hooks/DAO/proposal/useGetMetadata';
+import useAvatar from '../../hooks/utils/useAvatar';
 import useDisplayName from '../../hooks/utils/useDisplayName';
 import {
   Activity,
@@ -52,6 +53,8 @@ function ProposalAuthor({ activity }: { activity: Activity }) {
       : multisigProposal.confirmations[0].owner;
 
   const { displayName: author } = useDisplayName(proposer);
+  const avatarURL = useAvatar(author);
+
   return (
     <Flex
       gap={2}
@@ -61,7 +64,8 @@ function ProposalAuthor({ activity }: { activity: Activity }) {
     >
       <Avatar
         size="sm"
-        address={author}
+        address={proposer}
+        url={avatarURL}
       />
       <Box>{author}</Box>
     </Flex>
--- a/src/components/ui/menus/AccountDisplay/MenuButtonDisplay.tsx
+++ b/src/components/ui/menus/AccountDisplay/MenuButtonDisplay.tsx
@@ -25,7 +25,7 @@ export function Connected() {
   } = useFractal();
   const account = user.address;
   const { displayName: accountDisplayName } = useDisplayName(account);
-  const avatarURL = useAvatar(account);
+  const avatarURL = useAvatar(accountDisplayName);
 
   if (!account) {
     return null;


src/components/ui/menus/AccountDisplay/MenuButtonDisplay.tsx
```diff
@@ -25,7 +25,7 @@ export function Connected() {
   } = useFractal();
   const account = user.address;
   const { displayName: accountDisplayName } = useDisplayName(account);
-  const avatarURL = useAvatar(account);
+  const avatarURL = useAvatar(accountDisplayName);
```
adamgall, line 28: This is to fix the custom PFP showing up in header menu.

src/components/Activity/ActivityDescriptionGovernance.tsx
```diff
@@ -52,6 +53,8 @@ function ProposalAuthor({ activity }: { activity: Activity }) {
       : multisigProposal.confirmations[0].owner;
 
   const { displayName: author } = useDisplayName(proposer);
+  const avatarURL = useAvatar(author);
+
```
adamgall, line 57: This is to attempt to get the URL of a custom PFP for the given "author". 

If `author` is an ENS name, and that name has a PFP set, it will return back a URL for where to load that PFP from.

src/components/Activity/ActivityDescriptionGovernance.tsx
```diff
@@ -61,7 +64,8 @@ function ProposalAuthor({ activity }: { activity: Activity }) {
     >
       <Avatar
         size="sm"
-        address={author}
+        address={proposer}
```
adamgall, line 67: Want to pass the `proposer` into here, which is the full hexadecimal ethereum address. This is the seed to the deterministic colorful icon. In the `MenuButtonDisplay` we are also passing in the full hexadecimal address into `Avatar`, so now they match.

src/components/Activity/ActivityDescriptionGovernance.tsx
```diff
@@ -61,7 +64,8 @@ function ProposalAuthor({ activity }: { activity: Activity }) {
...
+        url={avatarURL}
```
adamgall, line 68: And finally, we can optionally pass in `avatarURL` to `Avatar` as well. If this is null, it'll be ignored and the colorful Jazzicon will be generated and rendered.

Otherwise, if `avatarURL` returns the URL of a custom PFP, the `Avatar` component will fetch and load that PFP instead of displaying the Jazzicon.

src/components/Activity/ActivityDescriptionGovernance.tsx
Da-Colon: oof I see the mistake here. got it.

Da-Colon (approved):
🔥

nicolaus-sherrill:
These new avatars and author addys look sliiiiiick

Can we nudge the spacing a bit wider between the proposal title and PFP circle? It's currently looking very same-y. Designs are spec'd at 24px spacing between title and PFP, and 8px spacing between PFP and author. Images attached.

Screenshot 2024-04-03 at 1 18 40 PM
Screenshot 2024-04-03 at 1 24 27 PM

PR #1502, opened 2024-03-29
Add better typing to `useAsyncRetry`

adamgall:
The `useAsyncRetry` hook returns a function called `requestWithRetries`.

`requestWithRetries` takes a function as an input parameter, and then returns back `Promise<any>`. 

What is happening in `requestWithRetries` is that it calls that input function many times until the function succeeds, and then returns back the result of that function. So, it's kind of generic.

However, I think it would be better if there was stronger typing on `requestWithRetries`. Instead of returning back `Promise<any>`, it should return `Promise<T>`, where `T` is the return type of its input function.

This PR does that, then cleans up a couple of new linter errors from that improvement.


Da-Colon (approved):
🔥 that worked like a charm

PR #1514, opened 2024-04-03
remove Current DAO from favorites list

Da-Colon:
No Issue. But I'll add This PR to the project board in its stead. 

Noticed this bug earlier today, decided to go ahead and fix it.

Reproduce: (on dev)
1. Go to any DAO thats is a favorite
2. Click the same DAO you are viewing

You will see that it gets stuck in loading state.

This fixes by removing it from the list in the first place.


adamgall (commented):
No thank you, I don't like this solution to the problem.

Kind of confusing behavior to _not_ see something that you've Favorited in the Favorites list.

PR #1516, opened 2024-04-04
Bug/multisig nonce selection

### LINKED ISSUE #1474: Nonce selection on new Multisig Proposal not taking queued transactions into account

adamgall:
Closes #1474 

### Outstanding questions

- Are there any other places where a multisig proposal is prepared, other than the obvious "new proposal" page which has been handled in this PR?
- [A dev comment](https://github.com/decentdao/fractal-interface/pull/1516#pullrequestreview-1980066871)

### Testing

1. Using a multisg Safe
2. Queue up a transaction so it's pending, but do not execute it! Let's assume this queued transaction has a Safe nonce of "5".
3. Then begin to create a new Proposal/Transaction in Fractal
4. See how the "nonce" field for this proposal you're beginning to create is "6"! Before this PR, it _also_ would have been "5" which would cause a conflict on the Safe.


src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -98,21 +96,22 @@ export const useFractalNode = (
...
         return;
       }
 
-      if (!safeInfo) {
+      // comment to pick up in review: will `safeInfoResponseWithGuard` ever not be set?
+      // typescript doesn't think so.
+      if (!safeInfoResponseWithGuard) {
         reset({ error: true });
         return;
       }
```
adamgall, line 117: I'm not sure if this block of code is needed. Would like some more eyes on it from the other devs.

DarksightKellar: Doesn't seem like it. `return` in catch block ensures execution will never reach this point if the try block didn't ran all the way to and including `safeInfoResponseWithGuard = ...`.

To be super duper certain, I'd move the try opening brace to include
`currentValidSafe.current = _addressPrefix + _daoAddress`, 
in case _that_ throws for some stupidly obscure reason. Murphy's Law or sum' like that lol.

DarksightKellar (dismissed):
LGTM

Knowledge gap question: How does `nonceWithPending` fix this bug exactly? I understand as much as it being equal to the tx count on the DAO, so as to somehow eagerly set the nonce of the next tx to that count (+1?). Just not sure sure how/where the flow makes that happen. We can chat about it later (or if it can be easily explained in a comment, I'll take ti)

DarksightKellar:
To consider regarding "Are there any other places where a multisig proposal is prepared, other than the obvious "new proposal" page which has been handled in this PR?":

Updating DAO settings still seems to keep the same nonce. But, David mentioned something about submitting a proposal of the same nonce as a currently pending proposal that one might want to invalidate. That works currently, but any other DAO setting update (after an unrelated setting update) is still stuck on the now used up nonce. Probably should separate the two intentions (or cook up a way to reject a proposal without having to manually overwrite it, which sounds like a much bigger lift).

What I tried:

- raised edit DAO name proposal. Not executed. Nonce was 2.
- raised 2nd edit DAO name proposal. Same nonce with manual change (this is the bug/feature)
- executed 2nd edit proposal. 1st proposal now shows as "rejected".
- attempted to add a signer to the DAO. Failed because nonce was auto-set to 2.

adamgall:
> To consider regarding "Are there any other places where a multisig proposal is prepared, other than the obvious "new proposal" page which has been handled in this PR?":
> 
> Updating DAO settings still seems to keep the same nonce. But, David mentioned something about submitting a proposal of the same nonce as a currently pending proposal that one might want to invalidate. That works currently, but any other DAO setting update (after an unrelated setting update) is still stuck on the now used up nonce. Probably should separate the two intentions (or cook up a way to reject a proposal without having to manually overwrite it, which sounds like a much bigger lift).
> 
> What I tried:
> 
> * raised edit DAO name proposal. Not executed. Nonce was 2.
> * raised 2nd edit DAO name proposal. Same nonce with manual change (this is the bug/feature)
> * executed 2nd edit proposal. 1st proposal now shows as "rejected".
> * attempted to add a signer to the DAO. Failed because nonce was auto-set to 2.

Awesome, great call out. I'll go investigate this section of the app.

adamgall:
Maybe a better question for the group (@tomstuart123 @Da-Colon @mudrila) is: are there any situations in which we _don't_ want to take into account pending multisig transactions, when crafting new ones?

src/hooks/DAO/loaders/useLoadDAONode.ts
```diff
@@ -54,14 +54,16 @@ export const useLoadDAONode = () => {
...
+          const nextNonce = await safeAPI.getNextNonce(sanitizedDaoAddress);
+          const safeInfoWithGuard = { ...safeInfo, nonce: nextNonce };
 
           const node: FractalNode = Object.assign(graphNodeInfo, {
-            daoName,
-            safe,
-            fractalModules,
+            daoName: await getDaoName(sanitizedDaoAddress, graphNodeInfo.daoName),
+            safe: safeInfoWithGuard,
+            fractalModules: await lookupModules(safeInfo.modules),
```
adamgall, line 66: Here's a place where we're loading up Safe data and not getting it from / putting it into global state, so want to make sure to put the proper next nonce on this one.

src/hooks/DAO/useClawBack.ts
```diff
@@ -25,7 +25,11 @@ export default function useClawBack({ childSafeInfo, parentAddress }: IUseClawBa
       const childSafeBalance = await safeAPI.getBalances(
         utils.getAddress(childSafeInfo.daoAddress),
       );
-      const parentSafeInfo = await safeAPI.getSafeInfo(utils.getAddress(parentAddress));
+
+      const santitizedParentAddress = utils.getAddress(parentAddress);
+      const parentSafeInfo = await safeAPI.getSafeInfo(santitizedParentAddress);
+      const parentSafeNextNonce = await safeAPI.getNextNonce(santitizedParentAddress);
+
```
adamgall, line 32: Here's a place where we're loading up Safe data and not getting it from / putting it into global state, so want to make sure to put the proper next nonce on this one.

tomstuart123:
Hey @adamgall , thanks for this

So tested a few proposals with this at this DAO - https://deploy-preview-1516.app.dev.fractalframework.xyz/home?dao=sep:0x463e491e72f20ba1bD70EF79d0D7a777Fd762Fa4

What worked:
- Submitting an unexecuted nonce at 4, led to nonce at 5 auto updating for all areas of the app I could think of (i.e. i] send assets ii] create proposal iii] create template iv] all settings changes v] add subsafe)

What didn't work:

- [note this is expected behaviour] if you try and create any of the proposals above BEFORE, the UI has read that the proposal is in the list, the old nonce will be used. As expected and no change required

- [to be debated @adamgall ] - if you add multiple proposals in the queue to be executed, the nonce doesn't continue increasing. It will only ever increase by max 1. e.g. in the below 'send token = nonce 4', 'update safe name = nonce 5' and 'update snapshot space = nonce 5'. SImilarly if I try and now create a new proposal across the app, the auto populate is nonce 5

Screenshot 2024-04-05 at 14 21 26

Screenshot 2024-04-05 at 14 22 52

- The benefits of this implementation - I guess this stops increasing the nonce for mult-proposals that are at the same nonce i.e. will be wiped out. 
- The cons of this implementation - If you're creating multi-proposals that go straight to pending, you see the nonce auto increase on the first... but then after that the nonce will stay flat


@adamgall can you confirm if this is intended behaviour

adamgall:
@tomstuart123 the "debatable" behavior you described is not how I intended it to work. looking at your Safe through the Safe's UI tells me that any new proposals being created in Fractal should have a nonce 6.

Now, I did explicitly test this, and it was working for me as I just described here. So, not sure what's going on. Care to hop on a call and figure this out?

Da-Colon (dismissed):

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -103,20 +101,15 @@ export const useFractalNode = (
...
         if (!safeAPI) throw new Error('SafeAPI not set');
 
-        safeInfo = await requestWithRetries(
-          () => safeAPI.getSafeInfo(utils.getAddress(_daoAddress)),
-          5,
-        );
+        const address = utils.getAddress(_daoAddress);
+        const safeInfoResponse = await safeAPI.getSafeInfo(address);
+        const nextNonce = await safeAPI.getNextNonce(address);
+        safeInfo = { ...safeInfoResponse, nonce: nextNonce };
```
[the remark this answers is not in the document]
DarksightKellar: Lool I'm glad you didn't discover this before having to implement it yourself, which led to me understanding a wee bit more what's going on behind the scenes

adamgall:
After speaking with @tomstuart123 , there's some more work to do on this branch.

The outstanding issue is that we only fetch the "next nonce" once, when a Safe is initially loaded in Fractal. If a transaction is created in Fractal (and not executed), then _another_ transaction is created, that second one will still be using the "next nonce" from the first one.

We need to figure out the cleanest way to re-fetch the "next nonce"...
- on an interval? blah
- whenever a proposal is being created? maybe...
- any other ideas?

Da-Colon:
> After speaking with @tomstuart123 , there's some more work to do on this branch.
> 
> The outstanding issue is that we only fetch the "next nonce" once, when a Safe is initially loaded in Fractal. If a transaction is created in Fractal (and not executed), then _another_ transaction is created, that second one will still be using the "next nonce" from the first one.
> 
> We need to figure out the cleanest way to re-fetch the "next nonce"...
> 
> * on an interval? blah
> * whenever a proposal is being created? maybe...
> * any other ideas?

The only other thing that comes to mind, is way more tedious but check nonce as things happen. 

- Loading the Form, check the nonce
- Clicking Submit, Check the nonce and show an error if nonce has changed or needs to be updated.

adamgall:
@tomstuart123 @Da-Colon 

Updated the code so that whenever pulling up the New Proposal page, we re-fetch the Safe Data to make sure the most recent "next nonce" is being used.

Still seeing some delay of about a minute after a new proposal is queued up before the nonce updates, but I think that's baked into the caching of our Safe Service client code.

This is ready for re-review.

Da-Colon (approved):
read through code, saw it live, Ship it!

Da-Colon:
Additional question. This covers "New Proposal", What about the other spots such as 'Settings' and 'Edit' and Treasury Pages (for Send Assets)?

adamgall:
> Additional question. This covers "New Proposal", What about the other spots such as 'Settings' and 'Edit' and Treasury Pages (for Send Assets)?

dammit

DarksightKellar:
Some observations:

- When there are multiple pending proposals, they *must* be executed in order of their nonces. Attempting to execute a nonce-7 proposal while a nonce-6 is pending fails with a non-descriptive error. If you go execute nonce-6, then nonce-7, that works fine. Tested with up to 4 pending proposals, and the behaviour is easily reproducible. This is likely expected behaviour though.

- I ran into an edge case where I created a proposal with nonce 7, and before it was visible on the UI, attempted to create another. The nonce assigned to it was 7. I went back, did a couple refreshes as I impatiently waited for the just-created proposal to show up. When it did I tried again, and still got nonce 7. (in the time it took me to think and type this out, without refreshing, the nonce is now 8, so, a timing thing, it appears. probably not worth digging into)

- I just used the 8th nonce to edit the dao name. waited for it to show up in proposal list, created new proposal, which was assigned nonce 9. DID NOT actually hit "create", went back to edit dao name (with previous edit still unexecuted). Also nonce 9.

For the most part these feel like edge case that won't matter in real world use.

adamgall:
Good notes, thanks @DarksightKellar 

> Some observations:
> 
> * When there are multiple pending proposals, they _must_ be executed in order of their nonces. Attempting to execute a nonce-7 proposal while a nonce-6 is pending fails with a non-descriptive error. If you go execute nonce-6, then nonce-7, that works fine. Tested with up to 4 pending proposals, and the behaviour is easily reproducible. This is likely expected behaviour though.

Yes this is a bug but out of scope for this PR. I've opened an issue: https://github.com/decentdao/fractal-interface/issues/1536

> * I ran into an edge case where I created a proposal with nonce 7, and before it was visible on the UI, attempted to create another. The nonce assigned to it was 7. I went back, did a couple refreshes as I impatiently waited for the just-created proposal to show up. When it did I tried again, and still got nonce 7. (in the time it took me to think and type this out, without refreshing, the nonce is now 8, so, a timing thing, it appears. probably not worth digging into)
>   EDIT: Sorry, I missed this part of your comment:
> 
> > Still seeing some delay of about a minute after a new proposal is queued up before the nonce updates
> 
> * I just used the 8th nonce to edit the dao name. waited for it to show up in proposal list, created new proposal, which was assigned nonce 9. DID NOT actually hit "create", went back to edit dao name (with previous edit still unexecuted). Also nonce 9.
> 
> For the most part these feel like edge case that won't matter in real world use.

These last two points, yeah, not ideal, but not sure how much more I can do about them.

adamgall:
To Do:

Make sure Nonce is updated on Send Tokens proposal creation, too.

tomstuart123 (approved):
Agreed with @adamgall  that we make the 'auto nonce update without refresh' in other areas of the app in another issue 

see here

This PR is approved

https://github.com/orgs/decentdao/projects/32/views/3?pane=issue&itemId=59938098

PR #1530, opened 2024-04-09
Fix DAO name doesn't update after edit

DarksightKellar:
This PR removes the local storage of the DAO name. Since this data apparently exists on a subgraph, it's auto-updated (after some time).

There isn't much of a tradeoff in completely removing the DAO name cache. There was a split second where, in the breadcrumbs, the name would be empty while being fetched. I have replaced this with the DAO address substring, so it's less jarring.

### DIFF
--- a/src/components/ui/page/Header/PageHeader.tsx
+++ b/src/components/ui/page/Header/PageHeader.tsx
@@ -11,6 +11,7 @@ import {
 } from '@chakra-ui/react';
 import { ReactNode, useEffect, useState } from 'react';
 import { DAO_ROUTES } from '../../../../constants/routes';
+import { createAccountSubstring } from '../../../../hooks/utils/useDisplayName';
 import { useFractal } from '../../../../providers/App/AppProvider';
 import { useNetworkConfig } from '../../../../providers/NetworkConfig/NetworkConfigProvider';
 import AddressCopier from '../../links/AddressCopier';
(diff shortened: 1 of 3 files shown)


adamgall (commented):
Looks like you left in code that fetches cached DAO name from local storage. Can we remove that, too?

DarksightKellar:
> Looks like you left in code that fetches cached DAO name from local storage. Can we remove that, too?

@adamgall ah yes, thanks for catching that. All done

adamgall (commented):
I would encourage you to keep going even deeper.

Does the logic in `useDAOName` make sense?
Do all the code paths get hit?
What are the actual differences between `useDAOName` and `useLazyDAOName`? Can we consolidate those functions into one thing?

Keep in mind my questions here are not rhetorical, I don't know the answers to them, I'm just superficially looking at the code and doing some searches for where these hooks/functions are being used and it seems ripe for some cleanup.

DarksightKellar:
> Does the logic in `useDAOName` make sense? Do all the code paths get hit? What are the actual differences between `useDAOName` and `useLazyDAOName`? Can we consolidate those functions into one thing?

Ah yeah I'm glad you brought this up, because I had similar questions too but focusing on understanding and fixing both code and bug ended up sidelining it, and I forgot.

I'd rather this potential cleanup not block this PR though. I would like to spend some time testing and working out what the hooks do differently, if they're both needed, and what a consolidated single hook might look like, but in a follow up PR.

tomstuart123:
From a quick look at our leading DAOs, everything seems to work as expected - no issue from my side

Da-Colon (approved):
Awesome, LGTM!

DarksightKellar:
Thanks for the reviews guys. New issue created for the refactoring here: https://github.com/decentdao/fractal-interface/issues/1531#issue-2236066624

Issue #1470, opened 2024-03-19
Create DAO - add snapshot space - character limit / helper text


tomstuart123:
Right now, in the app to create a dao AND in the settings page, it asks us in the helper text of the snapshot piece only to add the ENS name (e.g. example.eth)

![Image](https://github.com/decentdao/fractal-interface/assets/31554133/e0bc9da0-78ca-4da5-be0c-e463d3043cce)

To fix a bug way back, we made a change to have the entire snapshot URL be the link that is required (e.g. https://snapshot.org/#/example.eth)

**Issue #1** - currently the helper text is giving them wrong info. if they just put the ENS in, we won't recognise and display the snapshot proposals in the dao if the proposal passed

**Issue #2** - there is a character limit on the input field for the 'Create a DAO' flow. This limit probably worked for the prior 'just ENS'.... but for the entire URL that is now required, it doesn't work. Note this issue doesn't appear in the DAO settings area where you can also update the snapshot url

e.g. when copying Decent's snapshot URL in as a test (e.g. https://snapshot.org/#/decent-dao.eth) I get the below

![Image](https://github.com/decentdao/fractal-interface/assets/31554133/d6dc8540-d140-48b6-9c8a-62955a59384a)

adamgall:
We should also tackle the part where we're apparently requiring a full URL to Snapshot in that input box. That's dumb, all we need is the ENS name of the Snapshot Space. There will likely be some backwards compatibility code needed but whatever.

DarksightKellar:
> We should also tackle the part where we're apparently requiring a full URL to Snapshot in that input box. That's dumb, all we need is the ENS name of the Snapshot Space. There will likely be some backwards compatibility code needed but whatever.

@adamgall Does that mean the bug referenced in this issue's description is no longer present and the helper text ("example.eth") is, in fact, NOT wrong info?

DarksightKellar:
Actually what *is* the bug here? Besides removing the character limit on that input field, everything else seems to be working as far as I can tell with my limited context. How do I make sure I'm testing this right? cc @tomstuart123

tomstuart123:
Kelvin, Kirill and I spoke. 

We discussed going back to ENS for now as to @adamgall 's point

Then I'll a PR now in the backlog to explore a next implementation that supports a full URL (so snapshot testnet site can be linked too) as well as the ENS. We would do the input field validation. More to come

adamgall:
[...]

PR #1564, opened 2024-04-15
Snapshot space input bug

DarksightKellar:
In this PR, we have made the assumption that all snapshot spaces linked in Fractal will necessarily be production snapshots. We're also assuming that the "snapshot space" input when creating or editing a DAO will always be an ENS name in the format `example.eth`.

To that end, the only url used to setup Apollo client has been set to `https://hub.snapshot.org/graphql`. This PR also fixes the broken snapshot link on the DAO info card.

NOTE: this will be impossible to test on testnets if the linked snapshot space is a testnet one

~To consider: Should we hard enforce a NO-FULL-URL policy in the "snapshot space" inputs, IOW, ONLY accept `abc.eth` formats as inputs?~
In addition, by utilising the concept of normalising ENS names (thanks for the ref @adamgall!), we are able to reject http-urls completely, and really any input contains improperly formatted, normalise-able ENS names.

Rejection happens simply by disabling the "next" (or "propose changes") button for present but invalid input.


adamgall:
> To consider: Should we hard enforce a NO-FULL-URL policy in the "snapshot space" inputs, IOW, ONLY accept abc.eth formats as inputs?

FWIW, ENS names are not as simple as "abc.eth". There are a lot of forms they can take.

Play around with this tool: https://tools.ens.domains/check
Specifically, keep an eye on the "valid / invalid" tag that shows up in the "Normalized" section.

I think if we were to do what you're considering (which I do think we should do!), we should copy whatever is happening there. I would expect this is a good place to start: https://docs.ens.domains/resolution/names#normalize

adamgall:
@DarksightKellar I assume this PR close issue #1470?

src/components/ui/badges/Snapshot.tsx
```diff
@@ -6,9 +6,12 @@ interface Props extends ButtonProps {
 }
 
 export default function Snapshot({ snapshotURL, mt }: Props) {
+  const url = snapshotURL.includes('testnet')
+    ? snapshotURL
+    : `https://snapshot.org/#${snapshotURL}`;
```
adamgall, line 11: Is this for some backwards compatibility reasons?

DarksightKellar: Ah yes, I meant to add a comment here. But yeah, backward compatibility is the goal there

adamgall: Can you explain how this backwards compatibility works?

DarksightKellar: It's meant to maintain existing `snapshotURL`s that were entered and stored as full urls, instead of prepending the snapshot url to it, which would lead to an incorrect url: `https://snapshot.org/#https://snapshot.org/#example.eth`.

Aaaand now that I'm explaining it, I realise that initial assumption I'd made that ens names worked on mainnet continues to to plague me, and this won't work for fully qualified urls on prod.

Gonna update to `snapshotURL.startsWith('https://')`. So existing fully qualified urls of any form will be left alone, and ENS names entered after this PR will lead to the correct url.

At some point hopefully all DAOs will have updated their snapshot space and this can be removed. Orrrr we could not be backward compatible and communicate the necessary change with current users. Doubt that's the move to make though.

Have I missed anything?

adamgall: We really only have _1_ testnet user, and they don't have a Snapshot space set. So I would say we do not need to be backwards compatible.

adamgall: Ah right, mainnet users too. Well thankfully our only mainnet user that has their Snapshot set is us! And we can sneak in a fix for that during our initial set of proposals (cc @tomstuart123).

So I would say it's not necessary to build any backwards compatibility into this.

DarksightKellar: Nice. Agreed then.

src/components/DaoCreator/formComponents/EstablishEssentials.tsx
```diff
@@ -57,6 +64,18 @@ export function EstablishEssentials(props: ICreationStepProps) {
...
 
+  const handleSnapshotSpaceChange = (value: string) => {
+    setFieldValue('essentials.snapshotURL', value, true);
+    try {
+      ens_normalize(value);
+      setSnapshotSpaceValid(true);
+    } catch (error) {
+      console.log(error);
+      // If there's no input in the snapshot URL field, we don't need to check if it's valid
+      setSnapshotSpaceValid(!!value ? false : true);
```
adamgall, line 75: Could the conditional check here happen up before the `try`, and early exiting the function if there's nothing in the value field?

For example

```ts
const handleSnapshotSpaceChange = (value: string) => {
    setFieldValue('essentials.snapshotURL', value, true);
    
    if (!value) {
        setSnapshotSpaceValid(true);
        return;
    }
    
    try {
        ens_normalize(value);
        setSnapshotSpaceValid(true);
    } catch (error) {
        console.log(error);
        setSnapshotSpaceValid(false);
    }
}
```

I think I prefer this style of coding. Does it do the same thing though?

src/components/pages/DaoSettings/components/Metadata/index.tsx
```diff
@@ -10,7 +11,6 @@ import { createAccountSubstring } from '../../../../../hooks/utils/useDisplayNam
 import { useFractal } from '../../../../../providers/App/AppProvider';
 import { useNetworkConfig } from '../../../../../providers/NetworkConfig/NetworkConfigProvider';
 import { ProposalExecuteData } from '../../../../../types';
-import { couldBeENS } from '../../../../../utils/url';
```
adamgall, line 13: Looks like that helper function is used in a bunch of places in the app, blah.

It could be replaced with that `ens_normalize` function, right?

If so, can you please create an issue to replace `couldBeENS` with `ens_normalize`?

adamgall (changes requested):
Looking good! Just one small syntax suggestion, and some questions.

A question I didn't leave in the code-level review, is: Could we update the state variable name `daoSnapshotURL` to something more accurate now, like `daoSnapshotENS`? Same with changing `snapshotURL` to `snapshotENS` on the `DAOEssentials` type.

IMO that seems appropriate for this PR.

tomstuart123 (approved):
I just tested this with ENS at this DAO - https://deploy-preview-1564.app.dev.fractalframework.xyz/home?dao=eth:0xDBC4Ae5c0fdC921ee08B44031D6DA081De1d6485

Added Decent DAO's snapshot and then swapped it for Ehtlizards's snapshot.

Both worked. Feeling good. Thanks @DarksightKellar

src/components/DaoCreator/formComponents/EstablishEssentials.tsx
DarksightKellar: It does do the same thing. Early return looks nice. Will update and test

adamgall (approved):
@DarksightKellar doing a search in the codebase for "snapshoturl" still gives a bunch of hits, most of which can be updated.

The only place where we'll want to keep the "snapshoturl" nomenclature is right at the edges:
- outgoing: when building a transaction which includes the call to the "key value pairs" contract
- incoming: when consuming the "key value pairs" subgraph

Giving this an approval anyway, but it'd be nice to see the vast majority of those variables updated.

DarksightKellar:
Did a couple more rounds of quick tests after some cleanup. Merging now

PR #1584, opened 2024-04-22
`Reskin` | DAO hierarchy

DarksightKellar:
This PR updates the DAO hierarchy page:

Screenshot 2024-04-22 at 17 34 56

A lot of no longer needed components and logic have been nuked. Also, the info card component has been split into dedicated components for the DAO home page and the hierarchy node info card.

I've also lightly touched the breadcrumbs. Realised there's a dedicated task for it, stopped at a good stopping point. I think it's really only the divider that needs to be reskinned. I'm hoping we could create a dedicated divider component that can be reused everywhere, as the reskinned divider is more that just a single-coloured line.


adamgall (changes requested):
Looking real nice, but one requested change.

Screenshot 2024-04-22 at 5 26 29 PM

I would think (without looking at the Figma), that the green pill highlight should only wrap around the address and "copy" icon, not stretch to the width of the DAO Name.

edit: Just looked at the Figma (through the attached Issue), no specification on that highlight.

@xraystyle1980 do you have any input here?

DarksightKellar:
> Looking real nice, but one requested change.
> 
> This is what the address looks like when I hover over it with my mouse:
> 
> Screenshot 2024-04-22 at 5 26 29 PM
> I would think (without looking at the Figma), that the green pill highlight should only wrap around the address and "copy" icon, not stretch to the width of the DAO Name.
> 
> edit: Just looked at the Figma (through the attached Issue), no specification on that highlight.
> 
> @xraystyle1980 do you have any input here?

@adamgall Also see [this message](https://decentdao.slack.com/archives/C06QZTUTKRC/p1713544491754239) and [this conversation](https://decentdao.slack.com/archives/C06QZTUTKRC/p1713544630828249) on the builders channel for some more context.

It would seem that having the highlight wrap around just the address and copy icon is the correct move. I'm gonna go ahead and push an update to that effect. I'll raise another issue for the rest of the discussion, to tackle at another time.

adamgall (approved):

PR #1586, opened 2024-04-22
Don't crash with malformed proposal metadata

adamgall:
Wrap the JSON.parse in a try catch, so as to not break when metadata is malformed.

### Testing

Sepolia DAO sep:0x0FAB989675bcbE3adFBAE4eC8E54061CcCE85020 includes a proposal with bad JSON (prop # 29).

See how this DAO doesn't load proposal 29 (or below) on dev, but will load properly (but not display a proposal title, description, or url) for the malformed proposal 29 on this PR.


DarksightKellar (approved):
Tested. lgtm.

Should the offending proposal be displayed slightly differently (messaging and/or UI) to make it more obvious that it was malformed, or am I overthinking this?

PR #1609, opened 2024-04-26
Bugfix/abi encoding

adamgall:
i think i found

- 1 actual bug
- 2 instances where the code is misleading

in various places where we do ABI encoding

### DIFF
--- a/src/models/AzoriusTxBuilder.ts
+++ b/src/models/AzoriusTxBuilder.ts
@@ -348,7 +348,7 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
     const encodedInitTokenData = defaultAbiCoder.encode(
       ['uint32', 'address', 'address', 'address', 'uint256'],
       [
-        0, // deadlineBlock. do we capture this in the UI?
+        0, // deadlineBlock. We don't capture this in the UI. 0 means no deadline to claim.
         this.safeContract.address,
         this.parentTokenAddress,
         this.predictedTokenAddress,
--- a/src/models/AzoriusTxBuilder.ts
+++ b/src/models/AzoriusTxBuilder.ts
@@ -25,7 +25,6 @@ import { generateContractByteCodeLinear, generateSalt } from './helpers/utils';
 
 export class AzoriusTxBuilder extends BaseTxBuilder {
   private readonly safeContract: GnosisSafeL2;
-  private readonly predictedSafeAddress: string;
 
   private encodedSetupTokenData: string | undefined;
   private encodedSetupERC20WrapperData: string | undefined;
@@ -54,7 +53,6 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
     azoriusContracts: AzoriusContracts,
     daoData: AzoriusERC20DAO | AzoriusERC721DAO,
     safeContract: GnosisSafeL2,
-    predictedSafeAddress: string,
     parentAddress?: string,
     parentTokenAddress?: string,
   ) {
@@ -68,7 +66,6 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
     );
 
     this.safeContract = safeContract;
-    this.predictedSafeAddress = predictedSafeAddress;
 
     this.tokenNonce = getRandomBytes();
     this.claimNonce = getRandomBytes();
--- a/src/models/AzoriusTxBuilder.ts
+++ b/src/models/AzoriusTxBuilder.ts
@@ -346,8 +346,9 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
   private setEncodedSetupTokenClaimData() {
     const azoriusGovernanceDaoData = this.daoData as AzoriusERC20DAO;
     const encodedInitTokenData = defaultAbiCoder.encode(
-      ['address', 'address', 'address', 'uint256'],
+      ['uint32', 'address', 'address', 'address', 'uint256'],
       [
+        0, // deadlineBlock. do we capture this in the UI?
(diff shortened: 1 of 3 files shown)


src/models/AzoriusTxBuilder.ts
```diff
@@ -25,7 +25,6 @@ import { generateContractByteCodeLinear, generateSalt } from './helpers/utils';
 
 export class AzoriusTxBuilder extends BaseTxBuilder {
   private readonly safeContract: GnosisSafeL2;
-  private readonly predictedSafeAddress: string;
```
adamgall, line 28: Unused, removing.

src/models/AzoriusTxBuilder.ts
```diff
@@ -54,7 +53,6 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
     azoriusContracts: AzoriusContracts,
     daoData: AzoriusERC20DAO | AzoriusERC721DAO,
     safeContract: GnosisSafeL2,
-    predictedSafeAddress: string,
```
adamgall, line 57: Unused, removing.

src/models/AzoriusTxBuilder.ts
```diff
@@ -68,7 +66,6 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
     );
 
     this.safeContract = safeContract;
-    this.predictedSafeAddress = predictedSafeAddress;
```
adamgall, line 71: Unused, removing.

src/models/TxBuilderFactory.ts
```diff
@@ -130,7 +130,6 @@ export class TxBuilderFactory extends BaseTxBuilder {
       this.azoriusContracts!,
       this.daoData as AzoriusERC20DAO,
       this.safeContract!,
-      this.predictedSafeAddress!,
```
adamgall, line 133: Unused, removing.

src/models/AzoriusTxBuilder.ts
```diff
@@ -349,8 +346,9 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
   private setEncodedSetupTokenClaimData() {
     const azoriusGovernanceDaoData = this.daoData as AzoriusERC20DAO;
     const encodedInitTokenData = defaultAbiCoder.encode(
-      ['address', 'address', 'address', 'uint256'],
+      ['uint32', 'address', 'address', 'address', 'uint256'],
       [
+        0, // deadlineBlock. do we capture this in the UI?
```
adamgall, line 351: This seems like an actual bug. The contract is expecting a `uint32` before the rest of those parameters, and we weren't sending one in. Is this code path just rarely (ever) used in the app? I need to do some manual testing still.

src/models/FreezeGuardTxBuilder.ts
```diff
@@ -217,7 +217,7 @@ export class FreezeGuardTxBuilder extends BaseTxBuilder {
       'setUp',
       [
         ethers.utils.defaultAbiCoder.encode(
-          ['uint256', 'uint256', 'address', 'address', 'address'],
+          ['uint32', 'uint32', 'address', 'address', 'address'],
```
adamgall, line 220: The contract is expecting `uint32`s here, not `uint256`s.

Does this currently work as expected? I bet so, because we have tested freeze guard things before.

Will it be a problem to change this? Hope not.

`subDaoData.timelockPeriod` and `subDaoData.executionPeriod` are both typed as `bigint` though, which also "seems wrong" to me, because those things are `uint32`s on the contract. They don't need to be `bigint`s, they can just be `number`s. Updating those types makes this PR _much_ hairier though.

I need to do some manual testing to make sure this change still works as intended, just like this.

src/models/FreezeGuardTxBuilder.ts
```diff
@@ -231,19 +231,14 @@ export class FreezeGuardTxBuilder extends BaseTxBuilder {
...
 
   private setFreezeGuardCallDataAzorius() {
-    const subDaoData = this.daoData as SubDAO;
-
     this.freezeGuardCallData = AzoriusFreezeGuard__factory.createInterface().encodeFunctionData(
       'setUp',
       [
         ethers.utils.defaultAbiCoder.encode(
-          ['address', 'address', 'address', 'address', 'uint256'],
+          ['address', 'address'],
```
adamgall, line 238: We were straight up "encoding too much", here. The contract only expects two `address`es. So, slimmed it down.

I bet this still works currently because the smart contract _probably_ just discards everything after the first two address parameters.

Need to do some manual testing on this too of course.

src/models/AzoriusTxBuilder.ts
adamgall: Yeah, sure enough this is BROKEN. Can't execute a proposal in which we're creating a subdao with a token claim for parent holders.

adamgall: And I AM able to execute the same proposal with my changes here in place. Nice.

src/models/FreezeGuardTxBuilder.ts
adamgall: Did some more manual testing, this new code works as expected.

adamgall: Did manual testing, this works as expected.

src/models/FreezeGuardTxBuilder.ts
adamgall: Did manual testing, this works as expected.

PR #1638, opened 2024-05-05
Update VotesERC20Wrapper

adamgall:
Main thing that this PR does is:

- Remove all imports of the `VotesERC20Wrapper` TypeChain imports from `fractal-contracts` in the codebase.
- In favor using typed contract objects from viem using `getContract()`.

Additionally, this PR _starts_ to do some refactoring cleanup, removing some layers of abstraction:

- Remove `votesERC20WrapperMasterCopyContract` "asSigner / asProvider" object from `useSafeContracts`.
- Remove `votesERC20WrapperMasterCopyContract` from `AzoriusContracts` type and `FractalContracts` type.

There was some other light cleanup, which I'll document in some inline PR review notes.


src/assets/abi/GnosisSafeL2.ts
```diff
@@ -1135,4 +1135,6 @@ export default [
     stateMutability: 'payable',
     type: 'receive',
   },
-];
+] as const;
```
adamgall, line 1138: Adding `as const` here is necessary for viem's typescript support.

https://viem.sh/docs/typescript#type-inference

This small change resulted in some type cleanup in the one place where this ABI was being used!

src/models/helpers/safeData.ts
```diff
@@ -32,13 +32,13 @@ export const safeData = async (
...
-      1, // Threshold
+      signers.map(signer => getAddress(signer)),
+      1n, // Threshold
       zeroAddress,
       zeroHash,
-      fallbackHandler,
+      getAddress(fallbackHandler),
       zeroAddress,
-      0,
+      0n,
```
adamgall, line 41: As mentioned in the previous comment, these needed to be updated because due to type errors, after adding `as const` to the GnosisSafeL2 ABI! So nice

src/components/ui/modals/DelegateModal.tsx
```diff
@@ -44,7 +44,7 @@ export function DelegateModal({ close }: { close: Function }) {
       validAddress = getAddress(await signer.resolveName(values.address));
     }
     const votingTokenContract =
-      baseContracts.votesERC20WrapperMasterCopyContract.asSigner.attach(votesTokenContractAddress);
+      baseContracts.votesTokenMasterCopyContract.asSigner.attach(votesTokenContractAddress);
```
adamgall, line 47: This is a funny little update, worth narrowing in on. 

Basically, why were we previously using the `votesERC20WrapperMasterCopyContract` object/type here? 

I don't think it matters which contract object we use, because we're doing an `.attach` (which forces a new object instance of the same type but at a new address), and then the only contract function being called from this new `votingTokenContract` object is a function that exists on _both_ the `VotesToken` _and_ the `VotesERC20Wrapper` ABIs.

So, I don't really think it mattered which existing "baseContract" object we started from (between the two in this changeset), but I do think it makes more intentional sense to use the one that I changed to here.

src/hooks/utils/useTransaction.ts
```diff
@@ -76,7 +90,65 @@ const useTransaction = () => {
...
+          if (error.code === 4001) {
+            toast.error(t('errorUserDeniedTransaction'));
+            return;
+          }
+
+          toast.error(t('errorGeneral', { ns: 'common' }));
+        });
+    },
+    [t, publicClient],
+  );
```
adamgall, line 149: Added a new function here for the generic `contractCall`, but supporting viem-style contract calls. Was too messy to try to combine both ethers- and viem-style calls into one function. Eventually, the other function will be removed after we get through _all of the PRs_ that this one is kicking off.

src/components/ui/modals/UnwrapToken.tsx
```diff
@@ -1,31 +1,30 @@
...
 export function UnwrapToken({ close }: { close: () => void }) {
   const { governance, governanceContracts } = useFractal();
   const azoriusGovernance = governance as AzoriusGovernance;
-  const signer = useEthersSigner();
   const { address: account } = useAccount();
   const baseContracts = useSafeContracts();
   const { loadERC20TokenAccountData } = useERC20LinearToken({ onMount: false });
 
-  const [contractCall, pending] = useTransaction();
+  const [, pending, contractCallViem] = useTransaction();
```
adamgall, line 27: Here is an instance where we're using the new `contractCallViem` function instead of the original one, because we have a new viem-style contract object.

src/components/ui/modals/WrapToken.tsx
```diff
@@ -32,7 +33,7 @@ export function WrapToken({ close }: { close: () => void }) {
   const baseContracts = useSafeContracts();
 
   const { loadERC20TokenAccountData } = useERC20LinearToken({ onMount: false });
-  const [contractCall, pending] = useTransaction();
+  const [, pending, contractCallViem] = useTransaction();
```
adamgall, line 36: Here's another place where the new function is used.

Da-Colon (approved):
Awesome, LGTM, Code reviewed looks good to me, not Live QAed

PR #1641, opened 2024-05-05
Skip useDelegateVote

adamgall:
Please review and merge https://github.com/decentdao/fractal-interface/pull/1639 first.

### DIFF
--- a/src/hooks/DAO/useDelegateVote.ts
+++ b/src/hooks/DAO/useDelegateVote.ts
@@ -1,4 +1,3 @@
-import { VotesERC20 } from '@fractal-framework/fractal-contracts';
 import { useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import { LockRelease } from '../../assets/typechain-types/dcnt';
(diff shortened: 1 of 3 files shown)


src/hooks/DAO/useDelegateVote.ts
```diff
@@ -16,7 +15,7 @@ const useDelegateVote = () => {
       successCallback,
     }: {
       delegatee: string;
-      votingTokenContract: VotesERC20 | LockRelease;
+      votingTokenContract: LockRelease;
```
adamgall, line 18: Can downscope the type here to just LockRelease, because we'll skip over this `useDelegateVote` hook completely from the `DelegateModal` and just call `useTransaction` directly.

adamgall: @DarksightKellar I also do not know why this `useDelegateVote` hook exists!! I plan on getting rid of it entirely when I tackle the `LockRelease` typechain type.

DarksightKellar: Ah good to know!

DarksightKellar (approved):

PR #1645, opened 2024-05-05
Remove VotesERC20 from AzoriusTxBuilder

adamgall:
Please review and merge https://github.com/decentdao/fractal-interface/pull/1644 first.

### DIFF
--- a/src/helpers/crypto.ts
+++ b/src/helpers/crypto.ts
@@ -12,6 +12,8 @@ import {
   bytesToBigInt,
   Hex,
   isHex,
+  encodeFunctionData,
+  Abi,
 } from 'viem';
 import { ContractConnection } from '../types';
 import { MetaTransaction, SafePostTransaction, SafeTransaction } from '../types/transaction';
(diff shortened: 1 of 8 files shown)


src/helpers/crypto.ts
```diff
@@ -161,20 +163,18 @@ export const buildSafeAPIPost = async (
   };
 };
 
-export const buildContractCall = (
-  contract: Contract,
-  method: string,
-  params: any[],
+const finishBuildingConractCall = (
```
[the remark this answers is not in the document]
adamgall: gah. i almost just want to leave it as is for now, because fixing this will cause me to need to rebase all of the branches above this one 😅 

i plan on consolidating this back into one function anyway, once the the ethers-style function can be removed. I'll deal with it then.

src/helpers/crypto.ts
```diff
@@ -184,6 +184,45 @@ export const buildContractCall = (
   );
 };
 
+export const buildContractCall = (
```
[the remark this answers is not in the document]
adamgall: Sure is!

src/models/AzoriusTxBuilder.ts
```diff
@@ -547,6 +555,10 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
   }
 
   private setContracts() {
+    if (!this.predictedTokenAddress) {
+      return;
+    }
```
Da-Colon, line 560: Is this defined when existing token is used for DAO Creation? Ie set to the existing token's address

adamgall: TBH not sure. I assumed (perhaps incorrectly?) that this check here was safe, because if you look at the old code on line 561, we were forcing the value to be valid (`this.predictedTokenAddress!`).

I could just put that `!` back in on new line 574 and remove this conditional early exit, but I don't like `!` in typescript.

I suppose I should go and manually test this out, huh

adamgall: In the case when a user brings their own token to new Azorius DAO creation, `this.predictedTokenAddress` gets set to that existing token address. So, this function executes as expected. Thanks for the call out here.

PR #1650, opened 2024-05-05
Remove key value pairs contract from base contracts

adamgall:
Please review and merge https://github.com/decentdao/fractal-interface/pull/1645 first.

### DIFF
--- a/src/models/DaoTxBuilder.ts
+++ b/src/models/DaoTxBuilder.ts
@@ -1,7 +1,8 @@
 import { ethers } from 'ethers';
-import { PublicClient, zeroAddress } from 'viem';
+import { PublicClient, getAddress, zeroAddress } from 'viem';
+import KeyValuePairsAbi from '../assets/abi/KeyValuePairs';
 import { GnosisSafeL2 } from '../assets/typechain-types/usul/@gnosis.pm/safe-contracts/contracts';
-import { buildContractCall, encodeMultiSend } from '../helpers';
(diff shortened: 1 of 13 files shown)


src/models/DaoTxBuilder.ts
```diff
@@ -42,6 +45,7 @@ export class DaoTxBuilder extends BaseTxBuilder {
     createSafeTx: SafeTransaction,
     safeContract: GnosisSafeL2,
     txBuilderFactory: TxBuilderFactory,
+    keyValuePairsAddress: string,
```
Da-Colon, line 48: I assume eventually these will be updated to the `Address` type

Da-Colon (approved):
One comment, but approved otherwise

src/models/DaoTxBuilder.ts
adamgall: Oh yeah, I've been a bit sloppy in these builder classes with Address/string types.

My intention is that, once this current push to get TypeChain OUTTA HERE is finished, I'll go back through and grep the codebase for `getAddress`, and one by one, remove them, changing the type of the `string`s that they're wrapping with `Address`, fixing whatever that breaks down the stack with the same approach, until eventually all of the things that should be `Addresses` actually are, and we will have minimal/no `getAddress` wrappers anymore.

Da-Colon: ah cool cool, ignore my comment in a few PRs up lol

DarksightKellar (approved):

PR #1660, opened 2024-05-06
Remove the safeSingletonContract property from FractalContracts and BaseContracts interfaces

adamgall:
Please review and merge https://github.com/decentdao/fractal-interface/pull/1658 first.

### DIFF
--- a/src/hooks/DAO/useDeployAzorius.ts
+++ b/src/hooks/DAO/useDeployAzorius.ts
@@ -1,8 +1,9 @@
 import { useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
-import { getAddress, isHex } from 'viem';
+import { encodeFunctionData, getAddress, isHex } from 'viem';
 import { usePublicClient } from 'wagmi';
+import GnosisSafeL2Abi from '../../assets/abi/GnosisSafeL2';
 import { DAO_ROUTES } from '../../constants/routes';
 import { TxBuilderFactory } from '../../models/TxBuilderFactory';
 import { useFractal } from '../../providers/App/AppProvider';
@@ -29,6 +30,7 @@ const useDeployAzorius = () => {
       keyValuePairs,
       fractalRegistry,
       safeFactory,
+      safe: safeSingleton,
       zodiacModuleProxyFactory,
     },
     addressPrefix,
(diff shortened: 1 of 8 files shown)


src/hooks/DAO/useDeployAzorius.ts
```diff
@@ -110,14 +111,12 @@ const useDeployAzorius = () => {
...
       });
 
-      const encodedAddOwnerWithThreshold =
-        safeSingletonContract.asProvider.interface.encodeFunctionData('addOwnerWithThreshold', [
-          multiSendContract.asProvider.address,
-          1,
-        ]);
-      if (!isHex(encodedAddOwnerWithThreshold)) {
-        return;
-      }
```
Da-Colon, line 120: the `isHex` function seems to be off and on used in conjunction with the `encodeFunctionData`. is there a reason its being used in some places and not others?

src/models/helpers/safeData.ts
```diff
@@ -62,10 +61,7 @@ export const safeData = async (
     bytecodeHash: keccak256(
       encodePacked(
         ['bytes', 'uint256'],
-        [
-          safeFactoryContractProxyCreationCode,
-          hexToBigInt(getAddress(safeSingletonContract.address)),
-        ],
```
Da-Colon, line 68: Why the removal of the `getAddress` function here?

Da-Colon (commented):
Two questions and looks like github is picking up on a lint error in `AzoriusTxBuilder`

src/hooks/DAO/useDeployAzorius.ts
adamgall: viem's `encodeFunctionData()` will always return a "`Hex`" string, so no need to use the `isHex` function on data returned from that function.

It was needed when using ethers `interface.encodeFunctionData()`, because that function just returns a plain `string`, and typescript doesn't allow plain `strings` to be passed into viem functions that are expecting `Hex`.

As we progress through this refactor, we'll be removing all of these `isHex` checks, because we won't be using ethers any longer!

src/models/helpers/safeData.ts
adamgall: Presumably because the `safeSingletonContract` type went from an ethers contract object to a viem contract object.

ethers contract.address is a `string`
viem contract.address is an `Address`

Da-Colon (approved):

DarksightKellar (approved):

PR #1669, opened 2024-05-07
Remove linear erc721 voting

adamgall:
Please review and merge https://github.com/decentdao/fractal-interface/pull/1668 first.

### DIFF
--- a/src/hooks/DAO/proposal/useUserERC721VotingTokens.ts
+++ b/src/hooks/DAO/proposal/useUserERC721VotingTokens.ts
@@ -1,11 +1,8 @@
-import {
-  Azorius,
-  LinearERC721Voting__factory,
-  LinearERC721Voting,
-} from '@fractal-framework/fractal-contracts';
+import { Azorius } from '@fractal-framework/fractal-contracts';
 import { useState, useEffect, useCallback } from 'react';
-import { erc721Abi, getAddress, getContract } from 'viem';
+import { GetContractReturnType, PublicClient, erc721Abi, getAddress, getContract } from 'viem';
 import { usePublicClient } from 'wagmi';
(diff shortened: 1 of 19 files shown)


src/hooks/DAO/proposal/useUserERC721VotingTokens.ts
```diff
@@ -173,14 +179,18 @@ export default function useUserERC721VotingTokens(
...
               totalTokenAddresses.push(tokenAddress);
               totalTokenIds.push(tokenId);
               if (_proposalId) {
-                const tokenVoted = await votingContract!.hasVoted(
-                  _proposalId,
-                  tokenAddress,
-                  tokenId,
-                );
+                const tokenVoted = await votingContract.read.hasVoted([
+                  Number(_proposalId),
```
Da-Colon, line 190: Can proposal Id ever be anything but a number? possibly so? just wanted to ask anyway

Da-Colon (commented):
👍

Da-Colon (approved):

src/hooks/DAO/proposal/useUserERC721VotingTokens.ts
adamgall: Good call out. Found some bugs here. Fixing now.

adamgall: Well, fixed the bugs (5fade89d87a6765fbe6dbc8f58f5d0283f2270b5).

I also moved where this cast was happening, but TL;DR is that by the time this code here is executing, `_proposalId`, although a string, is a number. I certainly don't love this. The deeper issue is that we're sharing so many types/interfaces between erc20/erc721/multisig/snapshot proposals, when we really shouldn't be.

PR #1757, opened 2024-05-10
Stop masking session recording text and inputs

adamgall:
It's annoying not being able to see the toasts that are appearing in Sentry's session replays. I consider this change acceptable because _all of the data on these screens are public on the blockchain anyway_. We should be able to see the errors that our app is throwing.


Da-Colon (approved):
Make sense

PR #1761, opened 2024-05-10
Set and unset the page title when a DAO is mounted and unmounted

adamgall:
Closes https://github.com/decentdao/fractal-interface/issues/1691

- Adds DAO Name to page title when a DAO is loaded
- Removes DAO Name from page title when DAO is unloaded


Da-Colon (approved):
So much simpler without Nextjs.

nicolaus-sherrill (approved):

PR #1766, opened 2024-05-10
DaoInfoHeader hook cleanup

adamgall:
Type cleanup from moving some hooks around and not introducing `undefined` when not needed


src/components/pages/DaoDashboard/Info/DaoInfoHeader.tsx
```diff
@@ -39,14 +40,7 @@ export function DaoInfoHeader() {
...
           <InfoBox>
-            {/* TODO: The call to `useFractal` can probably be now safely moved into `DAOInfoCard`, so this prop passing can be trimmed out */}
-            <DAOInfoCard
-              parentAddress={node.nodeHierarchy.parentAddress || undefined}
-              node={node}
-              childCount={node.nodeHierarchy.childNodes.length}
-              freezeGuard={guard}
-              guardContracts={guardContracts}
-            />
+            <DAOInfoCard />
```
adamgall, line 43: Removing the params from `DAOInfoCard` is what kicked off all of the changes in this PR.

adamgall (commented):
A lot of the changes in this PR are the result of simplifying the type of `parentAddress` (which is passed around throughout these various components` from `string | null | undefined` to `string | null`

src/components/ui/cards/DAOInfoCard.tsx
```diff
@@ -1,39 +1,29 @@
...
  */
-export function DAOInfoCard({
-  parentAddress,
-  node,
-  childCount,
-  freezeGuard,
-  guardContracts,
-  ...rest
-}: InfoProps) {
+export function DAOInfoCard({ ...rest }: FlexProps) {
```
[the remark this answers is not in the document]
adamgall: Updated

Da-Colon (approved):
I don't know if any of these places do it but there were some places where 'undefined' means something different than 'null'. Just want to make sure that this change doesn't affect any areas that this might be the case. otherwise looks good to me


`undefined`: has not be defined.
`null`: intentionally say that it won't be defined.

adamgall:
> I don't know if any of these places do it but there were some places where 'undefined' means something different than 'null'. Just want to make sure that this change doesn't affect any areas that this might be the case. otherwise looks good to me
> 
> `undefined`: has not be defined. `null`: intentionally say that it won't be defined.

@Da-Colon do you see any places in this PR where the undefined/null behavior has changed? I do understand that they are separate types or whatever, but it didn't seem like they were being treated differently in the codebase.

We were only ever checking `if (property) ...` or `if (!property) ...` or `if (!!property) ...`

In all of these cases, `null` and `undefined` behave the same way. So all I did here was remove `undefined` as an option for these types.

Da-Colon:
> > I don't know if any of these places do it but there were some places where 'undefined' means something different than 'null'. Just want to make sure that this change doesn't affect any areas that this might be the case. otherwise looks good to me
> > `undefined`: has not be defined. `null`: intentionally say that it won't be defined.
> 
> @Da-Colon do you see any places in this PR where the undefined/null behavior has changed? I do understand that they are separate types or whatever, but it didn't seem like they were being treated differently in the codebase.
> 
> We were only ever checking `if (property) ...` or `if (!property) ...` or `if (!!property) ...`
> 
> In all of these cases, `null` and `undefined` behave the same way. So all I did here was remove `undefined` as an option for these types.
> 
> That was my intention at least

I don't have anywhere in particular in mind, Just wanted to state that those differences were introduced in the app. I didn't see any in the PR files, but if you didn't see any as you did this it should be good to go.

src/components/ui/cards/DAOInfoCard.tsx
```diff
@@ -1,39 +1,29 @@
...
 import { ManageDAOMenu } from '../menus/ManageDAO/ManageDAOMenu';
 
-export interface InfoProps extends FlexProps {
-  parentAddress?: Address;
-  node?: FractalNode;
-  childCount?: number;
-  freezeGuard?: FreezeGuard;
-  guardContracts?: FractalGuardContracts;
-}
-
```
DarksightKellar, line 22: 🤌🏾

DarksightKellar (approved):
🚀

PR #1779, opened 2024-05-14
`Reskin`- hover effect on parent link

### LINKED ISSUE #1714: Hover effect on parent dao breadcrumb
On child safe home page, the hover effect for the parent dao breadcrumb triggers across the entire content window width instead of just the text. Is this intentional? I kinda like it, but we would need to cook up a different design to give that space more presence.

Screenshot 2024-05-03 at 3 02 57 PM

_Originally posted by @nicolaus-sherrill in https://github.com/decentdao/fractal-interface/issues/1578#issuecomment-2093685918_

mudrila:
This PR contains following changes:

- Adjust `ParentLink` width to `fit-content` so that it's hover effect doesn't take whole row
- Slightly improve `InfoGovernance` - for long values `justify-content` wasn't working as expected and text was aligned to the left side - fix this

Closes #1714


DarksightKellar (approved):

tomstuart123 (approved):
nice, works better for me

adamgall (approved):

PR #1787, opened 2024-05-14
Semver and tagged releases

adamgall:
Closes https://github.com/decentdao/fractal-interface/issues/1659

Adds semantic versioning back into the `package.json`, which is then fed through to the app's footer.

The link created will then link to the Release on GitHub

Screenshot 2024-05-14 at 4 40 22 PM


Da-Colon (approved):
🫡

Issue #1793, opened 2024-05-15
Quorum bug - Proposal details


tomstuart123:
In testing a PR, I noticed a bug in the reskin proposal details

As you can see in screenshot below, this proposal passed with me holding 10 tokens out of 14.  This should have caused the quorum bar on the right to be fully filled (i.e show that quorum is passed). Instead the quorum bar remained stuck at 4% (which is the % quorum of the DAO; not its current quorum). 

![Image](https://github.com/decentdao/fractal-interface/assets/31554133/08d4f9a0-fb18-45d3-beeb-231a87d0bcec)

The proposal has since gone through timelock, executable and exceucted, so that is all functioning correctly.

![Image](https://github.com/decentdao/fractal-interface/assets/31554133/5744f7f7-136f-4060-8020-23d99175ac71)

i.e. its just the quorum bar in proposal details. 

To test yourself DAO link here - https://deploy-preview-1578.app.dev.fractalframework.xyz/proposals/1?dao=sep:0x7E9308800b7F3234982867b95481C8242C8AbFD1

adamgall:
@tomstuart123 the UI says you held 12 tokens, not 10. can you confirm your voting weight here?

tomstuart123:
you are correct it was 12 tokens out of a total of 14 (2 were in the treasury)

adamgall:
thanks for confirming @tomstuart123 

Is there language anywhere describing how exactly that quorum bar is supposed to work? Both in terms of "how full it is" and the various numbers that are displayed within it and underneath it?

tomstuart123:
Note, @decentdao/design can you check the figma designs track this. I think we may have missed when we updated this and need updates to the figma

tomstuart123:
More clarity on my issue @DarksightKellar / @adamgall 

I'm getting two visualisations of the quorum component within the same root reskin branch:

version 1 sepolia token voting dao
--> https://deploy-preview-1578.app.dev.fractalframework.xyz/proposals/1?dao=sep:0x7E9308800b7F3234982867b95481C8242C8AbFD1

![Screenshot 2024-05-16 at 16 58 20](https://github.com/decentdao/fractal-interface/assets/31554133/03857e24-37ba-42aa-9909-c85c2e8a5d25)

version 2 mainnet token voting dao
--> https://deploy-preview-1578.app.dev.fractalframework.xyz/proposals/17?dao=eth:0x36bD3044ab68f600f6d3e081056F34f2a58432c4
[...]

PR #1808, opened 2024-05-17
`Reskin` | Quorum progress bug

DarksightKellar:
Addresses #1793 

Turns out it wasn't really a reskin, but a till-now undiscovered bug.

It's a pretty wild edge case that appears because of two conditions:

- `bigint`s cannot have less-than-one values
- the required voting quorum is a percentage that leads to a token count less than 1. In the report's case, it was `4% of 14 == 0.56`

The `QuorumProgressBar` worked with `bigints`, and the `totalQuorum` calc resulted in 0 in this edge case, causing it to become false-y and leading to the bug. I have updated the calculation to use numbers instead.

...

As I write this I suddenly realise updating `QuorumProgressBar`'s `reachedQuorum` and `totalQuorum` to `numbers` might introduce another bug for potentially large values of `totalQuorum` that overflow the max `number` value. 

Is that likely to happen, seeing as `totalQuorum` will always be a percentage of `votesToken.totalSupply`? Thoughts?

### DIFF
--- a/src/components/Proposals/ProposalSummary.tsx
+++ b/src/components/Proposals/ProposalSummary.tsx
@@ -94,15 +94,18 @@ export default function ProposalSummary({ proposal }: { proposal: AzoriusProposa
       : isERC721
         ? votingStrategy.quorumThreshold!.value
         : 1n;
+
   const reachedQuorum = isERC721
     ? totalVotesCasted - no
     : votesToken
       ? (totalVotesCasted - no) / votesTokenDecimalsDenominator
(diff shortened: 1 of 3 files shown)


tomstuart123 (approved):
From a UX perspective, this solution works for both the DAOs I raised and a few others. Thanks @DarksightKellar 

I will let @decentdao/engineering comment on the risk you highlighted . If you want me to try and reproduce I can, but might need more info on how it would appear

Issue #1810, opened 2024-05-17
Re-arrange vote choices - proposal details


tomstuart123:
Please re-arrange vote choices to match the order in breakdown:
Yes
No
Abstain

![screenshot](https://netlify-cocoon.netlify.app/.netlify/functions/fetch?code=307&path=eyJzaXRlX2lkIjoiY2NhZDFmZmUtZDU4ZS00MjM4LWJiY2EtY2MwOTMxNTg1NWNmIiwiZGVwbG95X2lkIjoiNjY0NzgwODkyYzg2YzIwMDA4OGI1Nzg5IiwiaWQiOiJiOGFjOGRiNi02NmYzLTRmOWUtODAzZi00N2JiYjQ5MzJkZjYifQ==)

<details><summary>Browser metadata</summary>

```
Path:      /proposals/7?dao=sep:0x6DEBaE44A8f592733D1D4854b27108F145A1fdb2
Browser:   Chrome 125.0.0.0 on Mac OS 10.15.7
Viewport:  1705 x 1027 @2x
Language:  en-US
Cookies:   Enabled
```

[Open in BrowserStack](https://www.browserstack.com/user/try-live?url=https%3A%2F%2Fdeploy-preview-1578.app.dev.fractalframework.xyz%2Fproposals%2F7%3Fdao%3Dsep%3A0x6DEBaE44A8f592733D1D4854b27108F145A1fdb2&os=OS%20X&os_version=Catalina&browser=Chrome&browser_version=125.0&resolution=1705x1027&speed=1&start=true&ref=netlify-source)

</details>

[Open Deploy Preview](https://deploy-preview-1578.app.dev.fractalframework.xyz/proposals/7?dao=sep:0x6DEBaE44A8f592733D1D4854b27108F145A1fdb2) · [Mark as Resolved](https://app.netlify.com/cdp/resolve?deployID=664780892c86c200088b5789&commentID=664783ae7c57caf4a4ac9587&resolution=resolved)

_Originally posted by @xraystyle1980 in https://github.com/decentdao/fractal-interface/issues/1578#issuecomment-2117945233_

PR #1824, opened 2024-05-20
`Reskin` | 1810 rearrange vote choice

DarksightKellar:
Addresses #1810

### DIFF
--- a/src/components/Proposals/ProposalActions/CastVote.tsx
+++ b/src/components/Proposals/ProposalActions/CastVote.tsx
@@ -27,7 +27,7 @@ function Vote({
   onCastSnapshotVote?: () => Promise<void>;
 }) {
   const [pending, setPending] = useState<boolean>(false);
-  const [voteChoice, setVoiceChoice] = useState<AzoriusVoteChoice>();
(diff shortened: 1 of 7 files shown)


adamgall:
@DarksightKellar have you confirmed that voting for yes, no, abstain via the UI actually results in those proper selections onchain?

I just want to be extra sure that rearranging the radio icons _also_ results in the selection logic underneath them being updated too.

adamgall (commented):
@DarksightKellar have you confirmed that voting for yes, no, abstain via the UI actually results in those proper selections onchain?

I just want to be extra sure that rearranging the radio icons _also_ results in the selection logic underneath them being updated too.

DarksightKellar:
> @DarksightKellar have you confirmed that voting for yes, no, abstain via the UI actually results in those proper selections onchain?
> 
> I just want to be extra sure that rearranging the radio icons _also_ results in the vote selection logic underneath them being updated too.

Yep, the logic was broken. I've pushed a commit that aims both to fix this, and to reduce the dependency on the "magic" number values of each vote choice.

adamgall (approved):

Da-Colon (approved):

PR #1835, opened 2024-05-20
`BugFix` 1827 Safe Created DAO Page doesn't work

Da-Colon:
TL:DR found that a conditional we have had around for a while was preventing `Safe Created` DAOs from loading properly

## Testing
- Create a Safe via Safe's UI
- Everything should load correctly now in Decent's UI


adamgall:
@Da-Colon was there an issue associated with this? Can you explain how to recreate the issue?

tomstuart123 (approved):
the 'stuck on loading' stuff I was getting is fixed for safe UI created safes. 

I'm still not getting the 'add snapshot space' to work but discussed with david creating a new issue for this

adamgall (approved):

PR #1840, opened 2024-05-20
Smol update

adamgall:
- Remove instances of VITE_APP_TESTING_ENVIRONMENT because we don't do that anymore
- Use a bigger icon in wagmi metadata for wallet connect and coinbase wallet and things like that


Da-Colon (approved):
LGTM. You do have two lint warnings in Github

adamgall:
> LGTM. You do have two lint warnings in Github

Ah, good call! And that actually led to a bit more smol cleanup!

PR #1849, opened 2024-05-21
Remove languages

### LINKED ISSUE #1847: Temporarily remove multi-language support
Due to the staleness of our only other non-english language (ukranian), and because we're re-doing the copy in the app anyway, let's remove UK (and therefore, all) language support from the app for our rebrand launch.

This has the added benefit of not shipping a broken mobile nav language selection element.

adamgall:
Closes #1847 

- Removes the Language Switcher component from nav menu.
- Removes all Ukrainian language assets from the codebase. We'll add these back after revamping the English copy.


nicolaus-sherrill (approved):
🖖🏻

tomstuart123 (approved):
approved

PR #1896, opened 2024-05-23
convert to number before performing operations to get percentage

### LINKED ISSUE #1872: Expected behavior for total % of vote to not equal 100%?
Expected behavior for total % of vote to not equal 100%? 
https://app.dev.decentdao.org/proposals/17?dao=eth:0x36bD3044ab68f600f6d3e081056F34f2a58432c4

Screenshot 2024-05-22 at 9 13 05 AM

_Originally posted by @SpookyAction-BTC in https://github.com/decentdao/decent-interface/discussions/1863#discussioncomment-9524241_

Da-Colon:
Fixes #1872

Found that `bigint` doesn't support decimals directly. If there is a better way to convert and get these values let me know

localhost_3000_proposals_17_dao=eth_0x36bD3044ab68f600f6d3e081056F34f2a58432c4(Desktop)

### DIFF
--- a/src/components/Proposals/ProposalVotes/index.tsx
+++ b/src/components/Proposals/ProposalVotes/index.tsx
@@ -15,14 +15,14 @@ import ProgressBar from '../../ui/utils/ProgressBar';
 import ProposalERC20VoteItem from './ProposalERC20VoteItem';
 import ProposalERC721VoteItem from './ProposalERC721VoteItem';
 
-export function VotesPercentage({ label, percentage }: { label: string; percentage: number }) {
+export function VotesPercentage({ label, percentage }: { label: string; percentage: string }) {
   return (
     <Flex
       flexWrap="wrap"
       marginTop={2}
     >
       <ProgressBar
-        value={percentage}
+        value={Number(percentage)}
         label={label}
       />
     </Flex>
@@ -54,11 +54,11 @@ function ProposalVotes({
   );
 
   const getVotesPercentage = useCallback(
-    (voteTotal: bigint): number => {
+    (voteTotal: bigint): string => {
       if (totalVotesCasted === 0n) {
-        return 0;
+        return '0';
       }
-      return Number((voteTotal * 100n) / totalVotesCasted);
+      return ((Number(voteTotal) * 100) / Number(totalVotesCasted)).toFixed(2);
     },
     [totalVotesCasted],
   );
--- a/src/components/Proposals/ProposalVotes/index.tsx
+++ b/src/components/Proposals/ProposalVotes/index.tsx
@@ -15,14 +15,14 @@ import ProgressBar from '../../ui/utils/ProgressBar';
 import ProposalERC20VoteItem from './ProposalERC20VoteItem';
 import ProposalERC721VoteItem from './ProposalERC721VoteItem';
 
-export function VotesPercentage({ label, percentage }: { label: string; percentage: string }) {
+export function VotesPercentage({ label, percentage }: { label: string; percentage: number }) {
   return (
     <Flex
       flexWrap="wrap"
       marginTop={2}
     >
       <ProgressBar
-        value={Number(percentage)}
+        value={percentage}
         label={label}
       />
     </Flex>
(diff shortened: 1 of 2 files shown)


adamgall (commented):
Yeah, I don't really think it's ok to do this. You can't just cast a `bigint` as a `Number` and expect it to be ok, like you're doing with `voteTotal` and `totalVotesCasted`. Especially when we don't control what value is in there. There's a non-zero possibility that the number of tokens being used for votes exceeds what fits in a `number`.

Give something like this a shot: https://stackoverflow.com/a/63095380/991464

current_hp * 100n / max_hp

```ts
  const getVotesPercentage = useCallback(
    (voteTotal: bigint): string => {
      if (totalVotesCasted === 0n) {
        return 0;
      }
      return Number(voteTotal * 100n / totalVotesCasted);
    },
    [totalVotesCasted],
  );
```

Da-Colon:
> Yeah, I don't really think it's ok to do this. You can't just cast a `bigint` as a `Number` and expect it to be ok, like you're doing with `voteTotal` and `totalVotesCasted`. Especially when we don't control what value is in there. There's a non-zero possibility that the number of tokens being used for votes exceeds what fits in a `number`.
> 
> Give something like this a shot: https://stackoverflow.com/a/63095380/991464
> 
> ```ts
>   const getVotesPercentage = useCallback(
>     (voteTotal: bigint): number => {
>       if (totalVotesCasted === 0n) {
>         return 0;
>       }
>       return Number(voteTotal * 100n / totalVotesCasted);
>     },
>     [totalVotesCasted],
>   );
> ```

image


This is a lot cleaner, nice. Not sure why the its not grabbing that 19.9`1` 🤷

adamgall:
There is still a `toFixed(1)` in there, perhaps that's doing it.

Da-Colon:
> There is still a `toFixed(1)` in there, perhaps that's doing it.

Unfortunately unrelated. as something to do with Snapshot Proposals

DarksightKellar:
@Da-Colon this works:

```
  const roundedPercentage = (Number((voteTotal * 100000n) / totalVotesCasted) / 1000).toFixed(2);
  return Number(roundedPercentage);
```
Screenshot 2024-05-23 at 12 51 40

Pushing a quick commit

Da-Colon:
O sorry I have acommit locally I'll push up in a second. That gets it right

Da-Colon:
So I wanted to sleep on it :D but you can scale it up, I just didn't like the `.toFixed(2)` 

image

Da-Colon:
image

DarksightKellar (approved):
@Da-Colon Perfection

src/components/Proposals/ProposalVotes/index.tsx
```diff
@@ -71,9 +71,9 @@ function ProposalVotes({
     );
   }
 
-  const yesVotesPercentage = getVotesPercentage(yes);
-  const noVotesPercentage = getVotesPercentage(no);
-  const abstainVotesPercentage = getVotesPercentage(abstain);
+  const yesVotesPercentage = getVotesPercentage(yes).toFixed(2);
```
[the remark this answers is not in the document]
adamgall: i dig it. pushing a commit now.

src/components/Proposals/SnapshotProposalDetails/SnapshotProposalVotes.tsx
```diff
@@ -92,10 +92,7 @@ export default function SnapshotProposalVotes({ proposal }: ISnapshotProposalVot
                 >
                   {t('shutterVotesHidden')} |
                 </Text>
-                <ProposalCountdown
-                  proposal={proposal}
-                  showIcon={false}
-                />
+                <ProposalCountdown proposal={proposal} />
```
adamgall, line 95: @Da-Colon this change here is intended?

adamgall (approved):
@Da-Colon approving, but left a question

src/components/Proposals/SnapshotProposalDetails/SnapshotProposalVotes.tsx
Da-Colon: I'll revert this change

PR #1898, opened 2024-05-23
Snapshot Space Validation on Create Form

### LINKED ISSUE #1873: Can enter non-ENS in Snapshot ENS input
image

This is a bug. You should not be able to  write "no" and move on to next page

_Originally posted by @Delmslaugh in https://github.com/decentdao/decent-interface/discussions/1863#discussioncomment-9524351_

Da-Colon:
Fixes #1873 

use Yup for validating ENS field
- Looks like we had the functionality almost there. Added the validation test and added it to the create schema and remove the in component validation.

### DIFF
--- a/src/hooks/schemas/common/useValidationAddress.tsx
+++ b/src/hooks/schemas/common/useValidationAddress.tsx
@@ -1,3 +1,4 @@
+import { ens_normalize } from '@adraffy/ens-normalize';
 import { Signer } from 'ethers';
 import { useMemo, useRef, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
@@ -11,22 +12,15 @@ import { couldBeENS } from '../../../utils/url';
 import useSignerOrProvider from '../../utils/useSignerOrProvider';
 
 export function validateENSName({ ensName }: { ensName: string }) {
-  if (couldBeENS(ensName)) {
-    return {
-      validation: {
-        address: '',
-        isValidAddress: true,
-      },
-      isValid: false,
-    };
+  try {
+    if (couldBeENS(ensName)) {
+      ens_normalize(ensName);
+      return true;
+    }
+    return false;
+  } catch (error) {
+    return false;
   }
-  return {
-    validation: {
-      address: '',
-      isValidAddress: false,
-    },
-    isValid: false,
-  };
 }
 
 export async function validateAddress({
@@ -107,6 +101,18 @@ export const useValidationAddress = () => {
       },
     };
   }, [signerOrProvider, addressValidationMap, t]);
+
+  const ensNameValidationTest = useMemo(() => {
+    return {
+      name: 'ENS Validation',
+      message: t('errorInvalidENSName', { ns: 'common' }),
+      test: async function (ensName: string | undefined) {
+        if (ensName === undefined) return true;
+        const isValid = validateENSName({ ensName });
+        return isValid;
+      },
+    };
+  }, [t]);
 
   const addressValidationTestSimple = useMemo(() => {
     return {
@@ -208,6 +214,7 @@ export const useValidationAddress = () => {
   return {
     addressValidationTestSimple,
     addressValidationTest,
+    ensNameValidationTest,
     newSignerValidationTest,
     uniqueAddressValidationTest,
     uniqueNFTAddressValidationTest,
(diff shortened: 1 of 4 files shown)


src/hooks/schemas/common/useValidationAddress.tsx
```diff
@@ -11,22 +12,15 @@ import { couldBeENS } from '../../../utils/url';
...
-    return {
-      validation: {
-        address: '',
-        isValidAddress: true,
-      },
-      isValid: false,
-    };
+  try {
+    if (couldBeENS(ensName)) {
+      ens_normalize(ensName);
```
Da-Colon, line 17: I looked at the npm package and isn't clear what this adds as far as validating the string. only seem to go throw when I added a `.` (in the previous setup) but resolved so `no.no` would still pass

adamgall: I think we just utilize this library and not overthink it. If it passes this library's checks that means it's a valid ENS name. Simple as that.

Da-Colon: https://docs.ens.domains/registry/eth

Looks like there is a min character limit on ENS names. 3 min (and oof those prices if you want a short word)

But yeah tested.
`nope.no` ~ passes
`no.no` ~ fails
`nope` ~ fails
`works.eth` ~ passes

src/hooks/schemas/common/useValidationAddress.tsx
```diff
@@ -1,3 +1,4 @@
+import { ens_normalize } from '@adraffy/ens-normalize';
```
adamgall, line 1: Can you just use `import { normalize } from 'viem/ens';`

src/hooks/schemas/common/useValidationAddress.tsx
```diff
@@ -11,22 +12,15 @@ import { couldBeENS } from '../../../utils/url';
...
-  if (couldBeENS(ensName)) {
-    return {
-      validation: {
-        address: '',
-        isValidAddress: true,
-      },
-      isValid: false,
-    };
+  try {
+    if (couldBeENS(ensName)) {
```
adamgall, line 16: I want to get rid of this `couldBeENS`, it's dumb. see: https://github.com/decentdao/decent-interface/issues/1571

Speaking of which, would it make sense to tackle that issue in this PR?

adamgall (changes requested):

src/hooks/schemas/common/useValidationAddress.tsx
adamgall: so ACKSHUALLY, the 3 char limit only exists on the registry contract which controls the `.eth` TLD name registrations.

`no.no` _should be_ a valid ENS name, below the "registry" layer, assuming there was a `.no` TLD and it's registry contract didn't include any char limit restrictions.

there's a lot of context behind what I'm saying, feel free to look into how ENS works, it's fascinating.

Anyway, i bet that `no.no` is failing because of that homerolled `couldBeENS` function in our codebase, which I'd like to get rid of.

src/hooks/schemas/common/useValidationAddress.tsx
Da-Colon: ah okay. I can do that for sure.

src/utils/url.ts
```diff
@@ -11,23 +13,18 @@ export const isValidUrl = (urlString: string) => {
...
-  const validDomain = [...tld].length >= 2 && /^[-A-Za-z0-9]*$/.test(tld);
-
-  return validName && validDomain;
+  try {
+    normalize(ensAddress);
+    return true;
+  } catch (e) {
+    return false;
+  }
 };
```
Da-Colon, line 30: I repurposed and renamed CouldBeENS to `validateENSName`. 

It now uses the normalize function from viem as well as checks the `name` to make sure it is more than 3 characters in length. 

as far as the TLD. @adamgall do we need to do any verification here on supported TLD (such as .eth, etc) or is this good for now?

For example `epicbad.timing` or even `epicbadtimin.g` would pass currently

adamgall: I think this is fine as is, and maybe even still does too much.

A part of me says that we shouldn't even be doing the 3 char check, but I guess it's fine.

The question is, does Snapshot support any other TLDs other than `.eth`? If so, the 3 char check might not be valid for those other TLDs.

adamgall: Gonna give an approval to this as it currently stands, though.

adamgall (approved):

PR #1904, opened 2024-05-23
Feedback button mobile

### LINKED ISSUE #1874: report a bug button covers mobile toast banner
report a bug button covers mobile toast banner

_Originally posted by @azadikoff in https://github.com/decentdao/decent-interface/discussions/1863#discussioncomment-9524252_

adamgall:
Closes #1874 

In lieu of new designs, this PR does two things to address the various issues with the Report a Bug button covering stuff.

1. Increases the `z-index` of the toast element so that _it_ covers the "Report a Bug" button, which is the actual bug report referenced in #1874.
Screenshot 2024-05-23 at 5 21 03 PM

2. On mobile, adds extra padding to the footer, so that the "Report a Bug" button simply doesn't cover anything. I don't think we have an explicit issue about this, but it's been called out in the past.
Screenshot 2024-05-23 at 5 18 54 PM

### DIFF
--- a/src/components/ui/page/Footer.tsx
+++ b/src/components/ui/page/Footer.tsx
@@ -11,8 +11,6 @@ export function Footer() {
       flexWrap="wrap"
       justifyContent="center"
       alignItems="center"
-      p="3rem"
-      pt="1.5rem"
       gap={{ base: 4, md: 8 }}
     >
       <Flex gap={4}>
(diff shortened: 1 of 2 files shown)


tomstuart123:
Nice this works for me on cb wallet mobile well. 

Thanks for being creative @adamgall / @Da-Colon

tomstuart123 (approved):

PR #1909, opened 2024-05-24
Safe Loading/Unloading Problems

### LINKED ISSUE #1515: [Bug] DAO state resetting issue
Reproduce. (easy way to see it)

1. Go to any favorited DAO.
2. Open Favorites
3. Click on DAO currently viewing

This causes the DAO state to reset (due to a action.resetDAO on the button click).

But the dao is never refreshed.

### LINKED ISSUE #1582: Bug: Back Button DAO not unloading/reloading
Reproduce. 

- Go to a DAO. 
- Use favorite menu to go another DAO
- Hit Back button

- The address and DAO info updates. but proposals state doesn't reset.

### LINKED ISSUE #1911: Safes mix proposals when navigated via 'My Safes' AND back button.
can you check the video I just [sent here](https://decentdao.slack.com/archives/C06QZTUTKRC/p1716548806433229?thread_ts=1716548771.976629&cid=C06QZTUTKRC) - 

I went through the flow in issue [1910 ](https://github.com/decentdao/decent-interface/issues/1910)and for some reason the proposals from the 2nd DAO I navigated to appear in the proposal list of the 1st DAO I went back to (via back button). I just [sent a recording here](https://decentdao.slack.com/archives/C06QZTUTKRC/p1716548806433229?thread_ts=1716548771.976629&cid=C06QZTUTKRC) - 

Something must be happening as we navigate between Safes when the back button is involved that means state is not updated

_Originally posted by @tomstuart123 in https://github.com/decentdao/decent-interface/issues/1910#issuecomment-2129264742_

Da-Colon:
Fixes #1515 
Fixes #1582 
Fixes #1911

### DIFF
--- a/src/components/pages/DaoDashboard/Info/ParentLink.tsx
+++ b/src/components/pages/DaoDashboard/Info/ParentLink.tsx
@@ -12,7 +12,6 @@ import { useNetworkConfig } from '../../../../providers/NetworkConfig/NetworkCon
 export function ParentLink() {
   const {
     node: { nodeHierarchy },
-    action,
   } = useFractal();
   const { addressPrefix } = useNetworkConfig();
   const { t } = useTranslation('breadcrumbs');
@@ -26,7 +25,6 @@ export function ParentLink() {
       color="celery-0"
       _hover={{ textDecoration: 'none', color: 'celery--6' }}
       to={DAO_ROUTES.dao.relative(addressPrefix, nodeHierarchy.parentAddress)}
-      onClick={action.resetSafeState}
       marginBottom="1rem"
       as={RouterLink}
       width="fit-content"
--- a/src/components/ui/cards/DAONodeInfoCard.tsx
+++ b/src/components/ui/cards/DAONodeInfoCard.tsx
@@ -32,7 +32,6 @@ export interface InfoProps extends FlexProps {
 export function DAONodeInfoCard({ node, freezeGuard, guardContracts, ...rest }: InfoProps) {
   const {
     node: { daoAddress: currentDAOAddress }, // used ONLY to determine if we're on the current DAO
-    action,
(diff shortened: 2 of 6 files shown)


tomstuart123:
So 1 take and 1 quick issue with current implementation.

issue I face. If no other favorited safes, the 'You have not added any Safes yet.' doesn't show. [see video](https://decentdao.slack.com/archives/C06QZTUTKRC/p1716549145915569?thread_ts=1716548771.976629&cid=C06QZTUTKRC)

adamgall:
I agree with @tomstuart123 99%

The 1% I disagree with is that the "show but disabled" should be implemented in this PR!

adamgall (changes requested):
As mentioned already, we don't want to just remove things from the My Safes list, that'll be confusing to the user.

1. At a bare minimum, the current Safe should be shown but disabled.
2. Ideally, we figure out what is actually breaking when trying to load a Safe which is already loaded, and fix that.

Da-Colon:
@tomstuart123 

After some discussion and discovery, found that some old patterns can be removed, specifically resetting the global state BEFORE navigating. We were able to update the hook that loads in the data to also reset properly.

So when you navigate to another Safe (via search or favorites) and its the same Safe you are viewing we aren't going to disable the button. it basically just navigates to dashboard.

If you are on Safe Dashboard. and click the same viewed safe it does nothing (closes the menu)
if you are say the proposals page adn click the same viewed safe it will navigating back home.

tomstuart123:
Not sure if your done with the linter

But I tried the flow for all three of the issues listed above. This seems to solve it thanks @Da-Colon

tomstuart123 (approved):

adamgall (approved):
L E T S G O

PR #1913, opened 2024-05-24
Finish renaming hierarchy

adamgall:
On the renamed "Organization" page, we were still using the word "Hierarchy" in two places:
1. Main `H1` page title
2. Breadcrumb

This fixes that to use the word "Organization" as it should.

No issue, I just noticed it and decided to fix it right quick.


tomstuart123 (approved):
got context at standup

PR #1918, opened 2024-05-24
Show loading message when My Safe names are loading

### LINKED ISSUE #1531: Refactor `useDaoName`
The file `useDaoName.ts` currently has 2 hooks: `useDAOName` and `useLazyDAOName`, which both do very similar things. We should consider merging these two into one, and making sure any unused code paths are removed.

Context: [this PR comment](https://github.com/decentdao/fractal-interface/pull/1530#pullrequestreview-1990610612)

There's also a bug related to this where Safe Names that are loading in, don't show the "Loading Safe" message, just an empty string.

adamgall:
Closes #1531 

Product note:
- Now a "Loading Safe Name" message appears while My Safe names are loading in.

Dev notes:
- Consolidate logic in useDAOName, but need two hooks still. This one took me a while to work through, I spent too much time trying to make just one hook, but _I couldn't figure it out_


Da-Colon (approved):
Make sense to me. Tried it out and works well.

PR #1920, opened 2024-05-24
Home designs desktop top nav changes

adamgall:
Closes https://github.com/decentdao/decent-interface/issues/1905

Implements everything in the attached issue except for `max-width: 400px` on the search input. That causes all kinds of issues, including:
- ethereum address doesn't fit
- struggling to _not_ apply that max-width on the search input in full-screen modal, _and_ make the search result popup have the correct width

### DIFF
--- a/src/i18n/locales/en/dashboard.json
+++ b/src/i18n/locales/en/dashboard.json
@@ -1,6 +1,7 @@
 {
   "titleFavorites": "My Safes",
-  "emptyFavorites": "You have not added any Safes yet.",
+  "emptyFavorites": "You haven't added any Safes yet.",
+  "emptyFavorites2": "Create or find a Safe above.",
   "loadingFavorite": "Loading Safe Name",
   "errorInvalidSearch": "Oops! This Ethereum address is invalid.",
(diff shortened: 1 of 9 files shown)


src/i18n/locales/en/dashboard.json
```diff
@@ -1,6 +1,7 @@
 {
   "titleFavorites": "My Safes",
-  "emptyFavorites": "You have not added any Safes yet.",
+  "emptyFavorites": "You haven't added any Safes yet.",
+  "emptyFavorites2": "Create or find a Safe above.",
```
DarksightKellar, line 4: `emptyFavoritesSubtext` or `emptyFavoritesSubtitle`?

DarksightKellar (approved):
Name nitpick, looks good to me otherwise

src/i18n/locales/en/dashboard.json
adamgall: Haha good call

adamgall: Fixed and pushed

PR #1921, opened 2024-05-24
`1902` Update Signers UX for Multisig

### LINKED ISSUE #1902: Settings - Add & Remove signers pattern
In response to https://github.com/decentdao/decent-interface/issues/1883
@adamgall 
Settings for Multisig Safe's has been updated with a new pattern for adding and removing signers. The radio's (and checkboxes) have been removed entirely from the design and the "Remove Signer" button has been removed.
Button-icon's with minus have been added to the right of the signer's address. Clicking/tapping this button will open the confirmation modal.

[In Figma](https://www.figma.com/design/9hUYtvCKjOyB92vt8ukYtD/Product-Handoff?m=auto&node-id=282%3A37051&t=cmhPW3aWoEd2sPrM-1)

Image

Image

Da-Colon:
Fixes #1902 

image

## Dev notes
- I move the modal opening `useFractalModal` into the `Signer` Component passing the needed props through. Because of how the modal consumes the address, I was able to use the current state setup in this way, so removed that. 
- I use `Button` component with a `Icon` inside as the size property. `icon-sm` and `icon-lg` for IconButton didn't fit with figma design of `20x20` for Icon.

### DIFF
--- a/src/components/pages/DaoSettings/components/Signers/SignersContainer.tsx
+++ b/src/components/pages/DaoSettings/components/Signers/SignersContainer.tsx
@@ -1,4 +1,4 @@
-import { Button, Flex, HStack, Radio, RadioGroup, Show, Text, Hide } from '@chakra-ui/react';
+import { Button, Flex, HStack, Show, Text, Hide, Icon } from '@chakra-ui/react';
 import { PlusCircle, MinusCircle } from '@phosphor-icons/react';
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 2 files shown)


tomstuart123 (approved):
Nice, to the design and approved

Note - @decentdao/design  - I missed the figma so sorry for the late discussion topic. Can I understand the red warning a bit? My usual impression in our app is warnings are yellow in the app and red is errors ?

Devs - this is just an open question for my knowledge, don't let it stop pushing to production

adamgall (approved):
Approved, nice job. Also agree with Tom's question when removing a singer. Should that message be in red? or yellow. I'd suggest changing it to yellow.

adamgall:
@xraystyle1980 if red is intentional, I don't want to cause friction here. After @tomstuart123 raised it, my gut also went to "red == error" (which isn't the case here).

xraystyle1980:
> @xraystyle1980 if red is intentional, I don't want to cause friction here. My thinking was that after @tomstuart123 raised it, my gut also went to "red == error" (which isn't the case here).

Good to hear. As it stand, we don't currently have a pattern for making the input yellow. Can we adjust the input to not appear red? See example:
image

xraystyle1980 (changes requested):
Update the input with the address being removed to not be red
image

DarksightKellar (approved):

Da-Colon:
@xraystyle1980 replaced setup with the disabled input 🫡 

image

xraystyle1980 (approved):
Looks good. :shipit:

PR #1926, opened 2024-05-25
`1924` Favorites Drawer reload

### LINKED ISSUE #1924: Favorites toggling doesn't work in mobile drawer view
When on mobile (and therefore the "My Safes" view uses the drawer), toggling a Favorite safe via the star icon doesn't toggle that Safe within the drawer.

Only after a refresh is the Safe properly reflected as added or removed from the My Safes drawer.

Da-Colon:
Fixes #1924 

## Dev notes
So there is some difference on how the `Menu` component is mounted/unmounted and the `Drawer` component. I'm not entirely sure why when the `Drawer` component is closed, it appears that component isn't getting unmounted and/or is getting cached, the `useAccountFavorites` hook never refreshes between opening and closing. As it works fine in the Menu on desktop.

After many attempts with `useEffect` a `reloadFavorites` function `onOpen`, ultimately the cleanest solution ended up being just replacing the use of `favoritesList` out of `useAccountFavorites` and get the list directly in the component.

### DIFF
--- a/src/pages/home/AllSafesDrawer.tsx
+++ b/src/pages/home/AllSafesDrawer.tsx
@@ -10,7 +10,8 @@ import {
 import { useState, useRef, useEffect, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import { BACKGROUND_SEMI_TRANSPARENT } from '../../constants/common';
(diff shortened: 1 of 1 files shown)


adamgall (approved):
Simple changes, and works well for me!

DarksightKellar (approved):

PR #1929, opened 2024-05-27
1675 proposals card updates

adamgall:
Closes https://github.com/decentdao/decent-interface/issues/1675

Renames the two properties in the Governance card on DAO Dashboard to be

- Total
- Active

This one was a lot of work to deal with all of the potential permutations of azorius/multisig and snapshot/nonsnapshot.

But I think I got it.

Displays both values as quickly as able to.

### Onchain governance

- Multisig proposals are loaded in all at once, so we know this total number pretty quickly.
- Azorius proposals are loaded in one at a time, but from newest to oldest, and the proposal ID is incremented by one each time, so we know the total number of proposals as soon as the first one comes in by looking at it's ID.

### Offchain governance

- Snapshot proposals are loaded in all at once, so we know this total number pretty quickly.

### Cumulative

- Finally, we add the total number of Snapshot proposals (if linked) to the total number of onchain governance proposals, to get the total number.

### Onchain governance
- Multisig proposals are loaded in all at once, so we know the number of active proposals pretty quickly.
- Azorius proposals are loaded in one at a time, but from newest to oldest, and the proposal ID is incremented by one each time. Naively, we would say that we don't know the number of active proposals until each one is loaded in so we can look at its state. However, we can be clever here and notice when the _first non-active_ proposal is loaded. At that point, the chances are super low (but not zero*) that all remaining proposals are completed. So we can simply use this as our trigger point to display a value instead of "...".

### Offchain governance

- Snapshot proposals are loaded in all at once, so we know the number of active proposals pretty quickly.

### Cumulative

- Finally, we add the total number of active Snapshot proposals with the total number of active onchain governance proposals, to get the total number of active proposals.

## Footnote

\* "Active proposals" in this component is defined as "votable + executable". There's a chance that an "executable" proposal may be further back in time than an "executed" proposal, which will result in the "Active" number displaying a value before we truly know the actual number. However, it will update to its correct number as the new "executable" proposal is loaded in.

### DIFF
--- a/src/components/pages/DaoDashboard/Activities/hooks/useActivities.ts
+++ b/src/components/pages/DaoDashboard/Activities/hooks/useActivities.ts
@@ -1,6 +1,6 @@
 import { useMemo } from 'react';
 import { useFractal } from '../../../../../providers/App/AppProvider';
-import { Activity, ActivityEventType, SortBy } from '../../../../../types';
+import { Activity, SortBy } from '../../../../../types';
 
 export const useActivities = (sortBy: SortBy) => {
   const {
@@ -12,20 +12,14 @@ export const useActivities = (sortBy: SortBy) => {
    * updates when a different sort is selected
    */
   const sortedActivities: Activity[] = useMemo(() => {
-    return (
-      (proposals || [])
-        // Do not display Module transactions for now
-        // https://github.com/decentdao/fractal-interface/issues/1712
-        .filter(activity => activity.eventType !== ActivityEventType.Module)
-        .sort((a, b) => {
-          const dataA = new Date(a.eventDate).getTime();
-          const dataB = new Date(b.eventDate).getTime();
-          if (sortBy === SortBy.Oldest) {
-            return dataA - dataB;
-          }
-          return dataB - dataA;
-        })
-    );
+    return (proposals || []).sort((a, b) => {
+      const dataA = new Date(a.eventDate).getTime();
+      const dataB = new Date(b.eventDate).getTime();
+      if (sortBy === SortBy.Oldest) {
+        return dataA - dataB;
+      }
+      return dataB - dataA;
+    });
   }, [sortBy, proposals]);
 
   return { sortedActivities };
--- a/src/components/pages/DaoDashboard/Info/InfoProposals.tsx
+++ b/src/components/pages/DaoDashboard/Info/InfoProposals.tsx
@@ -33,21 +33,28 @@ const totalProposalsCount = (
 
   switch (type) {
     case GovernanceType.MULTISIG: {
+      // Multisig proposals (onchain) and Snapshot proposals (offchain)
+      // are all loaded "at once", so just return the length of this set.
       return proposals.length;
     }
     case GovernanceType.AZORIUS_ERC20:
     case GovernanceType.AZORIUS_ERC721: {
+      // First, we want to first filter out all of the Snapshot proposals...
       const nonSnapshot = nonSnapshotProposals(proposals);
+
+      // Then, find the highest ID from the Azorius proposals...
       const highestNonSnapshotProposalId = nonSnapshot.reduce((p, c) => {
-        // Proposal IDs are zero indexed!
+        // Remember, proposal IDs are one-indexed!
         const propId = Number(c.proposalId) + 1;
         if (propId > p) {
           return propId;
         }
         return p;
       }, 0);
-      const snapshotProposalCount = proposals.length - nonSnapshot.length;
-      return highestNonSnapshotProposalId + snapshotProposalCount;
+
+      // Then, return the highest Azorius proposal ID
+      // plus the number of Snapshot proposals.
+      return highestNonSnapshotProposalId + proposals.length - nonSnapshot.length;
     }
     default: {
       return 0;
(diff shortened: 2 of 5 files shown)


src/components/pages/DaoDashboard/Activities/hooks/useActivities.ts
```diff
@@ -12,20 +12,14 @@ export const useActivities = (sortBy: SortBy) => {
    * updates when a different sort is selected
    */
   const sortedActivities: Activity[] = useMemo(() => {
-    return (
-      (proposals || [])
-        // Do not display Module transactions for now
-        // https://github.com/decentdao/fractal-interface/issues/1712
-        .filter(activity => activity.eventType !== ActivityEventType.Module)
```
adamgall, line 19: No longer filtering out Module events, because we're not even storing them into state any longer.

src/components/pages/DaoDashboard/Info/InfoProposals.tsx
```diff
@@ -2,12 +2,120 @@ import { Box, Flex, Text } from '@chakra-ui/react';
...
+  GovernanceType,
+  SnapshotProposal,
+} from '../../../../types';
 import { BarLoader } from '../../../ui/loaders/BarLoader';
 
-interface IDAOGovernance {}
+const isSnapshotProposal = (proposal: FractalProposal) => {
+  const snapshotProposal = proposal as SnapshotProposal;
+  return !!snapshotProposal.snapshotProposalId;
+};
```
adamgall, line 16: Helper function to determine if a `FractalProposal` is actually type `SnapshotProposal`.

Little helper functions that do this same thing are scattered around the codebase. I obviously don't like it.

"Cleaning this up" (regarding our types and state structure, in general), is something else that's on my long todo list of things that need to happen in this app.

src/components/pages/DaoDashboard/Info/InfoProposals.tsx
```diff
@@ -2,12 +2,120 @@ import { Box, Flex, Text } from '@chakra-ui/react';
...
 
-export function InfoProposals({}: IDAOGovernance) {
+const snapshotProposals = (proposals: FractalProposal[]) => {
+  return proposals.filter(proposal => isSnapshotProposal(proposal));
+};
```
adamgall, line 20: Helper function to take list of `FractalProposal`s and return all of the `SnapshotProposals` that exist within it.

src/components/pages/DaoDashboard/Info/InfoProposals.tsx
```diff
@@ -2,12 +2,120 @@ import { Box, Flex, Text } from '@chakra-ui/react';
...
+
+const nonSnapshotProposals = (proposals: FractalProposal[]) => {
+  return proposals.filter(proposal => !isSnapshotProposal(proposal));
+};
```
adamgall, line 24: Helper function to take list of `FractalProposal`s and return a list of all the proposals which are _not_ `SnapshotProposal`s.

src/components/pages/DaoDashboard/Info/InfoProposals.tsx
```diff
@@ -2,12 +2,120 @@ import { Box, Flex, Text } from '@chakra-ui/react';
...
+    }
+  }
+};
+
+const isActiveProposal = (proposal: FractalProposal) => {
+  return (
+    proposal.state === FractalProposalState.ACTIVE ||
+    proposal.state === FractalProposalState.EXECUTABLE
+  );
+};
```
adamgall, line 63: Helper function to check if a proposal is in the state we're looking which we're calling "Active" in the UI.

src/components/pages/DaoDashboard/Info/InfoProposals.tsx
```diff
@@ -2,12 +2,120 @@ import { Box, Flex, Text } from '@chakra-ui/react';
...
+
+const activeProposals = (proposals: FractalProposal[]) => {
+  return proposals.filter(proposal => isActiveProposal(proposal));
+};
```
adamgall, line 67: Helper function to take list of `FractalProposal`s and return all of those which we deem "active".

src/components/pages/DaoDashboard/Info/InfoProposals.tsx
```diff
@@ -2,12 +2,120 @@ import { Box, Flex, Text } from '@chakra-ui/react';
...
+
+const nonActiveProposals = (proposals: FractalProposal[]) => {
+  return proposals.filter(proposal => !isActiveProposal(proposal));
+};
```
adamgall, line 71: Helper function to take list of `FractalProposal`s and return list of all of those we _don't_ deem "active".

src/hooks/utils/useSafeTransactions.ts
```diff
@@ -195,13 +195,16 @@ export const useSafeTransactions = () => {
         return [];
       }
 
+      const transactionsWithoutModuleTransactions = transactions.results.filter(
+        transaction => !isModuleTx(transaction),
+      );
+
       const activities = await Promise.all(
-        transactions.results.map(async (transaction, _, transactionArr) => {
+        transactionsWithoutModuleTransactions.map(async (transaction, _, transactionArr) => {
```
adamgall, line 203: We're filtering out "module" transactions here, not going to save them into state.

Da-Colon (approved):
Love the Clean up, Works as expected on my desktop and some stress testing (switching around DAOs). We did see the `QuorumBadge` error pop again but hard to recreate as it is still rare

tomstuart123:
Really like the hack of looking for the nearest 'non-active proposal'. Nice job

Verified to work on a few DAOs with a mix of executable / voting statuses on proposals

Approved

tomstuart123 (approved):

PR #1933, opened 2024-05-28
`1932`  adds fallback compatibility for older and mobile browsers

### LINKED ISSUE #1932: "Copy address" behavior not working on mobile
Experienced on:
- Mobile Chromium
- iOS

Da-Colon:
Fixes #1932 

See: https://stackoverflow.com/questions/51805395/navigator-clipboard-is-undefined

### DIFF
--- a/src/hooks/utils/useCopyText.ts
+++ b/src/hooks/utils/useCopyText.ts
@@ -1,14 +1,36 @@
 import { useTranslation } from 'react-i18next';
 import { toast } from 'react-toastify';
 
+/**
+ * Custom hook to copy text to the clipboard.
+ * Uses the Clipboard API if available and falls back to a textarea method for older browsers.
+ */
 export function useCopyText() {
   const { t } = useTranslation('common');
-  const copyTextToClipboard = (textToCopy?: string) => {
+
+
+  const copyTextToClipboard = async (textToCopy?: string) => {
     if (!textToCopy) return;
-    toast(t('toastClipboardCopy'), {
-      onOpen: () => navigator.clipboard.writeText(textToCopy),
-      autoClose: 1000,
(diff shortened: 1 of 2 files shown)


src/hooks/utils/useCopyText.ts
```diff
@@ -1,14 +1,35 @@
...
+        document.body.appendChild(textArea);
+        textArea.focus();
+        textArea.select();
+        document.execCommand('copy');
+        document.body.removeChild(textArea);
+      }
+      toast(t('toastClipboardCopy'), { autoClose: 1000 });
+    } catch (error) {
+      console.error('Failed to copy text to clipboard: ', error);
+      toast(t('toastClipboardCopyError'), { autoClose: 1000 });
```
adamgall, line 31: Where da strings at

adamgall (changes requested):

adamgall:
The implementation works for me tho

src/hooks/utils/useCopyText.ts
Da-Colon: Yep forgot to add the copy. I'll circle back in a bit

Da-Colon: would this change be acceptable, log the error for sentry and use the general error copy? As this would be an unexpected error if it were to occur.
```
logError(error);
toast(t('errorGeneral'), { autoClose: 1000 });
```

adamgall: @Da-Colon I think it's appropriate to use a specific error string here, because we're capturing this specific error. Let's just add a new error string to the localization.

I'll do it and push a commit.

adamgall (approved):

PR #1937, opened 2024-05-28
`1930` Move all theme files from `decent-ui`

Da-Colon:
Works on #1930 

This PR only moves over all theme files and folders needed to setup theme. PRs that follow will slowly replace implementing the theme in smaller chunks.

Next PRs:
- Replace all icon imports from `@decent-org/fractal-ui`
- Replace all `LabelWrapper` imports. Will also move the `LabelWrapper` from theme/custom, as it is just like any other component now
- Replace theme import and remove package

See individual commits in this PR for a break down of what was added.


adamgall:
@Da-Colon do you think it would make sense to target a new branch for all of this "deprecate `decent-ui`" work, and then merge that into `develop` when finished?

Da-Colon:
> @Da-Colon do you think it would make sense to target a new branch for all of this "deprecate `decent-ui`" work, and then merge that into `develop` when finished?
> 
> Or do you consider these changes safe-and-atomic enough to be deploying directly into `develop`, with the knowledge that `develop` is _always_ ready for a production push.

The latter is the goal. But this can also be considered the `root` PR. 

All this one does is move all the files into this repo. The other PRs will `slowly` implement these changes and will all be 'prod' safe, if merged. For readability of changes. I figured having the first PR just be a 'movement' PR will all files just copied over and getting into the codebase rather than adding + implementing each thing in its own PR

package.json
```diff
@@ -4,7 +4,9 @@
   "private": true,
   "dependencies": {
     "@apollo/client": "^3.7.10",
+    "@chakra-ui/anatomy": "^2.2.2",
     "@chakra-ui/react": "^2.8.2",
+    "@chakra-ui/theme-tools": "^2.1.2",
```
adamgall, line 9: I assume these packages are in `decent-ui`, and are somehow included thru that? But now we'll need them in this codebase directly as we eventually remove the `decent-ui` package?

adamgall (approved):

src/assets/theme/global/index.ts
```diff
@@ -0,0 +1,26 @@
...
+      backgroundAttachment: "fixed",
+      backgroundRepeat: "no-repeat",
+      scrollBehavior: "smooth",
+      height: "100%",
+    },
+    ...toastStyles(props),
+    ...scrollStyles,
+    ...inputStyles,
+  }),
+}
```
DarksightKellar: I can't quite remember, but weren't there some similar global css setup that recently had to be added to the app that can now be moved here now that we've got this? @decentdao/engineering

DarksightKellar (approved):

src/assets/theme/global/index.ts
adamgall: Hmmmmm, i don't recall such a global style. Other than the `sentry.css` file I recently made. But I don't think that's what ur talking about @DarksightKellar

package.json
Da-Colon: yessir

src/assets/theme/global/index.ts
Da-Colon: yeah I see `Markdown.css` Im sure there will be a few things that we can cleanup as we now have access to the theme right in the codebase.

PR #1939, opened 2024-05-28
Breakdown progress bars cleanup

### LINKED ISSUE #1934: Vote progress bars need some space between label and percentage
Example: https://app.dev.decentdao.org/proposals/17?dao=eth:0x36bD3044ab68f600f6d3e081056F34f2a58432c4

Figma looks like this

Image

Implementation looks like this

Image

@decentdao/design this design leaves some open questions. In the "Yes" and "No" options, the percentage is within the lilac bar, in "Abstain" it's outside of the lilac bar (because there is none).

Can you please describe how this should behave in a way that scales to a bar going from 0% to 100% filled?

My suggestion is to just show the percentage at the end of the bar, regardless of where the lilac fill is.

### LINKED ISSUE #1935: Weird z-index issues on Vote Progress Bars
The labels in Progress Bars have a z-index which displays _over_ the header.

Image

This should be fixed to ideally not need to specify the z-index of this content at all. Can't we lay out the HTML in a way so that the labels are painted over the bars?

adamgall:
Closes #1934 
Closes #1935 

See the two above linked issues to get an idea of what was broken.

Now see some screenshots of how it looks due to the changes in this PR

Screenshot 2024-05-28 at 4 41 35 PM
Screenshot 2024-05-28 at 4 41 42 PM
Screenshot 2024-05-28 at 4 42 05 PM
Screenshot 2024-05-28 at 4 42 09 PM

### DIFF
--- a/src/components/Proposals/ProposalVotes/index.tsx
+++ b/src/components/Proposals/ProposalVotes/index.tsx
@@ -1,4 +1,4 @@
-import { Box, Flex, Grid, GridItem, Text } from '@chakra-ui/react';
+import { Box, Flex, Grid, GridItem, Text, Progress } from '@chakra-ui/react';
 import { useCallback, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useFractal } from '../../../providers/App/AppProvider';
@@ -18,8 +18,8 @@ import ProposalERC721VoteItem from './ProposalERC721VoteItem';
 export function VotesPercentage({ label, percentage }: { label: string; percentage: number }) {
   return (
     <Flex
-      flexWrap="wrap"
       marginTop={2}
+      width="100%"
     >
       <ProgressBar
         value={percentage}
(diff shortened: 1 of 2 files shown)


xraystyle1980 (approved):
Looks good. :shipit:

tomstuart123 (approved):
yep this works for me. Approved

As a future step, I think it could be cool to have the original implementation on wide screens and then at a certain responsiveness break point it reverts to this implementation. Reasoning, it is nice having the % in the purple to be clear its referring to the votes. Saying that, lets not prioritise it now as this works

adamgall:
> yep this works for me. Approved
> 
> As a future step, I think it could be cool to have the original implementation on wide screens and then at a certain responsiveness break point it reverts to this implementation. Reasoning, it is nice having the % in the purple to be clear its referring to the votes. Saying that, lets not prioritise it now as this works

@tomstuart123 even still, without more direction from design, that's not possible. On widescreen, let's say one choice has 0.5% of votes. The purple bar will still just be maybe 10 or 20 px wide, and we can't fit everything within it.

That's my bigger point here: This (original) design isn't _finished_ yet. There are a lot of cases to cover and without rules that encompass all of them, we'll spend way too much time going back and forth, and we don't have the luxury of time for that.

tomstuart123:
ye you're right. Understood there's a few edge cases. @xraystyle1980 / @nicolaus-sherrill no time pressure on this from my side as this works for me

Once we're further along our feature development, we can return to the proposal details and have a think about this

PR #1947, opened 2024-05-29
1942 ens chain errors

adamgall:
Closes: https://github.com/decentdao/decent-interface/issues/1942

See: https://github.com/wevm/viem/discussions/781

TL;DR: simply capture the specific known and expected error, then do nothing about it. If the error _isn't_ this `ChainDoesNotSupportContract` error, then we actually do want to throw it up.


Da-Colon (commented):
Any particular way you want us to test this?

adamgall:
@Da-Colon commect to Polygon and confirm you don't see big red errors in the console yelling about not having an ENS Registry address.

DarksightKellar (approved):
Update looks good to me, I didn't see big red blotches in my console either, so, moneh

PR #1948, opened 2024-05-29
Formik-ify more send asset modal form fields

### LINKED ISSUE #1945: Cleanup from fixing mobile keyboard on Send Assets modal
The original PR was intentionally merged before it was fully completed.

https://github.com/decentdao/decent-interface/pull/1928

There are still a handful of "code cleanup" things to complete.

https://github.com/decentdao/decent-interface/pull/1928#issuecomment-2135413331
https://github.com/decentdao/decent-interface/pull/1928#pullrequestreview-2085606988

DarksightKellar:
Closes #1945 

See also: https://github.com/decentdao/decent-interface/pull/1928#discussion_r1619245208

I have chosen not to formik-ify the custom nonce input as it's got its own validation and wouldn't make sense to force that back outside to be handled by Formik/Yup

### DIFF
--- a/src/components/ui/forms/BigIntInput.tsx
+++ b/src/components/ui/forms/BigIntInput.tsx
@@ -19,6 +19,7 @@ export interface BigIntInputProps
   min?: string;
   max?: string;
   maxValue?: bigint;
+  currentValue?: BigIntValuePair;
 }
 /**
  * This component will add a chakra Input component that accepts and sets a bigint
@@ -39,6 +40,7 @@ export function BigIntInput({
   min,
   max,
   maxValue,
+  currentValue,
   ...rest
 }: BigIntInputProps) {
   const { t } = useTranslation('common');
@@ -141,6 +143,13 @@ export function BigIntInput({
     },
     [decimalPlaces, min, onChange],
   );
+
+  // if the parent Formik value prop changes, need to update the value
+  useEffect(() => {
+    if (!inputValue || inputValue === currentValue?.value ) return;
+    if (currentValue?.bigintValue === 0n) 
+      setInputValue('');
+  }, [currentValue, inputValue]);
 
   // if the decimalPlaces change, need to update the value
   useEffect(() => {
--- a/src/components/ui/forms/BigIntInput.tsx
+++ b/src/components/ui/forms/BigIntInput.tsx
@@ -146,9 +146,8 @@ export function BigIntInput({
 
   // if the parent Formik value prop changes, need to update the value
   useEffect(() => {
-    if (!inputValue || inputValue === currentValue?.value ) return;
-    if (currentValue?.bigintValue === 0n) 
-      setInputValue('');
+    if (!inputValue || inputValue === currentValue?.value) return;
+    if (currentValue?.bigintValue === 0n) setInputValue('');
   }, [currentValue, inputValue]);
 
   // if the decimalPlaces change, need to update the value
--- a/src/components/ui/modals/SendAssetsModal.tsx
+++ b/src/components/ui/modals/SendAssetsModal.tsx
@@ -32,11 +32,6 @@ export function SendAssetsModal({ close }: { close: () => void }) {
   const { t } = useTranslation(['modals', 'common']);
 
   const fungibleAssetsWithBalance = assetsFungible.filter(asset => parseFloat(asset.balance) > 0);
-
-  const [selectedAsset, setSelectedAsset] = useState<SafeBalanceResponse>(
-    fungibleAssetsWithBalance[0],
-  );
-  const [inputAmount, setInputAmount] = useState<BigIntValuePair>();
   const [nonceInput, setNonceInput] = useState<number | undefined>(safe!.nonce);
 
   const { submitProposal } = useSubmitProposal();
@@ -89,7 +84,8 @@ export function SendAssetsModal({ close }: { close: () => void }) {
           console.log(values);
 
           const overDraft =
-            Number(values.inputAmount?.value || '0') > formatCoinUnitsFromAsset(values.selectedAsset);
+            Number(values.inputAmount?.value || '0') >
+            formatCoinUnitsFromAsset(values.selectedAsset);
 
           // @dev next couple of lines are written like this, to keep typing equivalent during the conversion from BN to bigint
           const inputBigint = values.inputAmount?.bigintValue;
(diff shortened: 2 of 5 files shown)


src/components/ui/forms/BigIntInput.tsx
```diff
@@ -142,6 +144,12 @@ export function BigIntInput({
     [decimalPlaces, min, onChange],
   );
 
+  // if the parent Formik value prop changes, need to update the value
+  useEffect(() => {
+    if (!inputValue || inputValue === currentValue?.value) return;
+    if (currentValue?.bigintValue === 0n) setInputValue('');
+  }, [currentValue, inputValue]);
+
```
adamgall, line 152: Oof, I get anxious when I see changes in `BigIntInput`.

Why are these changes needed?

src/components/ui/modals/SendAssetsModal.tsx
```diff
@@ -70,18 +55,47 @@ export function SendAssetsModal({ close }: { close: () => void }) {
...
           destinationAddress: '',
+          selectedAsset: fungibleAssetsWithBalance[0],
+          inputAmount: undefined,
         }}
         onSubmit={submitSendAssets}
         validationSchema={sendAssetsValidationSchema}
       >
-        {({ errors }) => {
+        {({ errors, values, setFieldValue }) => {
+          console.log(values);
```
adamgall, line 84: Please remove console.log

src/components/ui/modals/SendAssetsModal.tsx
```diff
@@ -70,18 +55,47 @@ export function SendAssetsModal({ close }: { close: () => void }) {
...
   const sendAssetsValidationSchema = Yup.object().shape({
     destinationAddress: Yup.string().test(addressValidationTest),
+    selectedAsset: Yup.object()
+      .shape({
+        tokenAddress: Yup.string().required(),
+        token: Yup.object().shape({
+          name: Yup.string().required(),
+          symbol: Yup.string().required(),
+          decimals: Yup.number().required(),
+          logoUri: Yup.string().required(),
```
adamgall, line 65: Does this `logoUri` need to be here? AFAIK we don't ever take a "logo" as input from the user...

src/components/ui/forms/BigIntInput.tsx
DarksightKellar: @adamgall 
FWIW, the new optional `currentValue` is currently only set by the `SendAssetsModal`, so any regression can quickly be traced to it.

Now... for the reasoning behind this update (hopefully I don't do a crap job of explaining):

The `value` prop of `BigIntInput`'s wrapped `Input` is a string strictly for display purposes.

The `value` that Formik controls (and passes into `BigIntInput` via `{ ...field }` is instead a `BigIntValuePair`, which holds the true input data. This itself is swallowed pretty early on to set an initial `inputValue` state, never to be used again, it appears (I'm realising a potential bug here, which I'll address later).

This introduces a disconnect, as the underlying Formik `value` is never really reflected in the underlying `Input`'s `value` wrapped by `BigIntInput`. So, Formik value was getting updated just fine, but `inputValue` wasn't.

Then I noticed that `processValue` did the brunt of the work of making sure that the proper string is put into `processValue` each time the underlying `Input` triggers its `onChange`, which itself triggers the `onChange` prop passed into `BigIntInput`, which in the case of SendAssetsModal, calls formik's `setFieldValue` to update it's localised `value` state. So when the update happened because they user _typed_ into the field, all was good.

The only thing missing was if the formik `value` state is dynamically updated _outside_ the `BigIntInput`, a la `setFieldValue('inputAmount', { value: '0', bigintValue: 0n });` when selected asset changes. In this case `BigIntInput` is completely unaware that it needs to update itself. That's what this update does.

With the new `currentValue={values.inputAmount}`, whenever Formik's `values.inputAmount` is updated, _and_ the update wasn't from within `BigIntInput` (`inputValue !== currentValue?.value`), the useEffect hook is triggered. I've taken it a step further to only directly call `setInputValue` when the external update was intended to clear the input field (I experimented with calling `processValue` but concluded it was too risky).


Oh and the potential bug -- `value` prop of `BigIntInput` is a `bigint`, not `BigIntValuePair`, which is what Formik would pass into `BigIntInput`. Funny this didn't show up as a bug in my tests. I think I'll update to: 
```
{ ...field }, 
value={(field.value as BigIntValuePair)?.bigintValue}
```

src/components/ui/modals/SendAssetsModal.tsx
```diff
@@ -124,10 +142,14 @@ export function SendAssetsModal({ close }: { close: () => void }) {
                       <LabelWrapper label={t('amountLabel')}>
                         <BigIntInput
                           {...field}
-                          onChange={onChangeAmount}
-                          decimalPlaces={selectedAsset?.token?.decimals}
+                          value={(field.value as BigIntValuePair)?.bigintValue}
```
adamgall, line 145: Up above on line 140, can this be typed properly, not `FieldAttributes<any>`? I don't like the `any`.

src/components/ui/forms/BigIntInput.tsx
adamgall: Ok, I think i mostly followed that, thanks for the explanation.

I'm gonna say "ok" here and run with it.

Something on my mind though, is taking a look at these `Input` components again, and maybe cleaning them up.

I've written some pretty clean components like this for other projects, which AFAIK do the same functionality but in much simpler ways.

Checkout this `Input` component, which is made for entering "token amounts"
https://github.com/decent-labs/decent-token-vesting/blob/master/src/components/ui/Input.tsx#L43

It's used right here
https://github.com/decent-labs/decent-token-vesting/blob/master/src/components/TokenVests/New.tsx#L208

Then you can see that whenever the user changes the input of it, it results in this `BigNumber` being updated via a simple `useAmount` hook
https://github.com/decent-labs/decent-token-vesting/blob/master/src/components/TokenVests/New.tsx#L50

@decentdao/engineering does it make sense to look at how we've built these Inputs and maybe simplify them?

adamgall (dismissed):

Da-Colon (changes requested):
I'm getting an error when I navigate to Treasury Page.

- Multisig Safe
- Treasury has `0.04 sepEther`

image

Da-Colon (approved):
It loading now just fine 🤷

src/components/ui/modals/SendAssetsModal.tsx
```diff
@@ -70,24 +55,50 @@ export function SendAssetsModal({ close }: { close: () => void }) {
 
   const sendAssetsValidationSchema = Yup.object().shape({
```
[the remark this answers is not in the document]
DarksightKellar: @mudrila Yeah actually you're right. You mind creating a maintenance issue for this cleanup/optimisation point?

PR #1961, opened 2024-05-30
`1598` Support empty strings

### LINKED ISSUE #1598: Unable to pass "empty string" variable to function when crafting proposal
The proposal builder marks string inputs as "required", but in solidity there is no requirement that the string be present. Meaning, an "empty string" of zero bytes is valid input to a smart contract function that takes a string.

Currently, there's no way to pass an empty string in our proposal builder.

Da-Colon:
Fixes #1598

So I think I figured out a simple solution. If quotes are wrapping the text in the input field by a user simple remove the quotes.

I deployed and verified a simple contract to do some testing on Sepolia

Address: `0xEDFF6156e85ae11d2767Fc9Ec6E17352A928D0b7`

One function, a public state, and an event that check if its a empty string (for funsies).

I've tested:

`"foo"` -> foo
`'foo'` -> foo
`""` ->
`''` ->
`Foo'bar` -> `Foo'bar`
`"Foo'bar"` -> `Foo'bar`

Let me know yall's thoughts on this update

### DIFF
--- a/src/utils/crypto.ts
+++ b/src/utils/crypto.ts
@@ -46,6 +46,12 @@ export const encodeFunction = (
     } else if (param.endsWith(')')) {
       (parametersFixed!![tupleIndex!] as string[]).push(param.replace(')', ''));
       tupleIndex = undefined;
+    } else if (
+      (param.startsWith('"') && param.endsWith('"')) ||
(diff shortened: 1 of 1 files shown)


adamgall (approved):
love it

PR #1968, opened 2024-06-02
`1536` prevent execute if not next nonce

### LINKED ISSUE #1536: Execution order of multiple pending multisig transactions
When there are multiple pending proposals, they *must* be executed in order of their nonces. Attempting to execute a nonce-7 proposal while a nonce-6 is pending fails with a non-descriptive error. If you go execute nonce-6, then nonce-7, that works fine. Tested with up to 4 pending proposals, and the behaviour is easily reproducible. This is likely expected behaviour though.

_Originally posted by @DarksightKellar in https://github.com/decentdao/fractal-interface/issues/1516#issuecomment-2049950031_

Adam's Notes:
We should be (but are probably not) disabling a multisig transaction's "Execute" button if it's nonce is not the next one that can be executed.

Da-Colon:
Fixes #1536

I left two comments (I'll point them out) on some logic around nonce.

Basically just slightly reverted the change where safe.nonce is replaced by nextNonce. I've added a new property to the typing and place `nextNonce` as its own property so that we still have current nonce on hand that we can use to prevent execution of other proposals of higher nonce

### DIFF
--- a/src/components/pages/DaoSettings/components/Signers/modals/AddSignerModal.tsx
+++ b/src/components/pages/DaoSettings/components/Signers/modals/AddSignerModal.tsx
@@ -57,7 +57,7 @@ function AddSignerModal({
     address: Yup.string().test(addressValidationTest).test(newSignerValidationTest),
     nonce: Yup.number()
       .required()
-      .moreThan((!!safe && safe.nonce - 1) || 0),
+      .moreThan((!!safe && safe.nonce - 1) || 0), // ? @todo Is this correct? or should this be nextNonce?
     threshold: Yup.number().required(),
     thresholdOptions: Yup.array().of(Yup.number()).required(),
   });
@@ -67,7 +67,7 @@ function AddSignerModal({
       <Formik
         initialValues={{
           address: '',
-          nonce: safe?.nonce || 0,
+          nonce: safe?.nextNonce || 0,
           threshold: currentThreshold,
           thresholdOptions: Array.from({ length: signers.length + 1 }, (_, i) => i + 1),
         }}
--- a/src/components/pages/DaoSettings/components/Signers/modals/AddSignerModal.tsx
+++ b/src/components/pages/DaoSettings/components/Signers/modals/AddSignerModal.tsx
@@ -57,7 +57,7 @@ function AddSignerModal({
     address: Yup.string().test(addressValidationTest).test(newSignerValidationTest),
     nonce: Yup.number()
       .required()
-      .moreThan((!!safe && safe.nonce - 1) || 0), // ? @todo Is this correct? or should this be nextNonce?
+      .moreThan((!!safe && safe.nonce - 1) || 0),
     threshold: Yup.number().required(),
     thresholdOptions: Yup.array().of(Yup.number()).required(),
   });
--- a/src/components/ui/forms/CustomNonceInput.tsx
+++ b/src/components/ui/forms/CustomNonceInput.tsx
@@ -25,7 +25,7 @@ export function CustomNonceInput({
   } = useFractal();
   const { t } = useTranslation(['proposal', 'common']);
   const errorMessage =
-    nonce !== undefined && safe && nonce < safe.nonce ? t('customNonceError') : undefined;
+    nonce !== undefined && safe && nonce < safe.nonce ? t('customNonceError') : undefined; // ? @todo Is this correct? or should this also be minus 1?
 
   const tooltipContainer = useRef<HTMLDivElement>(null);
   if (dao?.isAzorius) return null;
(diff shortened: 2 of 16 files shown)


Da-Colon:
@decentdao/design @decentdao/product @adamgall

Two Things. 

image

- Thoughts on copy?
- Tooltip background is hard to see against the background of the app. This okay for now?

src/components/pages/DaoSettings/components/Signers/modals/AddSignerModal.tsx
```diff
@@ -57,7 +57,7 @@ function AddSignerModal({
     address: Yup.string().test(addressValidationTest).test(newSignerValidationTest),
     nonce: Yup.number()
       .required()
-      .moreThan((!!safe && safe.nonce - 1) || 0),
+      .moreThan((!!safe && safe.nonce - 1) || 0), // ? @todo Is this correct? or should this be nextNonce?
```
Da-Colon, line 60: I believe this should be the current nonce. since this just checks if the proposal could be created (someone may want to override a previously created nonce so shouldn't block that).

src/components/ui/forms/CustomNonceInput.tsx
```diff
@@ -25,7 +25,7 @@ export function CustomNonceInput({
   } = useFractal();
   const { t } = useTranslation(['proposal', 'common']);
   const errorMessage =
-    nonce !== undefined && safe && nonce < safe.nonce ? t('customNonceError') : undefined;
+    nonce !== undefined && safe && nonce < safe.nonce ? t('customNonceError') : undefined; // ? @todo Is this correct? or should this also be minus 1?
```
Da-Colon, line 28: Same as above. just some validation on the input that it shouldn't be less than the current nonce. I'll erase the comments in another commit

tomstuart123 (approved):
will leave to devs to answer your questions.

Send assets modal seems to be performing as desired when it comes to nonce proposals in pending though so approved from my side

tomstuart123 (approved):
see comment two above

xraystyle1980 (approved):
I'm not in love with the hard to view dark tooltip but I don't think it's getting lost or being mistaken for anything else. I say let's move forward with the dark tooltip for now.

src/providers/App/hooks/useSafeAPI.ts
```diff
@@ -209,10 +209,10 @@ class CachingSafeServiceClient extends SafeServiceClient {
     return value;
   }
 
-  async getSafeData(safeAddress: string): Promise<SafeInfoResponse> {
+  async getSafeData(safeAddress: string): Promise<SafeInfoResponse & { nextNonce: number }> {
```
adamgall, line 212: Can this type `Promise<SafeInfoResponse & { nextNonce: number }>` be written as `Promise<SafeInfoResponseWithGuard>`?

Does that type represent what the intention is here?

src/i18n/locales/en/proposal.json
```diff
@@ -111,6 +111,7 @@
   "customNonce": "Custom Nonce (advanced)",
   "customNonceTrimmed": "Custom Nonce",
   "customNonceTooltip": "Optionally set a custom proposal nonce to avoid collisions.",
+  "notActiveNonceTooltip": "Proposals must be executed in order based on their nonce.",
```
adamgall, line 114: To Tom's point, let's make this language more action oriented.

Something like: "Execute pending proposal with lower nonce before this one."

adamgall (approved):

src/providers/App/hooks/useSafeAPI.ts
Da-Colon: it sure can

src/components/ui/forms/CustomNonceInput.tsx
DarksightKellar: Theoretically should it not be possible to attempt to override _any_ currently active proposal out of a more-than-one number? If true, then specifically in custom nonce input, safe.nonce checks shouldn't even be happening.

DarksightKellar (approved):
Not tested manually but code looks good. Left a thought on your comment/question

src/components/ui/forms/CustomNonceInput.tsx
DarksightKellar: > The way to override current nonce and thus reject other proposals with same nonce would be via submitting "On Chain Rejection" - basically, empty transaction that executes from Safe with the same nonce as some existing proposal(s).

I like this. It doesn't look like there's any way to do this "rejection proposal" from within the app, is there? It'd be good to have, that way we can confidently validate the next nonce argument of any _new_ proposal, plus users will have a better UX for "cancelling" proposals.

adamgall: fwiw, _any_ transaction which is executed with a nonce (empty or not), will automatically nullify any other pending transactions with that same nonce.

PR #1969, opened 2024-06-03
`1568` nonce page refresh

### LINKED ISSUE #1568: Make sure correct nonce is being used on Send Assets proposal creation modal
Make sure Nonce is updated on Send Tokens proposal creation, too.

_Originally posted by @adamgall in https://github.com/decentdao/fractal-interface/issues/1516#issuecomment-2051991810_

Da-Colon:
Fixes #1568

Not sure if we should also cache and/or exclude routes. but move the implementation done in the New Proposals page to the `DAOController` and set it up to make the request whenever the url changes. 

The alternative to this, would be to refactor this hook back to the `isMounted` ref setup and add the hook to the particular pages/models/components that may create a proposal. But then we have to add it to any new pages/components that may need this features. Thoughts?

### DIFF
--- a/src/hooks/utils/useUpdateSafeData.ts
+++ b/src/hooks/utils/useUpdateSafeData.ts
@@ -1,0 +1,32 @@
+import { useRef, useEffect } from 'react';
+import { useLocation } from 'react-router-dom';
+import { useFractal } from '../../providers/App/AppProvider';
+import { useSafeAPI } from '../../providers/App/hooks/useSafeAPI';
+import { NodeAction } from '../../providers/App/node/action';
+
+export const useUpdateSafeData = () => {
+  const {
+    action,
+    node: { daoAddress },
+  } = useFractal();
+  const safeAPI = useSafeAPI();
+  const location = useLocation();
+  const prevPathname = useRef(location.pathname);
+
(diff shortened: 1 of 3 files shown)


tomstuart123 (approved):
will leave to devs to answer your questions.

Send assets modal seems to be performing as desired when it comes to nonce proposals in pending though so approved from my side

adamgall:
This is a good general solution to our problem, for sure.

A part of me is on the fence about whether this top-down approach is the right way, or if the approach you outlined as an alternative would be better.

"General solution that might not be very efficient" vs "precision solution that won't waste network calls".

adamgall (approved):
Adding approval because this does seem to work well, but would like to hear your thoughts @Da-Colon on if you think the alternative solution is "better". There are tradeoffs in both directions of course.

Da-Colon:
> Adding approval because this does seem to work well, but would like to hear your thoughts @Da-Colon on if you think the alternative solution is "better". There are tradeoffs in both directions of course.

Better as far as maybe more fine tuned say, when you open a model, or when you go to the final step to create a proposal, it makes the request so its the most up to date as possible. 

IMO  it doesn't need to be fine tuned right now.

PR #1974, opened 2024-06-03
Fix inconsistent menu hover styling

### LINKED ISSUE #1889: inconsistent hover styling on menu items
Screenshot 2024-05-22 at 11 53 50 AM
Screenshot 2024-05-22 at 11 53 59 AM
Screenshot 2024-05-22 at 11 53 40 AM
@decentdao/engineering inconsistent hover styling on menu items. Let's match all of those to the sidebar behavior.

_Originally posted by @nicolaus-sherrill in https://github.com/decentdao/decent-interface/discussions/1863#discussioncomment-9524718_

DarksightKellar:
Closes #1889 

Screenshot 2024-06-03 at 16 12 41
Screenshot 2024-06-03 at 16 12 49
Screenshot 2024-06-03 at 16 13 01
Screenshot 2024-06-03 at 16 13 09


Da-Colon (changes requested):
The Favorites List, Sort Dropdown, Proposals Filter:
- MenuItems aren't picking up the text color change on hover.

xraystyle1980 (approved):

xraystyle1980:
> The favorites list, sort dropdown, proposals filter and settings dropdown:
> 
> * MenuItems aren't picking up the text color change on hover.

*Text color does not change on menu item hover
[In Figma](https://www.figma.com/design/ur5hElxplkmlY9gnGKlwrm/Components-%26-Styles---Decent-Design-System-1.0.0?m=auto&node-id=2168-20014&t=1U7DVbWOBQZCmhow-1)

Da-Colon (approved):
My Mistake, It looks great ha!

src/components/ui/menus/OptionMenu/index.tsx
```diff
@@ -33,15 +33,17 @@ export function OptionMenu({
   ...rest
 }: OptionMenuProps) {
   const { t } = useTranslation(namespace);
+
   const menuList = (
     <MenuList
       borderWidth="1px"
       borderColor="neutral-3"
-      borderRadius="0.75rem"
+      borderRadius="0.58rem"
```
adamgall, line 41: Is this "0.58rem" value accurate? Seems oddly unique.

adamgall (approved):
Approved but please see question in this review

tomstuart123 (approved):
looks good

src/components/ui/menus/OptionMenu/index.tsx
DarksightKellar: Ah yes, you're right in that it's uncomfortably unique. I'd meant to comment on this with screenshots. Unfortunately I'm not at my desk today so I'll try to describe with words:

I've noticed in general that when there's a box-within-border setup, both require a border radius, otherwise the inside box non-rounded corners overflow the rounded border edges. Buut simply using the same border radius value causes the inner container corner to "pull away" from the border corner, leaving visible gaps on all rounded corners.

(I think it's just a limitation of the math of circles and the fact that technically there're different radii involved.)

There's been one other instance I've encountered this problem, but it was only barely noticeable when zoomed in. So I was fine leaving as-is.

In this case though, it was too ugly to leave alone, so I've forced the inner container curve to "stretch" out to touch the outer border curve.

adamgall: I understand what you're saying, but I'm still stuck on it... I know this code is merged already but I wonder if there's something weird about the HTML structure.

PR #1975, opened 2024-06-03
`1931` Part 2: `1710` Cross network favorites

### LINKED ISSUE #1710: Cross network favorites
recording

Can we make favorites viewable across networks? Currently seeing no favorites if on a different network. Means we're asking users to manage multiple favorites lists.

<details><summary>Browser metadata</summary>

```
Path:      /home?dao=eth:0xD26c85D435F02DaB8B220cd4D2d398f6f646e235
Browser:   Firefox 125.0 on Mac OS 10.15
Viewport:  1402 x 912 @2x
Language:  en-US
Cookies:   Enabled
```

[Open in BrowserStack](https://www.browserstack.com/user/try-live?url=https%3A%2F%2Fdeploy-preview-1578.app.dev.fractalframework.xyz%2Fhome%3Fdao%3Deth%3A0xD26c85D435F02DaB8B220cd4D2d398f6f646e235&os=OS%20X&os_version=Catalina&browser=Firefox&browser_version=125.0&resolution=1402x912&speed=1&start=true&ref=netlify-source)

</details>

[Open Deploy Preview](https://deploy-preview-1578.app.dev.fractalframework.xyz/home?dao=eth:0xD26c85D435F02DaB8B220cd4D2d398f6f646e235) · [Mark as Resolved](https://app.netlify.com/cdp/resolve?deployID=663b8f9fcd44440008320663&commentID=663baeeacc87930557124f80&resolution=resolved)

_Originally posted by @nicolaus-sherrill in https://github.com/decentdao/fractal-interface/issues/1578#issuecomment-2101014045_

Da-Colon:
Fixes #1710

Kinda a weird one to fit in the middle but wanted this work to be done prior to the Local Storage migration actually is done so that the typing is setup.

This PR handles adding the network prefix to the cached favorite string. It also adds some helper functions to get the network prefix and chain id via the network prefix to allow for the UI around the favorites.

### DIFF
--- a/src/components/ui/icons/FavoriteIcon.tsx
+++ b/src/components/ui/icons/FavoriteIcon.tsx
@@ -4,6 +4,7 @@ import { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { getAddress } from 'viem';
 import { useAccountFavorites } from '../../../hooks/DAO/loaders/useFavorites';
+import { useNetworkConfig } from '../../../providers/NetworkConfig/NetworkConfigProvider';
 
 interface Props extends BoxProps {
   safeAddress: string;
@@ -11,9 +12,11 @@ interface Props extends BoxProps {
 
 export function FavoriteIcon({ safeAddress, ...rest }: Props) {
   const { favoritesList, toggleFavorite } = useAccountFavorites();
+  const { addressPrefix } = useNetworkConfig();
   const isFavorite = useMemo(
-    () => (!!safeAddress ? favoritesList.includes(getAddress(safeAddress)) : false),
-    [favoritesList, safeAddress],
+    () =>
+      !!safeAddress ? favoritesList.includes(addressPrefix + ':' + getAddress(safeAddress)) : false,
+    [favoritesList, safeAddress, addressPrefix],
   );
   const { t } = useTranslation();
   return (
--- a/src/components/ui/icons/FavoriteIcon.tsx
+++ b/src/components/ui/icons/FavoriteIcon.tsx
@@ -15,7 +15,7 @@ export function FavoriteIcon({ safeAddress, ...rest }: Props) {
   const { addressPrefix } = useNetworkConfig();
   const isFavorite = useMemo(
     () =>
-      !!safeAddress ? favoritesList.includes(addressPrefix + ':' + getAddress(safeAddress)) : false,
+      !!safeAddress ? favoritesList.includes(`${addressPrefix}:${getAddress(safeAddress)}`) : false,
     [favoritesList, safeAddress, addressPrefix],
   );
   const { t } = useTranslation();
--- a/src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
+++ b/src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
@@ -7,20 +7,20 @@ import { DAO_ROUTES } from '../../../../constants/routes';
 import { useGetDAOName } from '../../../../hooks/DAO/useGetDAOName';
 
 export interface SafeMenuItemProps {
-  network: string;
   address: string;
   showAddress?: boolean;
   onClick?: () => void;
 }
 
-export function SafeMenuItem({ network, address }: SafeMenuItemProps) {
-  const { daoName } = useGetDAOName({ address: getAddress(address) });
+export function SafeMenuItem({ address }: SafeMenuItemProps) {
+  const [networkPrefix, daoAddress] = address.split(':');
+  const { daoName } = useGetDAOName({ address: getAddress(daoAddress) });
   const navigate = useNavigate();
 
   const { t } = useTranslation('dashboard');
 
   const onClickNav = () => {
-    navigate(DAO_ROUTES.dao.relative(network, address));
+    navigate(DAO_ROUTES.dao.relative(networkPrefix, daoAddress));
   };
 
   return (
--- a/src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
+++ b/src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
@@ -8,19 +8,19 @@ import { useGetDAOName } from '../../../../hooks/DAO/useGetDAOName';
 
 export interface SafeMenuItemProps {
   address: string;
+  network: string;
   showAddress?: boolean;
   onClick?: () => void;
 }
 
-export function SafeMenuItem({ address }: SafeMenuItemProps) {
-  const [networkPrefix, daoAddress] = address.split(':');
-  const { daoName } = useGetDAOName({ address: getAddress(daoAddress) });
+export function SafeMenuItem({ address, network }: SafeMenuItemProps) {
+  const { daoName } = useGetDAOName({ address: getAddress(address) });
   const navigate = useNavigate();
 
   const { t } = useTranslation('dashboard');
 
   const onClickNav = () => {
-    navigate(DAO_ROUTES.dao.relative(networkPrefix, daoAddress));
+    navigate(DAO_ROUTES.dao.relative(network, address));
   };
 
   return (
--- a/src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
+++ b/src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
@@ -1,10 +1,13 @@
-import { Box, Button, MenuItem, Text } from '@chakra-ui/react';
-import { Star } from '@phosphor-icons/react';
+import { Box, Button, Flex, Image, MenuItem, Spacer, Text } from '@chakra-ui/react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
 import { getAddress } from 'viem';
 import { DAO_ROUTES } from '../../../../constants/routes';
 import { useGetDAOName } from '../../../../hooks/DAO/useGetDAOName';
+import useAvatar from '../../../../hooks/utils/useAvatar';
+import useDisplayName from '../../../../hooks/utils/useDisplayName';
+import { getNetworkIcon } from '../../../../utils/url';
+import Avatar from '../../page/Header/Avatar';
 
 export interface SafeMenuItemProps {
   address: string;
@@ -16,6 +19,8 @@ export interface SafeMenuItemProps {
 export function SafeMenuItem({ address, network }: SafeMenuItemProps) {
   const { daoName } = useGetDAOName({ address: getAddress(address) });
   const navigate = useNavigate();
+  const { displayName: accountDisplayName } = useDisplayName(address);
+  const avatarURL = useAvatar(accountDisplayName);
 
   const { t } = useTranslation('dashboard');
 
@@ -37,12 +42,23 @@ export function SafeMenuItem({ address, network }: SafeMenuItemProps) {
         justifyContent="flex-start"
         gap={2}
       >
-        <Star
-          size="1.5rem"
-          weight="fill"
+        <Avatar
+          address={address}
+          url={avatarURL}
         />
+        <Flex flexDir="column">
+          <Text
+            color="white-0"
+            textStyle="button-base"
+          >
+            {daoName ?? t('loadingFavorite')}
+          </Text>
+        </Flex>
 
-        <Text color={daoName ? 'white-0' : 'neutral-6'}>{daoName ?? t('loadingFavorite')}</Text>
+        <Spacer />
+
+        {/* Network Icon */}
+        <Image src={getNetworkIcon(network)} />
       </MenuItem>
     </Box>
   );
--- a/src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
+++ b/src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
@@ -2,11 +2,13 @@ import { Box, Button, Flex, Image, MenuItem, Spacer, Text } from '@chakra-ui/rea
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
 import { getAddress } from 'viem';
+import { useSwitchChain } from 'wagmi';
 import { DAO_ROUTES } from '../../../../constants/routes';
 import { useGetDAOName } from '../../../../hooks/DAO/useGetDAOName';
 import useAvatar from '../../../../hooks/utils/useAvatar';
 import useDisplayName from '../../../../hooks/utils/useDisplayName';
-import { getNetworkIcon } from '../../../../utils/url';
+import { useNetworkConfig } from '../../../../providers/NetworkConfig/NetworkConfigProvider';
+import { getChainIdFromPrefix, getNetworkIcon } from '../../../../utils/url';
 import Avatar from '../../page/Header/Avatar';
 
 export interface SafeMenuItemProps {
@@ -17,6 +19,9 @@ export interface SafeMenuItemProps {
 }
 
 export function SafeMenuItem({ address, network }: SafeMenuItemProps) {
+  const { addressPrefix } = useNetworkConfig();
+  const { switchChain } = useSwitchChain();
+
   const { daoName } = useGetDAOName({ address: getAddress(address) });
   const navigate = useNavigate();
   const { displayName: accountDisplayName } = useDisplayName(address);
@@ -25,7 +30,16 @@ export function SafeMenuItem({ address, network }: SafeMenuItemProps) {
   const { t } = useTranslation('dashboard');
 
   const onClickNav = () => {
-    navigate(DAO_ROUTES.dao.relative(network, address));
+    if (addressPrefix !== network) {
+      switchChain(
+        { chainId: getChainIdFromPrefix(network) },
+        {
+          onSuccess: () => navigate(DAO_ROUTES.dao.relative(network, address)),
+        },
+      );
+    } else {
+      navigate(DAO_ROUTES.dao.relative(network, address));
+    }
   };
 
   return (
--- a/src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
+++ b/src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
@@ -30,7 +30,7 @@ export function SafeMenuItem({ address, network }: SafeMenuItemProps) {
 
   const { displayName: accountDisplayName } = useDisplayName(
     address,
-    undefined,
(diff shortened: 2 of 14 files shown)


src/utils/url.ts
```diff
@@ -28,3 +28,36 @@ export const validateENSName = (ensAddress?: string): boolean => {
...
+export const getChainIdFromNetworkPrefix = (_networkPrefix: string): number => {
+  return Number(
+    Object.keys(networkPrefix).find(key => networkPrefix[parseInt(key)] === _networkPrefix),
+  );
+};
+
+export const getNetworkIcon = (_networkPrefix: string): string => {
+  const chainId = getChainIdFromNetworkPrefix(_networkPrefix);
+  return networkIcons[chainId];
+};
```
Da-Colon, line 63: This is the main part of this update. Added helper functions to help get/retrieve the prefix and icons for given chainId. If there is more scaleable way to do this by using the information in `NetworkConfig` files, I'm open to suggestions.

adamgall: We do already store the network prefix strings and network icons in the `NetworkConfig` files, so we need to find a way to reuse those, not duplicate them here. It will be no bueno if we'll be required to remember to come update this file when a new network is added.

Da-Colon: Yeah I didn't feel good about it. I'll take another look.

Da-Colon: oof should have done this to begin with. throws when network isn't found (it shouldn't ever throw)

src/components/ui/icons/FavoriteIcon.tsx
```diff
@@ -4,16 +4,19 @@ import { useMemo } from 'react';
...
 
 export function FavoriteIcon({ safeAddress, ...rest }: Props) {
   const { favoritesList, toggleFavorite } = useAccountFavorites();
+  const { addressPrefix } = useNetworkConfig();
   const isFavorite = useMemo(
-    () => (!!safeAddress ? favoritesList.includes(getAddress(safeAddress)) : false),
-    [favoritesList, safeAddress],
+    () =>
+      !!safeAddress ? favoritesList.includes(addressPrefix + ':' + getAddress(safeAddress)) : false,
+    [favoritesList, safeAddress, addressPrefix],
```
adamgall, line 19: This line is stumping me, but probably just because it's the first thing I'm looking at and have no context yet.

But it's stumping me because this doesn't look like cross-network-friendly code. Looks like it'll only truthy `isFavorite` if the given address is a favorite _on the current network_.

src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
```diff
@@ -1,26 +1,45 @@
...
+    if (addressPrefix !== network) {
+      switchChain(
+        { chainId: getChainIdFromPrefix(network) },
+        {
+          onSuccess: () => navigate(DAO_ROUTES.dao.relative(network, address)),
+        },
+      );
+    } else {
+      navigate(DAO_ROUTES.dao.relative(network, address));
+    }
```
adamgall, line 50: Okkkkk I see u

src/components/ui/menus/SafesMenu/SafesList.tsx
```diff
@@ -25,14 +23,13 @@ export function SafesList() {
...
-            overflowY="scroll"
             className="scroll-dark"
           >
             {favoritesList.map(favorite => (
               <SafeMenuItem
                 key={favorite}
-                network={addressPrefix}
-                address={favorite}
+                network={favorite.split(':')[0]}
+                address={favorite.split(':')[1]}
```
adamgall, line 32: We need to do some validation on this _user input_ before attempting to use it like this. Otherwise it'll throw when someone ruins a favorite in localstorage and Sentry will yell at us

src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
```diff
@@ -1,26 +1,53 @@
...
+
+  const { daoName } = useGetDAOName({
+    address: getAddress(address),
+    chainId: getChainIdFromPrefix(network),
+  });
   const navigate = useNavigate();
 
+  const { displayName: accountDisplayName } = useDisplayName(
+    address,
+    undefined,
```
adamgall, line 33: Just say `false` here plz

Da-Colon:
Found a much better solution. (kinda a duh. moment). 

Soo. one thing is missing here. pulling the DAO's name on a different network. I have it almost setup but the last thing is targeting the right registry contract. I tried to find a way to make this work but ultimately I started rubbing up against the updates @adamgall is making in the typechain rework. I need those updates ha! then this will be a piece of cake.

src/pages/home/SafeDisplayRow.tsx
```diff
@@ -2,32 +2,52 @@ import { Flex, Image, Show, Spacer, Text } from '@chakra-ui/react';
...
+  });
+
   const navigate = useNavigate();
 
   const { t } = useTranslation('dashboard');
 
-  const { displayName: accountDisplayName } = useDisplayName(address);
+  const { displayName: accountDisplayName } = useDisplayName(
+    address,
+    undefined,
```
adamgall, line 30: Just say `false` here plz

src/hooks/utils/useDisplayName.ts
```diff
@@ -14,13 +14,14 @@ export const createAccountSubstring = (account: string) => {
  *
  * This is intended to be used for NON DAO display names.  If you would like to get the
  * display name for a DAO, use the useDAOName hook instead.
+ * @todo Should switch to object for props
  */
-const useDisplayName = (account?: string | null, truncate?: boolean) => {
+const useDisplayName = (account?: string | null, truncate?: boolean, chainId?: number) => {
```
Da-Colon, line 19: I would prefer these args be an object but didn't want to fill up this review with this change to all the other files

src/components/ui/menus/SafesMenu/SafesList.tsx
adamgall: Ooh better yet, can we get a dedicated set of functions to "encode" and "decode" an object `{ network: "eth", address: "0x123 }` into a string of format `eth:0x123`

And those functions do our validation

src/pages/home/AllSafesDrawer.tsx
```diff
@@ -135,11 +132,11 @@ export function AllSafesDrawer({ isOpen, onClose }: AllSafesDrawerProps) {
...
         </DrawerHeader>
         <DrawerBody padding="0">
-          {getValue({ cacheName: CacheKeys.FAVORITES }).map((favorite: string) => (
+          {(getValue({ cacheName: CacheKeys.FAVORITES }) || []).map((favorite: string) => (
             <SafeDisplayRow
               key={favorite}
-              network={addressPrefix}
-              address={favorite}
+              address={favorite.split(':')[1]}
+              network={favorite.split(':')[0]}
```
adamgall, line 139: This would be another place where some validation would be good.

src/pages/home/MySafes.tsx
```diff
@@ -52,8 +50,8 @@ export function MySafes() {
             {favoritesToShow.map(favorite => (
               <SafeDisplayRow
                 key={favorite}
-                network={addressPrefix}
-                address={favorite}
+                address={favorite.split(':')[1]}
+                network={favorite.split(':')[0]}
```
adamgall, line 54: And here

src/pages/home/SafeDisplayRow.tsx
```diff
@@ -2,32 +2,52 @@ import { Flex, Image, Show, Spacer, Text } from '@chakra-ui/react';
...
+import { getChainIdFromPrefix, getChainName, getNetworkIcon } from '../../utils/url';
 
 export function SafeDisplayRow({ address, network, onClick, showAddress }: SafeMenuItemProps) {
-  const { daoName } = useGetDAOName({ address: getAddress(address) });
+  const { addressPrefix } = useNetworkConfig();
+  const { switchChain } = useSwitchChain();
+
+  const { daoName } = useGetDAOName({
+    address: getAddress(address),
+    chainId: getChainIdFromPrefix(network),
```
adamgall, line 21: If this throws, do we want to catch it in here?

src/pages/home/SafeDisplayRow.tsx
```diff
@@ -2,32 +2,52 @@ import { Flex, Image, Show, Spacer, Text } from '@chakra-ui/react';
...
+
   const navigate = useNavigate();
 
   const { t } = useTranslation('dashboard');
 
-  const { displayName: accountDisplayName } = useDisplayName(address);
+  const { displayName: accountDisplayName } = useDisplayName(
+    address,
+    undefined,
+    getChainIdFromPrefix(network),
```
adamgall, line 31: If this throws, do we want to catch it in here?

src/pages/home/SafeDisplayRow.tsx
```diff
@@ -67,9 +87,9 @@ export function SafeDisplayRow({ address, network, onClick, showAddress }: SafeM
 
       {/* Network Icon */}
       <Flex gap="0.5rem">
-        <Image src={networkConfig.nativeTokenIcon} />
+        <Image src={getNetworkIcon(network)} />
         <Show above="md">
-          <Text>{networkConfig.chain.name}</Text>
+          <Text>{getChainName(network)}</Text>
```
adamgall, line 92: What happens when these throw?

src/utils/url.ts
```diff
@@ -28,3 +28,41 @@ export const validateENSName = (ensAddress?: string): boolean => {
     return false;
   }
 };
+
+export const addNetworkPrefix = (address: string, chainId: number): string => {
+  const network = Object.values(networks).find(_network => _network.chain.id === chainId);
+  if (!network) {
+    throw new Error(`No network found for chainId ${chainId}`);
+  }
+  return `${network?.addressPrefix}:${address}`;
```
adamgall, line 37: Do you need the `?` on this `network`?

src/utils/url.ts
```diff
@@ -28,3 +28,41 @@ export const validateENSName = (ensAddress?: string): boolean => {
...
+};
+
+export const getNetworkIcon = (_networkPrefix: string): string => {
+  const network = Object.values(networks).find(
+    _network => _network.addressPrefix === _networkPrefix,
+  );
+  if (!network) {
+    throw new Error(`No network found for networkPrefix ${_networkPrefix}`);
+  }
+  return network?.nativeTokenIcon;
```
adamgall, line 47: Do you need the `?`?

src/utils/url.ts
```diff
@@ -28,3 +28,41 @@ export const validateENSName = (ensAddress?: string): boolean => {
...
+};
+
+export const getChainName = (_networkPrefix: string): string => {
+  const network = Object.values(networks).find(
+    _network => _network.addressPrefix === _networkPrefix,
+  );
+  if (!network) {
+    throw new Error(`No network found for chainId ${_networkPrefix}`);
```
adamgall, line 55: This error should say `No chain found for network blah`

src/utils/url.ts
```diff
@@ -28,3 +28,41 @@ export const validateENSName = (ensAddress?: string): boolean => {
...
+  }
+  return network.chain.name;
+};
+
+export const getChainIdFromPrefix = (_networkPrefix: string): number => {
+  const network = Object.values(networks).find(
+    _network => _network.addressPrefix === _networkPrefix,
+  );
+  if (!network) {
+    throw new Error(`No network found for chainId ${_networkPrefix}`);
```
adamgall, line 65: `No chainId found for network prefix blah`

adamgall (changes requested):
Nice work here, I have a couple of questions n comments

src/components/ui/icons/FavoriteIcon.tsx
Da-Colon: The `FavoriteIcon` component is now only being used on the DAO Dashboard and Organization Pages which are network specific. I've removed this from the Favorites menu (with Design's permission)

src/components/ui/menus/SafesMenu/SafesList.tsx
Da-Colon: Woot I did it. I'll point out the two function

src/utils/address.ts
```diff
@@ -0,0 +1,13 @@
...
+  const [prefix, addressWithoutPrefix] = address.split(':');
+  return {
+    network: prefix,
+    address: addressWithoutPrefix,
+  };
+};
+
+export const encodePrefixedAddress = (address: Address, network: string) => {
+  return `${network}:${address}`;
+};
```
Da-Colon, line 13: This what you mean? @adamgall

adamgall, line 13: Ok yes! This is a great start, but I also wanted to see some validation in here.

Like, `addressWithoutPrefix` will be `undefined` if `address` doesn't contain a "`:`".

src/pages/home/AllSafesDrawer.tsx
```diff
@@ -138,8 +136,7 @@ export function AllSafesDrawer({ isOpen, onClose }: AllSafesDrawerProps) {
           {(getValue({ cacheName: CacheKeys.FAVORITES }) || []).map((favorite: string) => (
             <SafeDisplayRow
               key={favorite}
-              network={addressPrefix}
-              address={favorite}
+              {...decodePrefixedAddress(favorite)}
```
Da-Colon, line 139: Do you have a preference about doing it like this?

adamgall (changes requested):

src/pages/home/AllSafesDrawer.tsx
adamgall: I'm cool with this

src/utils/address.ts
adamgall: Yes it's the start of what I meant!

Da-Colon:
As far as the throw errors. I don't believe it will EVER hit here. Unsupported networks are caught way up the food chain. I did this more or less so I didn't have to type our a `null` or `undefined`. Since these settings come from hardcoded settings it shouldn't ever throw

src/components/ui/icons/FavoriteIcon.tsx
adamgall: Ooo ok that makes sense

src/utils/address.ts
Da-Colon: ah I see. wanted a clean return instead of `undefined` but I can do that.

adamgall: I think if you `throw` it doesn't add `undefined` to the return type

adamgall:
> As far as the throw errors. I don't believe it will EVER hit here. Unsupported networks are caught way up the food chain. I did this more or less so I didn't have to type our a `null` or `undefined`. Since these settings come from hardcoded settings it shouldn't ever throw

Which comment was this in reference to?

Da-Colon:
> > As far as the throw errors. I don't believe it will EVER hit here. Unsupported networks are caught way up the food chain. I did this more or less so I didn't have to type our a `null` or `undefined`. Since these settings come from hardcoded settings it shouldn't ever throw
> 
> Which comment was this in reference to?

image


these comments

src/pages/home/SafeDisplayRow.tsx
DarksightKellar: I feel like these (and `getChainIdFromPrefix` usage above) could be wrapped a `useMemo`, where we could gracefully handle exceptions and return some acceptable default/error data.

DarksightKellar (changes requested):
Looking good here too! Had a couple thoughts in the review, if you could check 'em out

src/utils/address.ts
```diff
@@ -0,0 +1,16 @@
+import { Address } from 'viem';
+
+export const decodePrefixedAddress = (address: string) => {
+  const [prefix, addressWithoutPrefix] = address.split(':');
+  if (!prefix || !addressWithoutPrefix) {
+    throw new Error('Invalid address format');
+  }
+  return {
+    network: prefix,
+    address: addressWithoutPrefix,
```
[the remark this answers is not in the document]
Da-Colon: hmm. Might be overhead for this PR. as the address it returns goes through `getAddress` function where its used else where. so would doubling up

Da-Colon:
> I feel like these (and getChainIdFromPrefix usage above) could be wrapped a useMemo, where we could gracefully handle exceptions and return some acceptable default/error data.

So similar to my answer to @adamgall is this should never throw here. As the app won't let you connect to a network that isn't supported and since we are directly using the same configuration files.

So IMO, its not needed. But if yall (@adamgall had similar thoughts) are not satisfied with my answer and can figure out the changes needed to make this happen, Though hard to test since the app itself won't let you switch to an unsupported network. 

image

adamgall:
> > I feel like these (and getChainIdFromPrefix usage above) could be wrapped a useMemo, where we could gracefully handle exceptions and return some acceptable default/error data.
> 
> 
> 
> So similar to my answer to @adamgall is this should never throw here. As the app won't let you connect to a network that isn't supported and since we are directly using the same configuration files.
> 
> 
> 
> So IMO, its not needed. But if yall (@adamgall had similar thoughts) are not satisfied with my answe, then I can figure out the changes needed to make this happen, Though hard to test since the app itself won't let you switch to an unsupported network. 
> 
> 
> 
> image
> 
>  

Try/catching errors like this (ones that we know "should never happen") are precisely the point.

I'd like us to get into the habit of building more bulletproof code.

If there's a possibility for a line of code to throw (when evaluating it without any extra context) (and there is in this instance) let's catch it and bubble it up.

adamgall:
@Da-Colon j/w, do you plan on doing anything more with this PR? https://decentdao.slack.com/archives/C06QZ68L5SA/p1718121588983869

Da-Colon:
> @adamgall Yessir. I'm creating a
> 
> Placing an error Boundary with a Fallback UI, around Where this might fail.
> 
> localhost_3005_(Desktop)

@decentdao/product @decentdao/design Yall good with the look and copy?

Da-Colon:
@DarksightKellar I didn't add the `useMemo` as suggested with the error boundary this handles this error as is

DarksightKellar (approved):
> @DarksightKellar I didn't add the `useMemo` as suggested with the error boundary this handles this error as is, wdyt?
 

@Da-Colon Yhup, me likey.

adamgall (approved):

PR #1986, opened 2024-06-10
`1916` Add insufficient funds error

### LINKED ISSUE #1916: Don't silently fail when attempting to create transaction and connect walled doesn't have enough funds for gas
When the user is connected with a wallet that doesn't have enough funds for gas, the app will silently fail when clicking the transaction action button.

We should be showing an error toast, ideally with an appropriate error message.

At least, that's what happens on my setup: Chrome / Frame

Da-Colon:
Fixes #1916

### DIFF
--- a/src/hooks/utils/useTransaction.ts
+++ b/src/hooks/utils/useTransaction.ts
@@ -42,6 +42,7 @@ const useTransaction = () => {
         .then(([txReceipt, toastID]) => {
           toast.dismiss(toastID);
           if (txReceipt.status === 0) {
+            console.log("🚀 ~ txReceipt:", txReceipt)
             toast.error(params.failedMessage);
             if (params.failedCallback) params.failedCallback();
           } else if (txReceipt.status === 1) {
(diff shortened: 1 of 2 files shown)


DarksightKellar (approved):
lgtm, but leftover log. Also, how do I test this properly? Coinbase doesn't even allow me to continue if insufficient funds

tomstuart123:
Similar. I read the github issues but still unclear on the prior vs. intended behaviour? Also if this is just a frame wallet issue or with all wallets.

Could you map the test cases to replicate?

Da-Colon:
> Similar. I read the github issues but still unclear on the prior vs. intended behaviour? Also if this is just a frame wallet issue or with all wallets.
> 
> Could you map the test cases to replicate?

Yeah, Metamask prevents the same thing. To recreate, as far as I know you need frame. 

Prior to change: (With Frame): when attempting to 'execute' a transaction with an account with 0 ether (sepolia in this case), Frame will pop up as normal and when you sign, just the generic Error toast was used `Unexpected Error~~~` Instead of a more clear message on the toast. 

Now: we capture this particular error code and now show the user the new message that `Inefficient funds~~`


## Test Case
(Multisig Safe, Frame Wallet)
- Add an account to the Safe that has 0 ETH
- Then create a proposal (like change name)
- Attempt to execute the transaction
- Should now see new error message for Frame Wallets (or any wallet that prevents you)

tomstuart123:
Thanks @decentdao/design / @decentdao/engineering - do any of you already have a Frame test wallet setup to do a product UX test? 

If not, nw as I can download it and have a crack at setting it up to do it myself.

adamgall:
i haz frame, i willz test

adamgall (approved):
Nice, it works for me w/ Frame.

Issue #1511, opened 2024-04-01
Execute Button Go away after executing a multisig transaction


Da-Colon:
While this is kinda known issue. as it takes some time before Safe Transaction Service to index the transaction. We do have some polling in place. But we may want to relook at what we can do here. @adamgall changes to the loading proposals could help fix this as we can look into only updating a specific proposal rather than going refreshing all the proposals

![Image](https://github.com/decentdao/fractal-interface/assets/38386583/e404e184-afe3-4ec9-8cf1-7a5bd005909c)

adamgall:
@Da-Colon is there a typo in the title? What's the issue here?

adamgall:
Closing due to https://github.com/decentdao/decent-interface/pull/1990#issuecomment-2163179907

PR #1990, opened 2024-06-11
Fix execution button delay

### LINKED ISSUE #1558: Proposal Page - Time Misalignment between Toaster execution alert and the execution button
See video at 5.23 and 3.23

https://www.loom.com/share/3af51acc15ae4f3282c3287eebcc44d8

When the toaster pops up showing a 'executed successfully' for a brief second the button on the proposal page (currently greyed out as it should be) will light up orange again. If the user clicks it, it will error (as expected) ... but why is it lighting up at all (i.e. prompting a user action)

There's clearly a lag when the UI hasn't fully caught up with the toaster on the current proposal status. 

This lag maybe impossible to fix, ...but lets explore removing the 'light up' of the button. The button should remain grey until an actual transaction possibility is on the cards

DarksightKellar:
Closes #1558

This PR patches up the time delay between when the app indicates a proposal timelock and/or execute transaction has been successfully ran and when the data actually indicates as much.

We have achieved this by simply keeping the execute button disabled after the success toast. If the user chooses to stay on the screen, eventually the node data will be updated, and the button will disappear.

In the case of timelock, the button will re-enable itself after timelock when the proposal becomes executable.

I expect, if the user refreshes the page in that small window where the node hasn't reflected the updated proposal state, the button will then be clickable -- not much we can do in that case beyond localstorage caching, which we _have_ discussed. Problem is multisig safe proposals are currently not cached as they're loaded all at once every time, and choosing to selectively cache them will take a bit more thought. 

There's a separate task for this caching [here](https://github.com/orgs/decentdao/projects/32/views/3?pane=issue&itemId=65769569), if we end up deciding it's something worth doing.

### DIFF
--- a/src/components/Proposals/MultisigProposalDetails/TxActions.tsx
+++ b/src/components/Proposals/MultisigProposalDetails/TxActions.tsx
@@ -125,9 +125,7 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
         failedMessage: t('failedExecute', { ns: 'transaction' }),
         pendingMessage: t('pendingExecute', { ns: 'transaction' }),
         successMessage: t('successExecute', { ns: 'transaction' }),
-        successCallback: async () => {
-          await loadSafeMultisigProposals();
-        },
+        successCallback: loadSafeMultisigProposals,
       });
     } catch (e) {
       logError(e, 'Error occurred during transaction execution');
@@ -145,6 +143,7 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
         return;
       }
       const safeContract = GnosisSafeL2__factory.connect(safe.address, signerOrProvider);
+
       const safeTx = buildSafeTransaction({
         ...multisigTx,
         to: getAddress(multisigTx.to),
@@ -152,6 +151,7 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
         data: multisigTx.data as Hex | undefined,
         operation: multisigTx.operation as 0 | 1,
       });
+
       const signatures = buildSignatureBytes(
         multisigTx.confirmations.map(confirmation => {
           if (!isHex(confirmation.signature)) {
@@ -163,6 +163,7 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
           };
         }),
       );
+
       contractCall({
         contractFn: () =>
           safeContract.execTransaction(
@@ -206,7 +207,7 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
       action: () => Promise<any>;
       text: string;
       pageTitle: string;
-      icon: undefined | JSX.Element;
+      icon?: JSX.Element;
     };
   };
 
@@ -215,7 +216,6 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
       action: signTransaction,
       text: 'approve',
       pageTitle: 'signTitle',
-      icon: undefined,
     },
     [FractalProposalState.EXECUTABLE]: {
       action: executeTransaction,
@@ -227,13 +227,11 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
       action: timelockTransaction,
       text: 'timelock',
       pageTitle: 'timelockTitle',
-      icon: undefined,
     },
     [FractalProposalState.TIMELOCKED]: {
       action: async () => {},
       text: 'execute',
       pageTitle: 'executeTitle',
-      icon: undefined,
     },
   };
   const isActiveNonce = !!safe && multisigTx.nonce === safe.nonce;
@@ -243,7 +241,7 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
   return (
     <ContentBox containerBoxProps={{ bg: BACKGROUND_SEMI_TRANSPARENT }}>
       <Flex justifyContent="space-between">
-        <Text textStyle="text-lg-mono-medium">{t(buttonProps[proposal.state!].pageTitle)}</Text>
+        <Text>{t(buttonProps[proposal.state!].pageTitle)}</Text>
         <ProposalCountdown proposal={proposal} />
       </Flex>
       <Box marginTop={4}>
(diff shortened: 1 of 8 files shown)


adamgall:
My take is that trying to handle that small timeframe window of a user potentially refreshing the page, using some local storage caching, is way overkill.

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
```diff
@@ -30,12 +31,25 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
   const signerOrProvider = useSignerOrProvider();
   const safeAPI = useSafeAPI();
 
+  const [isSubmitDisabled, setIsSubmitDisabled] = useState(false);
+
+  let wasTimelocked = useRef(
```
adamgall, line 36: `const`, no?

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
```diff
@@ -181,6 +199,7 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
         pendingMessage: t('pendingExecute', { ns: 'transaction' }),
         successMessage: t('successExecute', { ns: 'transaction' }),
         successCallback: async () => {
+          setIsSubmitDisabled(true);
```
adamgall, line 202: Do we maybe want to set this property when the user initially clicks the "execute" button? or does it need to be in this `successCallback` for a specific reason?

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
```diff
@@ -125,8 +139,9 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
         failedMessage: t('failedExecute', { ns: 'transaction' }),
         pendingMessage: t('pendingExecute', { ns: 'transaction' }),
         successMessage: t('successExecute', { ns: 'transaction' }),
-        successCallback: async () => {
-          await loadSafeMultisigProposals();
+        successCallback: () => {
+          setIsSubmitDisabled(true);
+          loadSafeMultisigProposals();
```
adamgall, line 144: Doesn't this call to `loadSafeMultisigProposals` need to be `async`?

src/hooks/DAO/loaders/useLoadDAOProposals.ts
```diff
@@ -15,20 +15,21 @@ export const useLoadDAOProposals = () => {
...
+      return loadAzoriusProposals(proposal => {
         action.dispatch({
           type: FractalGovernanceAction.SET_AZORIUS_PROPOSAL,
           payload: proposal,
         });
       });
     } else if (type === GovernanceType.MULTISIG) {
       // load mulisig proposals
+      // @dev what is the point of setMethodOnInterval here?
       setMethodOnInterval(loadSafeMultisigProposals);
```
adamgall, line 33: Doesn't this call to `loadSafeMultisigProposals` need to be `async`?

adamgall (changes requested):
@DarksightKellar can you please follow this up with a PR to address the various `@todo`s you created for the names of various types/interfaces/properties?

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
DarksightKellar: `const`, yep

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
DarksightKellar: This was to account for potential errors (after which the button would need to be "re-enabled"). This way only successful executes would actually disable the button.

src/hooks/DAO/loaders/useLoadDAOProposals.ts
DarksightKellar: Eh I didn't think so at first, but I suppose in `setMethodOnInterval`'s first-time ran line `returnValue = await Promise.resolve(getMethod());`, awaiting the call will actually wait for `loadSafeMultisigProposals` to complete?

tomstuart123:
Nice @DarksightKellar / @adamgall .

I tested this with the same childSafe [here ](https://deploy-preview-1990.app.dev.decentdao.org/home?dao=sep:0x773eC5B85aebAa323e700E9bbAA7Cb2eFeAdf56F)where I was originally having issues. The timelock and execute button remained greyed out during the process after click/tx.


Approved

tomstuart123 (approved):

adamgall:
@DarksightKellar i fixed the merge conflict and pushed, fyi

adamgall (approved):

adamgall:
@DarksightKellar does this PR address #1511 also?

Da-Colon:
> @DarksightKellar does this PR address #1511 also?

It helps it for sure but doesn't 'fix' it. With this fix relies on the user not refreshing, right? The button would reappear at that point able to be pushed?

DarksightKellar:
> > @DarksightKellar does this PR address #1511 also?
> 
> It helps it for sure but doesn't 'fix' it. With this fix relies on the user not refreshing, right? The button would reappear at that point able to be pushed?

@Da-Colon I had the same observation. See my comment on this: 
> I expect, if the user refreshes the page in that small window where the node hasn't reflected the updated proposal state, the button will then be clickable -- not much we can do in that case beyond localstorage caching, which we have discussed. Problem is multisig safe proposals are currently not cached as they're loaded all at once every time, and choosing to selectively cache them will take a bit more thought.
> There's a separate task for this caching [here](https://github.com/orgs/decentdao/projects/32/views/3?pane=issue&itemId=65769569), if we end up deciding it's something worth doing.

And @adamgall's take:
> My take is that trying to handle that small timeframe window of a user potentially refreshing the page, using some local storage caching, is way overkill, and not something we need to try to solve for.

which I'm inclined to agree with. It's such a uniquely specific, yet non-breaking flow, that solving for it specifically using caching feels overkill. If we agree that caching multisig safe proposals _is_ worth doing though (I can't argue for it atp), then this will sort itself out as a side effect.

Da-Colon:
> @Da-Colon @DarksightKellar so is #1511 related to the button re-appearing only after immediately refreshing after it was executed the first time?
> 
> If so, I'd say we can close #1511 as a wontfix

Coolio I agree with that

PR #1991, opened 2024-06-12
Generics instead of `any` for `setMethodOnInterval`

adamgall:


Da-Colon (approved):
I dig it, it now shows up as the correct return type of method that is passed in.

DarksightKellar (approved):
Yup, nice

PR #1992, opened 2024-06-12
`1672` update animations

### LINKED ISSUE #1672: Top nav animation transitions
**Page/Section/Component**
Top Nav, all interaction transitions globally

**What is the problem?**
Top Nav lacks animation transition, hover state appears instantly on buttons

**Why is this a problem?**
Without the transition property set, the state changes appear instantly and not as a smooth animation

**Solution**
set default transitions to `ease-out` at `300ms`

Da-Colon:
Fixes #1672

Went a little beyond the original scope>
- All Menus across the app (Sort, Proposal Filter, Favorites)
- Buttons (Covers anything that is `as={Button}` as well as buttons themselves
- ProposalCard (just cleans up to single property)
- NetworkSelector 'Button'
- NavigationLink

It may be hard to noticed but I tested these change by increasing the duration so that I can see the animation happen. 

Only one I was unable to add the transition to (for some reason) is the `WalletMenu`, Neither wrapping with the new `EaseOutComponent` or adding the property directly anywhere, seems to work.

### DIFF
--- a/src/components/ui/utils/Sort.tsx
+++ b/src/components/ui/utils/Sort.tsx
@@ -1,10 +1,11 @@
-import { Flex, Icon, Menu, MenuButton, MenuItem, MenuList, Text } from '@chakra-ui/react';
+import { Button, Flex, Icon, Menu, MenuButton, MenuItem, MenuList, Text } from '@chakra-ui/react';
 import { CaretDown } from '@phosphor-icons/react';
 import { Dispatch, SetStateAction } from 'react';
 import { useTranslation } from 'react-i18next';
 import { NEUTRAL_2_82_TRANSPARENT } from '../../../constants/common';
 import { SortBy } from '../../../types';
 import Divider from './Divider';
+import { EaseOutComponent } from './EaseOutComponent';
 
 function SortMenuItem({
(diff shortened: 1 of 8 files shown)


src/components/ui/utils/Sort.tsx
```diff
@@ -42,18 +43,11 @@ interface ISort {
 export function Sort({ sortBy, setSortBy, buttonProps }: ISort) {
   const { t } = useTranslation();
   return (
-    <Menu direction="ltr">
+    <Menu isLazy>
```
adamgall, line 46: nice

adamgall (approved):

DarksightKellar:
@Da-Colon not sure if relevant to this PR (because I think it's how it's always been, just never pointed out), but I noticed the favourites menu and connect menu have different opening animations. The "slide/grow out" animation is more consistent across other menus in the app. Not sure why the faves behaves differently

https://github.com/decentdao/decent-interface/assets/7101382/219a8dcf-b12e-4b17-957a-a8f5dad29138

xraystyle1980 (approved):
Thanks for adding, looks good. 👍

PR #1996, opened 2024-06-13
`1931` local storage migration

### LINKED ISSUE #1931: Local Storage Migrations
@Da-Colon plz put some content in here

Da-Colon:
Fixes #1931 

End of Series.

This adds the hook `useMigrateLocalStorageV1` Which is placed inside the `Providers` component (at top of tree).

This hook is called when keys that include `fract_` for V1 are found. filters for Favorites and combines the networks into a single favorites cache adding the network prefix for network of the favorite.

It then removes all old `fract_` keys from local storage clearing out the old cache. 

## Thoughts
I thought about adding a `V1_MIGRATION`: true to local storage then use that to make sure this didn't run after running. but felt the search for `fract_` should handle this. Thoughts?


Da-Colon:
@decentdao/engineering Since this is kinda hard to test (I guess you can add a old favorite cache manually in the deploy preview), I added a test and some test cases for it to make sure this works as expected

adamgall (commented):
Two pieces of feedback, which seem in line with your thoughts too:

1. Yes, let's set up the structure now for multiple migrations. I do think adding a "migration version" to local storage is a good idea at this point.
2. Just a bit more abstraction please:
   - `useMigrate` hook
   - the `useMigrate` hook checks for existence of "migration version" value
     - if no value exists
       - run the `v1` migration
       - set "migration version" to `1`
     - if it does exist and is >= `1`, don't run the `v1` migration

Da-Colon:
> Two pieces of feedback, which seem in line with your thoughts too:
> 
> 1. Yes, let's set up the structure now for multiple migrations. I do think adding a "migration version" to local storage is a good idea at this point.
> 2. Just a bit more abstraction please:
>    
>    * rename the hook used in `Providers` to `useMigrate`
>    * the `useMigrate` hook checks for existence of "migration version" value
>      
>      * if no value exists
>        
>        * run the `v1` migration
>        * set "migration version" to `1`
>      * if it does exist and is >= `1`, don't run the `v1` migration

Great ideas, I'll get on it.

DarksightKellar (approved):

src/hooks/utils/cache/useMigrate.ts
```diff
@@ -1,26 +1,29 @@
...
 // This should be a temporary hook to migrate the old local storage to the new one
 // and should be removed after a few months
 
-import { CacheKeys } from './cacheDefaults';
-import { setValue } from './useLocalStorage';
+import { CACHE_VERSIONS, CacheKeys } from './cacheDefaults';
+import { getValue, setValue } from './useLocalStorage';
 
 //@dev for testing seperated the function from the hook and export
 export function migrateCacheToV1(): boolean {
```
adamgall, line 10: Does this need to be `export`ed to be accessible to the test?

src/hooks/utils/cache/useMigrate.ts
```diff
@@ -1,26 +1,29 @@
...
+  const cacheVersion = getValue({ cacheName: CacheKeys.MIGRATION });
+  if (cacheVersion && cacheVersion === CACHE_VERSIONS[CacheKeys.MIGRATION]) {
```
adamgall, line 12: I think this should be `>=` (or `>==`?), not `===`. Think about the second migration

src/hooks/utils/cache/useMigrate.ts
```diff
@@ -1,26 +1,29 @@
...
+    return true;
+  }
```
adamgall, line 14: Also, I think this code here should be _outside_ of the specific `v1 migration` function. Let the layer above this one decide which migration functions to call.

src/hooks/utils/cache/useMigrate.ts
```diff
@@ -1,26 +1,29 @@
...
   // Migrate old cache keys to new format
   const keys = Object.keys(localStorage);
   const fractKeys = keys.filter(key => key.startsWith('fract_'));
   if (!fractKeys.length) {
+    // update migration version if no old cache keys are found
+    setValue({ cacheName: CacheKeys.MIGRATION }, CACHE_VERSIONS[CacheKeys.MIGRATION]);
```
adamgall, line 20: Same idea as previous comment: let the layer above this function decide which migration version to set into local storage.

src/hooks/utils/cache/useMigrate.ts
```diff
@@ -46,10 +49,13 @@ export function migrateCacheToV1(): boolean {
   fractKeys.forEach(key => {
     localStorage.removeItem(key);
   });
+  // set migration version
+
+  setValue({ cacheName: CacheKeys.MIGRATION }, CACHE_VERSIONS[CacheKeys.MIGRATION]);
```
adamgall, line 54: ditto

adamgall (changes requested):

src/hooks/utils/cache/useMigrate.ts
Da-Colon: correct

src/hooks/utils/cache/useMigrate.ts
Da-Colon: I think I need to push back a little on this. 

two part. I never imagined that this migration would stay in the codebase forever. I was thinking after a month or two (or longer) this migration is removed. 

The other part, if we did want to leave it in. I agree I can move this conditional outside, but the `setValue` should be in the V1 function

adamgall: oh snap, yeah, i imagined that these migrations would live here forever.

why should the `setValue` live in this `v1` function rather than the "controller" above it?

src/hooks/utils/cache/useMigrate.ts
```diff
@@ -17,6 +18,7 @@ const runMigrations = () => {
       try {
         migration();
       } catch (e) {
+        logError(e)
```
adamgall, line 21: Let's put `const migration = require(`./migrations/${i}`);` into this try/catch too, and write a test for it (malformed / missing filenames).

test/migrateCacheToV1.test.ts
adamgall, line 1: I don't see it here yet, but I'd like to see separate tests for testing the `runMigrations` function specifically.

adamgall (changes requested):

test/runMigration.test.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    await runMigrations();
+
+    
+    const migrationCache = getValue({ cacheName: CacheKeys.MIGRATION });
+    if (!migrationCache) {
+      throw new Error('Migration cache not found');
+    }
+    expect(migrationCache).toBe(CACHE_VERSIONS[CacheKeys.MIGRATION]);
+  });
```
Da-Colon, line 49: Do you want to see anything else in these tests? I don't believe there is much else to test here unless I change to be able to pass in the version as a param (defaulted to latest) that would allow me to play with the current version. @adamgall

adamgall:
@Da-Colon now that we have a directory full of standard-ly named files, do we even need that "current migration variable"? Can't we just assume it's whatever the largest migration file is?

Da-Colon:
> @Da-Colon now that we have a directory full of standard-ly named files, do we even need that "current migration variable"? Can't we just assume it's whatever the largest migration file is?

hmmm now thats a thought, I'll give it a go

src/hooks/utils/cache/useMigrate.ts
```diff
@@ -1,30 +1,29 @@
...
-        setValue({ cacheName: CacheKeys.MIGRATION }, i - 1);
-        return;
-      }
+  const actualCacheVersion = cacheVersion || 0;
+  const migrationsPath = path.resolve(__dirname, './migrations');
+  const migrationFiles = fs.readdirSync(migrationsPath);
+  const migrationCount = migrationFiles.length;
+  // loop through each pending migration and run in turn
+  for (let i = actualCacheVersion + 1; i <= migrationCount; i++) {
+    const migration = await import(`./migrations/${i}`);
```
adamgall, line 17: Move this line into the `try`, please. Then, I'd like to see a test that had malformed migration filenames and/or gaps in the numbers

adamgall (changes requested):

Da-Colon:
@adamgall Phew. Okay. I realized that I was attempting to use the filesystem (`fs`) which won't be available in the frontend environment. I considered using a `migrations/index` to get the file count but didn't like that extra overhead. After some soul searching I discovered `import.meta.glob` which allows you to dynamically import multiple modules based on glob patterns. This gets us the count we need.

The next problem is that this isn't `mockable` and since its compiled at runtime can't override it. So added a param for migrationCount that is defaulted to the globs key length, this allows me to now only mock the files in the test but pass in the count for the loop. 

tests have been added.

adamgall (changes requested):
Love the tests. Have one question about a small thing, but otherwise looks great

src/hooks/utils/cache/useMigrate.ts
```diff
@@ -0,0 +1,38 @@
...
+) => {
+  const cacheVersion = getValue({ cacheName: CacheKeys.MIGRATION });
+
+  const actualCacheVersion = cacheVersion || 0;
+  // loop through each pending migration and run in turn
+  for (let i = actualCacheVersion + 1; i <= migrationCount; i++) {
+    try {
+      const migration = await import(`./migrations/${i}`);
+      migration.default();
+      setValue({ cacheName: CacheKeys.MIGRATION }, migrationCount);
```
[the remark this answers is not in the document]
Da-Colon: That is correct. left in a change during testing. Updated that. I also went ahead and removed the localStorage update from the loop. I'd rather do that once on a success or failure. 

I also went ahead and added another test for the `runMigration` button for a successful run (only had error states tested).

src/hooks/utils/cache/migrations/1.ts
```diff
@@ -0,0 +1,45 @@
...
+      // Parse favorites value and add network prefix
+      const parsedValue: { v: string[] } = JSON.parse(favoritesValue);
+      parsedValue.v.forEach((value: string) => {
+        newFavorites.push(addNetworkPrefix(value, Number(chainId)));
+      });
+    }
+  }
+  if (newFavorites.length) {
+    // Set new Favorites cache
+    setValue({ cacheName: CacheKeys.FAVORITES }, newFavorites);
```
adamgall, line 38: This puts an expiration of 1 week on the new favorites. OOPSIE

adamgall: We have both:

"user preferences" and "temporary disposable cache data"

being shoehorned into the same constructs, which I've been pointing out.

This is an example of an unexpected bug (consumer being _unaware_ of a very "cache-specific" optional default parameter value to `setValue`) arising due to that shoehorning of two separate things into one system.

I'm just saying this to say it at this point, we don't need to do anything about the content of this message, in this PR.

other than adding `expire: NEVER` to this `setValue` call....and the `MIGRATION setValue` too!

adamgall (changes requested):
Found a bug or two.

---

And then I have yet other idea about the migration code. should it perform some self-tests?

// dev: as i wrote out the below, proper implementation became more unrealistic with each line. keeping for posterity, but these are just the ramblings of a dreamer for now

For example, consider each migration file was split into 4 distinct steps.

1. gather all relevant source data
2. use the source data to CREATE NEW keys only. NEVER modify existing localstorage keys or values in this step.
3. run tests... make sure new data looks like it should look like, considering the source data.
  a. if the tests don't pass then throw an error or something. DON'T DELETE source data or the invalid new data.
  b. if they do pass, update migration

src/hooks/utils/cache/useMigrate.ts
```diff
@@ -0,0 +1,39 @@
...
+      const migration: { default: () => void } = await import(`./migrations/${i}`);
+      migration.default();
+      newVersion = i;
+    } catch (e) {
+      logError(e);
+      newVersion = i - 1;
+    }
+  }
+  setValue({ cacheName: CacheKeys.MIGRATION }, newVersion);
+};
```
adamgall, line 28: So, considering that this will be code that executes every time in its entirety each time the app loads in a browser, I think we should early exit as soon as possible, and not perform that `setValue` at the end.

We just want that last `setValue` line to be executed one time per-browser-per-migration.

src/hooks/utils/cache/migrations/1.ts
```diff
@@ -0,0 +1,45 @@
...
+import { setValue } from '../useLocalStorage';
+
+//@dev for testing seperated the function from the hook and export
+export default function v1MigrateFavorites() {
+  // Migrate old cache keys to new format
+  const keys = Object.keys(localStorage);
+  const fractKeys = keys.filter(key => key.startsWith('fract_'));
+  if (!fractKeys.length) {
+    return;
+  }
```
adamgall, line 15: Can safely remove this early exit if you'd like.

I think the migration one-time-use code looks nicer without that early exit tbh.

test/migrations.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    });
+    localStorage.setItem(oldKey, oldValue);
+
+    const expectedNewValue = ['sep:0xd418E98a11B9189fCc05cddfbB10F4Cee996C749'];
+
+    migrateCacheToV1();
+    const favoriteCache = getValue({ cacheName: CacheKeys.FAVORITES });
+    if (!favoriteCache) {
+      throw new Error('Favorites cache not found');
+    }
```
adamgall, line 27: can delete

test/migrations.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    const expectedNewValue = [
+      'sep:0xd418E98a11B9189fCc05cddfbB10F4Cee996C749',
+      'eth:0xabcdE98a11B9189fCc05cddfbB10F4Cee996C999',
+    ];
+
+    migrateCacheToV1();
+    const favoriteCache = getValue({ cacheName: CacheKeys.FAVORITES });
+    if (!favoriteCache) {
+      throw new Error('Favorites cache not found');
+    }
```
adamgall, line 57: can delete

test/migrations.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    vi.doMock('./migrations/4', () => ({ default: vi.fn() }));
+  });
+
+  it('should stop migration at first gap and log an error', async () => {
+    await runMigrations(3);
+    expect(logging.logError).toHaveBeenCalledOnce();
+    const migrationCache = getValue({ cacheName: CacheKeys.MIGRATION });
+    if (!migrationCache) {
+      throw new Error('Migration cache not found');
+    }
```
adamgall, line 105: can delete

test/migrations.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    vi.doMock('./migrations/foo', () => ({ default: vi.fn() }));
+  });
+
+  it('should stop migration at malformed file name and log an error', async () => {
+    await runMigrations(3);
+    expect(logging.logError).toHaveBeenCalledOnce();
+    const migrationCache = getValue({ cacheName: CacheKeys.MIGRATION });
+    if (!migrationCache) {
+      throw new Error('Migration cache not found');
+    }
```
adamgall, line 131: can delete

test/migrations.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    vi.doMock('./migrations/2', () => ({ default: () => {} }));
+    vi.doMock('./migrations/3', () => ({ default: () => {} }));
+  });
+
+  it('should successfully migrate to the latest version', async () => {
+    await runMigrations(3);
+    const migrationCache = getValue({ cacheName: CacheKeys.MIGRATION });
+    if (!migrationCache) {
+      throw new Error('Migration cache not found');
+    }
```
adamgall, line 154: can delete

adamgall (changes requested):

test/migrations.test.ts
```diff
@@ -0,0 +1,161 @@
...
+    expect(migrationCache).toBe(3);
+  });
```
adamgall, line 156:
```ts
  it('should successfully migrate to the latest version', async () => {
    const migrationCacheBefore = getValue({ cacheName: CacheKeys.MIGRATION });
    expect(migrationCacheBefore).toBe(null);

    await runMigrations(3);

    const migrationCacheAfter = getValue({ cacheName: CacheKeys.MIGRATION });
    expect(migrationCacheAfter).toBe(3);
  });
```

adamgall (changes requested):

adamgall:
@Da-Colon i'm pushing one commit. then i'm good with merging 👍

adamgall (changes requested):
It doesn't seem to work any longer

Screenshot 2024-06-17 at 11 19 54 AM

PR #1998, opened 2024-06-13
Add indicator for pending proposal

### LINKED ISSUE #1983: Indicate pending created proposal in proposal list
In the time between the toast claiming a proposal has been successfully created and when the proposal _actually_ shows up on the UI, the proposal list should indicate that a recently created proposal has indeed been created and will be available soon.

DarksightKellar:
This PR temporarily caches recently created proposal data (transaction hash and metadata). The application can then read this cache and indicate to the user that there are pending proposals waiting to be shown on the UI when ready.

The cached proposal is removed when proposals are loaded and any of them has a transaction hash that matches its transaction hash.

Pending toast:
Screenshot 2024-06-13 at 18 01 24

Toast dismissed when proposal shows up:
Screenshot 2024-06-13 at 18 01 43

Note: the toast is a temporary workaround while waiting on @decentdao/design to finalise the actual look for pending proposal(s) UI.

closes #1983

### DIFF
--- a/src/components/Proposals/ProposalsList.tsx
+++ b/src/components/Proposals/ProposalsList.tsx
@@ -18,8 +18,7 @@ export function ProposalsList({ proposals }: { proposals: FractalProposal[] }) {
   } = useFractal();
   const { canUserCreateProposal } = useCanUserCreateProposal();
   const { addressPrefix } = useNetworkConfig();
-
-  const { t } = useTranslation('proposal');
+  const { t } = useTranslation(['proposal', 'common']);
   return (
     <Flex
       flexDirection="column"
@@ -44,12 +43,7 @@ export function ProposalsList({ proposals }: { proposals: FractalProposal[] }) {
         <EmptyBox emptyText={t('emptyProposals')}>
           {canUserCreateProposal && daoAddress && (
             <Link to={DAO_ROUTES.proposalNew.relative(addressPrefix, daoAddress)}>
-              <Button
-                variant="text"
-                textStyle="text-xl-mono-bold"
-              >
-                {t('createProposal')}
-              </Button>
+              <Button variant="text">{t('createProposal')}</Button>
             </Link>
           )}
         </EmptyBox>
--- a/src/components/Proposals/ProposalsList.tsx
+++ b/src/components/Proposals/ProposalsList.tsx
@@ -18,7 +18,7 @@ export function ProposalsList({ proposals }: { proposals: FractalProposal[] }) {
   } = useFractal();
   const { canUserCreateProposal } = useCanUserCreateProposal();
   const { addressPrefix } = useNetworkConfig();
-  const { t } = useTranslation(['proposal', 'common']);
+  const { t } = useTranslation(['proposal']);
   return (
     <Flex
       flexDirection="column"
--- a/src/components/pages/DaoDashboard/Activities/index.tsx
+++ b/src/components/pages/DaoDashboard/Activities/index.tsx
@@ -1,7 +1,6 @@
 import { Box, Flex } from '@chakra-ui/react';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import useProposals from '../../../../hooks/DAO/proposal/useProposals';
 import { useFractal } from '../../../../providers/App/AppProvider';
 import { ActivityEventType, SortBy, TreasuryActivity, FractalProposal } from '../../../../types';
 import { ActivityTreasury } from '../../../Activity/ActivityTreasury';
@@ -21,7 +20,6 @@ export function Activities() {
   const [sortBy, setSortBy] = useState<SortBy>(SortBy.Newest);
 
   const { sortedActivities } = useActivities(sortBy);
-  useProposals({ sortBy, filters: [] });
 
   const { t } = useTranslation('dashboard');
 
--- a/src/components/pages/DaoDashboard/Activities/index.tsx
+++ b/src/components/pages/DaoDashboard/Activities/index.tsx
@@ -1,6 +1,7 @@
 import { Box, Flex } from '@chakra-ui/react';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
+import useProposals from '../../../../hooks/DAO/proposal/useProposals';
 import { useFractal } from '../../../../providers/App/AppProvider';
 import { ActivityEventType, SortBy, TreasuryActivity, FractalProposal } from '../../../../types';
 import { ActivityTreasury } from '../../../Activity/ActivityTreasury';
@@ -18,7 +19,9 @@ export function Activities() {
     governance: { type, loadingProposals, allProposalsLoaded },
   } = useFractal();
   const [sortBy, setSortBy] = useState<SortBy>(SortBy.Newest);
+
   const { sortedActivities } = useActivities(sortBy);
+  const { tempProposals } = useProposals({ sortBy, filters: [] });
 
   const { t } = useTranslation('dashboard');
 
@@ -49,6 +52,9 @@ export function Activities() {
             flexDirection="column"
             gap="1rem"
           >
+            {!!tempProposals.length && (
+              <Box mb="1rem">Waiting for confirmation on {tempProposals.length} proposals</Box>
+            )}
             {sortedActivities.map((activity, i) => {
               if (activity.eventType === ActivityEventType.Governance) {
                 return (
--- a/src/components/pages/DaoDashboard/Activities/index.tsx
+++ b/src/components/pages/DaoDashboard/Activities/index.tsx
@@ -1,7 +1,6 @@
 import { Box, Flex } from '@chakra-ui/react';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import useProposals from '../../../../hooks/DAO/proposal/useProposals';
 import { useFractal } from '../../../../providers/App/AppProvider';
 import { ActivityEventType, SortBy, TreasuryActivity, FractalProposal } from '../../../../types';
 import { ActivityTreasury } from '../../../Activity/ActivityTreasury';
@@ -21,7 +20,6 @@ export function Activities() {
   const [sortBy, setSortBy] = useState<SortBy>(SortBy.Newest);
 
   const { sortedActivities } = useActivities(sortBy);
-  const { tempProposals } = useProposals({ sortBy, filters: [] });
 
   const { t } = useTranslation('dashboard');
 
@@ -52,9 +50,6 @@ export function Activities() {
             flexDirection="column"
             gap="1rem"
           >
-            {!!tempProposals.length && (
-              <Box mb="1rem">Waiting for confirmation on {tempProposals.length} proposals</Box>
-            )}
             {sortedActivities.map((activity, i) => {
               if (activity.eventType === ActivityEventType.Governance) {
                 return (
--- a/src/components/pages/DaoDashboard/Activities/index.tsx
+++ b/src/components/pages/DaoDashboard/Activities/index.tsx
@@ -18,7 +18,6 @@ export function Activities() {
     governance: { type, loadingProposals, allProposalsLoaded },
   } = useFractal();
   const [sortBy, setSortBy] = useState<SortBy>(SortBy.Newest);
-
   const { sortedActivities } = useActivities(sortBy);
 
   const { t } = useTranslation('dashboard');
--- a/src/components/pages/DaoDashboard/Activities/index.tsx
+++ b/src/components/pages/DaoDashboard/Activities/index.tsx
@@ -1,6 +1,7 @@
 import { Box, Flex } from '@chakra-ui/react';
 import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
+import useProposals from '../../../../hooks/DAO/proposal/useProposals';
 import { useFractal } from '../../../../providers/App/AppProvider';
 import { ActivityEventType, SortBy, TreasuryActivity, FractalProposal } from '../../../../types';
 import { ActivityTreasury } from '../../../Activity/ActivityTreasury';
@@ -20,6 +21,7 @@ export function Activities() {
   const [sortBy, setSortBy] = useState<SortBy>(SortBy.Newest);
 
   const { sortedActivities } = useActivities(sortBy);
+  useProposals({ sortBy, filters: [] });
 
   const { t } = useTranslation('dashboard');
 
--- a/src/hooks/DAO/loaders/useLoadDAOProposals.ts
+++ b/src/hooks/DAO/loaders/useLoadDAOProposals.ts
@@ -1,10 +1,9 @@
 import { useCallback } from 'react';
 import { useFractal } from '../../../providers/App/AppProvider';
 import { FractalGovernanceAction } from '../../../providers/App/governance/action';
-import { useNetworkConfig } from '../../../providers/NetworkConfig/NetworkConfigProvider';
 import { GovernanceType } from '../../../types';
 import { CacheKeys } from '../../utils/cache/cacheDefaults';
-import { getValue } from '../../utils/cache/useLocalStorage';
+import { useLocalStorage } from '../../utils/cache/useLocalStorage';
 import { useUpdateTimer } from '../../utils/useUpdateTimer';
 import { TempProposalData } from '../proposal/useSubmitProposal';
 import { useAzoriusProposals } from './governance/useAzoriusProposals';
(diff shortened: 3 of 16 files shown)


nicolaus-sherrill:
@DarksightKellar are you thinking that this would be a new card added to the top of the list that communicates "this is temporary until your actual proposal is visible"?

DarksightKellar:
> @DarksightKellar are you thinking that this would be a new card added to the top of the list that communicates "this is temporary until your actual proposal is visible"?

@nicolaus-sherrill Yep, I had 2 options in mind:

- a single new card that communicates that one or more proposals are pending

- a new _list_ that shows basic data for each pending proposal. These won't be clickable, of course

This PR's set up for both, so that won't be an issue, just need to decide which is better UX

src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -188,9 +207,17 @@ export default function useSubmitProposal() {
...
+
+        const responseData = JSON.parse(response.config.data);
+        const txHash = responseData.contractTransactionHash;
         await loadDAOProposals();
         if (successCallback) {
+          cacheTemporaryProposal({
+            safeAddress,
+            txHash,
+            proposalMetadata: proposalData.metaData,
+          });
```
Da-Colon, line 219: I don't think this should be inside this conditional? whether a `successCallback` func is passed in doesn't matter to the success of the creation of the proposal, right?

src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -260,8 +288,15 @@ export default function useSubmitProposal() {
...
-          successCallback(addressPrefix, safeAddress!);
+
+        if (successCallback && !!safeAddress) {
+          cacheTemporaryProposal({
+            safeAddress,
+            txHash: receipt.transactionHash,
+            proposalMetadata: proposalData.metaData,
+          });
+
+          successCallback(addressPrefix, safeAddress);
```
Da-Colon, line 299: same question about being inside the successCallback conditional.

Da-Colon (changes requested):
a few minor comments

src/hooks/DAO/proposal/useSubmitProposal.ts
DarksightKellar: Yeahhh, you're right

src/components/Proposals/ProposalsList.tsx
```diff
@@ -18,8 +18,7 @@ export function ProposalsList({ proposals }: { proposals: FractalProposal[] }) {
   } = useFractal();
   const { canUserCreateProposal } = useCanUserCreateProposal();
   const { addressPrefix } = useNetworkConfig();
-
-  const { t } = useTranslation('proposal');
+  const { t } = useTranslation(['proposal', 'common']);
```
adamgall, line 21: Is the addition of `common` here needed?

src/hooks/DAO/loaders/useLoadDAOProposals.ts
```diff
@@ -43,3 +47,12 @@ export const useLoadDAOProposals = () => {
 
   return loadDAOProposals;
 };
+
+export const useLoadTempProposals = () => {
+  const { chain } = useNetworkConfig();
+  const loadTempDAOProposals = useCallback(async () => {
+    return (getValue(CacheKeys.TEMP_PROPOSALS, chain.id) || []) as TempProposalData[];
```
adamgall, line 54: @Da-Colon are your changes to local storage going to add type safety to `getValue` calls like this, so we don't need to do unsafe `as` casting?

src/hooks/DAO/proposal/useProposals.ts
```diff
@@ -13,6 +18,25 @@ export default function useProposals({
     governance: { proposals },
   } = useFractal();
 
+  const loadTempDAOProposals = useLoadTempProposals();
+  const [tempProposals, setTempProposals] = useState<TempProposalData[]>([]);
+  const [pendingProposalsToastId, setPendingProposalsToastId] = useState<Id>();
+
+  useEffect(() => {
+    if (tempProposals.length) {
+      const toastId = toast.info(`Waiting for confirmation on ${tempProposals.length} proposals`, {
```
adamgall, line 27: This string should go into a translation file please

src/hooks/DAO/proposal/useProposals.ts
```diff
@@ -22,9 +46,31 @@ export default function useProposals({
...
+    },
+    [getValue, pendingProposalsToastId, setValue, tempProposals.length],
+  );
+
   const sortedAndFilteredProposals = useMemo(() => {
-    return [...(proposals || [])]
-      .filter(proposal => filters.includes(proposal.state!))
+    return (proposals || [])
+      .filter(proposal => {
+        removeTemporaryProposal(proposal.transactionHash!);
```
adamgall, line 71: Hmm, at first glance it feels weird to have this state-changing code within a `filter`....

src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -91,6 +99,16 @@ export default function useSubmitProposal() {
   const { chain, safeBaseURL, addressPrefix } = useNetworkConfig();
   const ipfsClient = useIPFSClient();
 
+  const cacheTemporaryProposal = useCallback(
+    (tempProposal: TempProposalData) => {
+      const temporaryProposals = (getValue(CacheKeys.TEMP_PROPOSALS, chain.id) ||
+        []) as TempProposalData[];
```
adamgall, line 105: another callout, that it would nice to have proper generics typing on `getValue`

src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -91,6 +99,16 @@ export default function useSubmitProposal() {
...
+      const updatedProposals = [...temporaryProposals, tempProposal];
+      setValue(CacheKeys.TEMP_PROPOSALS, updatedProposals, CacheExpiry.ONE_DAY);
```
adamgall, line 107: Hmm, but actually, how was the decision made to store these "temporary pending" proposals in local storage?

I understand that these will persist through a refresh, but do we really need that?

imo seems a bit overkill to utilize localstorage for this functionality.

src/hooks/utils/useSafeTransactions.ts
```diff
@@ -307,7 +307,9 @@ export const useSafeTransactions = () => {
                 : undefined,
             proposalId: eventSafeTxHash,
             targets,
-            transactionHash: multiSigTransaction.transactionHash,
+            transactionHash:
+              multiSigTransaction.transactionHash ||
+              (transaction as SafeMultisigTransactionWithTransfersResponse).safeTxHash,
```
adamgall, line 312: Is this needed? The code builds for me without this change.

Logically, I suppose the reason why it would needed is if `transactionHash` is the empty string (Typescript says it's type is simply `string`).

empty string is `falsey`, so `safeTxHash` will be used in the `||` conditional.

Smells to me like maybe this is an opportunity to tighten up some typing.

Do you follow where my head's at? and if so do you agree? If so, Please add a `TODO` comment here for someone searching the codebase later for things to do.

src/utils/azorius.ts
```diff
@@ -218,6 +222,8 @@ export const mapProposalCreatedEventToProposal = async (
       event => BigInt(event.args[0]) === proposalId,
     );
     transactionHash = executedEvent?.transactionHash;
+  } else {
+    transactionHash = createdEvent.transactionHash;
```
adamgall, line 226: Can you please add some more explicit code here?

```ts
// src/types/fractal.ts

export interface ActivityBase {
  eventDate: Date;
  eventType: ActivityEventType;
  transaction?: ActivityTransactionType;
  transactionHash: string;
}
```

```ts
  // src/utils/azorius.ts

  let transactionHash: string;
  if (state === FractalProposalState.EXECUTED) {
    const executedEvent = (await executedEvents)?.find(
      event => BigInt(event.args[0]) === proposalId,
    );
    if (!executedEvent) {
      throw new Error('proposal state is executed, but no event found');
    }
    transactionHash = executedEvent.transactionHash;
  } else {
    transactionHash = createdEvent.transactionHash;
  }
```

src/hooks/DAO/proposal/useSubmitProposal.ts
adamgall: @DarksightKellar this is probably my biggest-picture question about this PR

tomstuart123:
@DarksightKellar - I maybe missing something but I couldn't replicate this and get the toaster to show up.

I tried 4 proposals here - https://deploy-preview-1998.app.dev.decentdao.org/proposals?dao=sep:0x40C6E5eFCE3b1f101bFB59e876F14aa5455D3E19

Video here - https://decentdao.slack.com/archives/C06QZTUTKRC/p1718379588738149?thread_ts=1718379569.781799&cid=C06QZTUTKRC

src/hooks/DAO/loaders/useLoadDAOProposals.ts
```diff
@@ -43,3 +46,13 @@ export const useLoadDAOProposals = () => {
 
   return loadDAOProposals;
 };
+
+export const useLoadTempProposals = () => {
+  const { getValue } = useLocalStorage();
```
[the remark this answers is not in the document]
DarksightKellar: @Da-Colon I've actually reverted the update i did wrt to this comment. It broke the cache -- used a weird key that couldn't be read back. Instead of spending more time fixing, I'd rather wait for your merge to break this(these) line(s), and update accordingly, when I know keys are uniform app-wide.

src/components/Proposals/ProposalsList.tsx
DarksightKellar: Nope, updated

src/hooks/DAO/proposal/useProposals.ts
DarksightKellar: Done-zo

src/hooks/DAO/proposal/useProposals.ts
DarksightKellar: Yeahhh I was trying to avoid having run yet _another_ loop over the proposals just to remove potentially 1 (if ever) pending proposal. Besides `filter` is just a fancy way of doing `map` plus manually adding to a pre-declared array based on the `predicate`. I'm totally fine with it in this case.

src/hooks/utils/useSafeTransactions.ts
DarksightKellar: It _is_ needed, yep. The typing in that case is misleading, and `multiSigTransaction.transactionHash` turns out to always be an empty string (or was it `undefined`?). I chose not to replace the line with `transactionHash: (transaction as SafeMultisigTransactionWithTransfersResponse).safeTxHash` in case there's some legacy reasoning behind the choice in the first place.

I'll add the `todo` for the team to revisit this.

src/utils/azorius.ts
DarksightKellar: Oh right, I see

DarksightKellar:
> @DarksightKellar - I maybe missing something but I couldn't replicate this and get the toaster to show up.
> 
> I tried 4 proposals here - https://deploy-preview-1998.app.dev.decentdao.org/proposals?dao=sep:0x40C6E5eFCE3b1f101bFB59e876F14aa5455D3E19
> 
> Video here - https://decentdao.slack.com/archives/C06QZTUTKRC/p1718379588738149?thread_ts=1718379569.781799&cid=C06QZTUTKRC

@tomstuart123 Yikes, one of the newer cleanup commits broke the cache. I've reverted -- should be working now

src/hooks/DAO/proposal/useSubmitProposal.ts
DarksightKellar: Hmm I suppose this could all be implemented with "simpler" app state. But I've tested (and tasted) the ability to refresh and still seeing that persistent pending toast. I like it and don't wanna go back.

Besides, won't the logic largely be the same? The only real difference feels like it'd be _where_ the temp data is coming from (+ the tiny extra ms needed to read/write to storage).

src/hooks/DAO/proposal/useSubmitProposal.ts
DarksightKellar: I expect these will need update after David's work is merged in, and I think we'll get type safety, yeah

src/hooks/DAO/loaders/useLoadDAOProposals.ts
Da-Colon: Yes thats correct, I've updated so that each cache's getValue is inferred from the cacheName key

src/hooks/DAO/proposal/useSubmitProposal.ts
Da-Colon: We could go crazy and make it slightly more 'complex' and use netlify/function and the store? though probably not worth the lift to show all viewers of pending proposals.

src/utils/azorius.ts
```diff
@@ -217,7 +221,14 @@ export const mapProposalCreatedEventToProposal = async (
...
-    transactionHash = executedEvent?.transactionHash;
+
+    if (!executedEvent) {
+      throw new Error('Proposal state is EXECUTED, but no event found');
+    }
+
+    transactionHash = executedEvent.transactionHash;
+  } else {
+    transactionHash = createdEvent.transactionHash;
   }
```
adamgall, line 232: And then I guess I do have an outstanding question at the moment... what is going on here with finding a new use for this `transactionHash` property with a brand new `createdEvent` object?

src/components/pages/DaoDashboard/Activities/index.tsx
```diff
@@ -18,7 +19,9 @@ export function Activities() {
     governance: { type, loadingProposals, allProposalsLoaded },
   } = useFractal();
   const [sortBy, setSortBy] = useState<SortBy>(SortBy.Newest);
+
   const { sortedActivities } = useActivities(sortBy);
+  useProposals({ sortBy, filters: [] });
```
adamgall, line 24: What does calling this here do?

src/hooks/DAO/proposal/useSubmitProposal.ts
adamgall: It wouldn't make sense to show anyone other than the creator of the proposal, anything about that proposal before their RPC node picked up the block and added it to the official in-memory list.

adamgall: I do still think localstorage is a bit excessive for this implementation, but that's on me for not being clearer on upfront on what I was expecting.

I'll need to go look at the code again that manages adding/removing from this "pending transaction" datastore in localstorage...

src/hooks/DAO/proposal/useProposals.ts
```diff
@@ -13,6 +19,42 @@ export default function useProposals({
...
+
+  const dismissPendingProposalsToast = useCallback(() => {
+    if (pendingProposalsToastIdRef.current) {
+      toast.dismiss(pendingProposalsToastIdRef.current);
+      pendingProposalsToastIdRef.current = undefined;
+    }
+  }, [pendingProposalsToastIdRef]);
+
+  useEffect(() => {
+    if (pendingProposalsToastIdRef.current) return;
```
adamgall, line 36: Why this funny ref code? Is using the `toastId` and dismissing it in the effect's return not enough?

src/hooks/DAO/proposal/useProposals.ts
adamgall: I just don't think this code belongs here. We're way to far up at the presentation layer to be fucking around with _cleaning up our localstorage state_.

Could this go into the the reducer actions for when saving actual proposals into state?

adamgall (changes requested):
Looks good, just a few comments and questions

tomstuart123:
Awesome, this works so smoothly for 'Create Proposal'. Thanks @DarksightKellar 

Last question, pre approval --> which other types of proposals should this work with? I tried it with SendAssets and it worked but not when I tried a change Safename proposal

adamgall:
> Awesome, this works so smoothly for 'Create Proposal'. Thanks @DarksightKellar
> 
> Last question, pre approval --> which other types of proposals should this work with? I tried it with SendAssets and it worked but not when I tried a change Safename proposal



> Awesome, this works so smoothly for 'Create Proposal'. Thanks @DarksightKellar
> 
> Last question, pre approval --> which other types of proposals should this work with? I tried it with SendAssets and it worked but not when I tried a change Safename proposal

This is a great question and makes me question whether `useSubmitProposal` is the appropriate place to trigger adding a temp proposal record... 

where else in the codebase can it go to capture _all_ proposal creation, or does that place not exist (yet)?

src/hooks/DAO/proposal/useProposals.ts
adamgall: This is fixed

src/components/pages/DaoDashboard/Activities/index.tsx
adamgall: Removed

adamgall (dismissed):

adamgall:
ok @decentdao/engineering and @decentdao/product, this is ready for a full code re-review and UX re-test

adamgall (approved):

tomstuart123:
Nice, this works for all of the below:

- Create child safe
- Create proposal
- send assets
- change name
- change snapshot space

Approved. Great work

tomstuart123 (approved):

Da-Colon (approved):

PR #2002, opened 2024-06-13
Remove typechain cleanup 1

adamgall:
Depends on https://github.com/decentdao/decent-interface/pull/2001

- Remove `FractalContracts` interface and `baseContracts` property from `FractalStore`
- Remove `ContractConnection` type
- Remove `BaseContracts` interface
- Remove `ContractEvent` interface
- Remove `useSafeContracts` hook


src/hooks/DAO/loaders/governance/useERC721LinearStrategy.ts
```diff
@@ -34,13 +32,8 @@ export const useERC721LinearStrategy = () => {
...
   const loadERC721Strategy = useCallback(async () => {
-    if (
-      !azoriusContractAddress ||
-      !provider ||
-      !baseContracts ||
-      !erc721LinearVotingContract ||
-      !publicClient
-    ) {
+    if (!azoriusContractAddress || !provider || !erc721LinearVotingContract || !publicClient) {
+      // TODO i don't like this
```
Da-Colon, line 36: Don't like what? Also why is it returning a empty object? The doesn't return anything otherwise.

Da-Colon (approved):

src/hooks/DAO/loaders/governance/useERC721LinearStrategy.ts
adamgall: That's what I don't like, that we're just returning an empty object here!

You're also right in that this return value doesn't matter. I will update to just return undefined.

PR #2007, opened 2024-06-14
Remove typechain cleanup 6

adamgall:
Depends on https://github.com/decentdao/decent-interface/pull/2006

- Remove _almost_ all uses of `useSignerOrProvider`, `useSigner`, and `useProvider`


adamgall:
@mudrila rather than fix here, and rebase all the rest of these branches, i'm not going to address your comments here because i do take care of them in a future PR.

Da-Colon (approved):
One Comment, but otherwise looks good to me.

PR #2025, opened 2024-06-18
move styling to box surrounding list

Da-Colon:
- update with proper overflow

image


tomstuart123 (approved):
good catch. I hate having loads of safes saved so I never hit this bug.

This works well and Approved

adamgall (approved):

PR #2039, opened 2024-06-22
Remove typechain cleanup 17

adamgall:
This PR represents the first time I'm trying to actually load the app in a browser since this whole project started.

### DIFF
--- a/src/hooks/DAO/loaders/governance/useAzoriusProposals.ts
+++ b/src/hooks/DAO/loaders/governance/useAzoriusProposals.ts
@@ -91,7 +91,7 @@ export const useAzoriusProposals = () => {
       return;
     }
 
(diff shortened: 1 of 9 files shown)


adamgall:
@decentdao/engineering i could use some help with this code...

- first thing I'm struggling with is getting the test to pass
- second is trying to figure out why i'm getting a few `ContractFunctionExecutionError`s in the console, for specific lines in the codebase but idk it doesn't make sense to me.

Da-Colon (commented):
So found a fix for one of them...Not really sure what made me try it, as the stackoverflow answers don't really point to this. But one of the `ContractFunctionExecutionErrors` points to this hook `useCanUserCreateProposal`

The error is removed when you replace `publicClient` with `walletClient`? Not sure why this removes the error. for this particular spot

PR #2044, opened 2024-06-22
Fix crosschain favorites names

### LINKED ISSUE #2043: Cross-chain My Safes names don't load
When My Safes from multiple networks are saved, only the Safes names from the network that we're currently connected to will load.

adamgall:
Fixes #2043 

Now, all Favorited Safe names will load regardless of the currently connected network.

### DIFF
--- a/src/hooks/DAO/useGetDAOName.ts
+++ b/src/hooks/DAO/useGetDAOName.ts
@@ -1,11 +1,10 @@
-import { FractalRegistry } from '@fractal-framework/fractal-contracts';
 import { useCallback, useEffect, useState } from 'react';
-import { Address, PublicClient } from 'viem';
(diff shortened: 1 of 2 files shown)


adamgall:
@Da-Colon @mudrila can you help look into why the build is failing? it's broken for me locally, too, but I don't see why.

Da-Colon:
@adamgall 
image

`useGetDAOname` or in this case `useGetDAONameDeferred` returns any (inferred) it looks like which is why we don't see the error in IDE.

adamgall:
duh, thank you @Da-Colon

tomstuart123 (approved):
two things 
1) approved - the names seem to work well

2) This is the first time I've tested this cross network favourites. Huge props to @adamgall / @Da-Colon / @mudrila . The moment I clicked a safe on another chain and it instantly took me to the right chain, I smiled from ear to ear :)

PR #2060, opened 2024-07-01
Tree and Hats type sanitization

adamgall:
When loading in a Hat Tree, sanitize the Hats Tree by making sure it fits the shape we care about perfectly

### DIFF
--- a/src/state/useRolesState.ts
+++ b/src/state/useRolesState.ts
@@ -1,18 +1,150 @@
-import { Tree } from '@hatsprotocol/sdk-v1-subgraph';
+import { Tree, Hat } from '@hatsprotocol/sdk-v1-subgraph';
+import { Hex } from 'viem';
 import { create } from 'zustand';
+
+export class DecentHatsError extends Error {
+  constructor(message: string) {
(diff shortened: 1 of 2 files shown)


src/state/useRolesState.ts
```diff
@@ -1,26 +1,158 @@
...
+
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(this, DecentHatsError);
+    }
+  }
+}
+
+interface DecentHat {
+  id: Hex;
+  prettyId?: string;
```
adamgall, line 18: This is a string of the form "0x0000.aaaa.bbbb.cccc.etc"

Da-Colon (approved):
Coolio Looks good to me. I'll take it and adjust my components to work with this new Typing. I like the additional mocking assistance

src/state/useRolesState.ts
```diff
@@ -1,26 +1,158 @@
...
+  name?: string;
+  description?: string;
```
adamgall, line 20: This PR doesn't include any of our necessary IPFS integration.

PR #2066, opened 2024-07-02
Add more logic to Roles store

adamgall:


Da-Colon (approved):
Looks like exactly what we needed.

PR #2078, opened 2024-07-04
Hot-fix navigating to subDAO from hierarchy page when parent DAO has large amount of subDAOs

mudrila:
- Prevent setting `readOnlyValues` if they haven't changed
- Prevent setting `canUserSubmitProposal` if that value hasn't changed

For testing - try to visit `0x374D5a17D2433415C48C735A493F0523DaB99041` DAO's hierarchy page on Base

The actual problem is connected to rate limits and requests failure, which causes infinitely reiterate callbacks in `useEffect`


tomstuart123 (approved):

tomstuart123:
Approved as clicked through to multiple subDAOs from both Myosin and Shutter's parent Safes

adamgall (approved):

PR #2103, opened 2024-07-08
Fix "Top Hat ID is invalid" error that pops up right after new role creation is executed

### LINKED ISSUE #2083: `[roles-0.1.0]` Bug | Hat Id not valid
After succesful execution of the proposal to create a new hat tree. The Toast pops up that the Hat ID is not valid and when visiting Roles page it is stuck in loading (hatTree === undefined). Fixes after refresh

adamgall:
Closes #2083

After hearing live event for new Hat Tree Id, pause 10 seconds before adding that to state

### DIFF
--- a/src/hooks/DAO/loaders/useHatsTree.ts
+++ b/src/hooks/DAO/loaders/useHatsTree.ts
@@ -20,6 +20,7 @@ const useHatsTree = () => {
       if (hatsTreeId === undefined || hatsTreeId === null) {
         return;
       }
+
       try {
         const tree = await hatsSubgraphClient.getTree({
           chainId: chain.id,
@@ -79,7 +80,7 @@ const useHatsTree = () => {
           }
         }
       } catch (e) {
-        setHatsTree(undefined);
+        setHatsTree(null);
         const message = 'Hats Tree ID is not valid';
         toast(message);
         console.error(e, {
--- a/src/hooks/DAO/useKeyValuePairs.ts
+++ b/src/hooks/DAO/useKeyValuePairs.ts
@@ -17,37 +17,37 @@ const getHatsTreeId = (
   }
 
   // get most recent event where `hatsTreeId` was set
-  const hatsTreeIdEvent = events
+  const topHatIdEvent = events
     .filter(event => event.args.key && event.args.key === 'topHatId')
     .pop();
 
-  if (!hatsTreeIdEvent) {
+  if (!topHatIdEvent) {
     return null;
   }
 
-  if (!hatsTreeIdEvent.args.value) {
+  if (!topHatIdEvent.args.value) {
     logError({
-      message: "KVPairs 'hatsTreeIdEvent' without a value",
+      message: "KVPairs 'topHatIdEvent' without a value",
       network: chainId,
       args: {
-        transactionHash: hatsTreeIdEvent.transactionHash,
-        logIndex: hatsTreeIdEvent.logIndex,
+        transactionHash: topHatIdEvent.transactionHash,
+        logIndex: topHatIdEvent.logIndex,
       },
     });
     return undefined;
   }
 
   try {
-    const topHatId = BigInt(hatsTreeIdEvent.args.value);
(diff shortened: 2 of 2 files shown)


src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -79,7 +80,7 @@ const useHatsTree = () => {
           }
         }
       } catch (e) {
-        setHatsTree(undefined);
+        setHatsTree(null);
```
adamgall, line 83: This is a sub-fix. The real fix is to not try to fetch from subgraph immediately after receiving event. Give Subgraph a few seconds to index the new Hat Tree.

But, even if we do get to it too early, let's set the hat Tree to `null` not `undefined`, so the loading screen goes away and it's as if the Safe has no Roles (which is technically accurate here because the given Tree ID doesn't line up with a valid Tree in the shape that we want (or rather, we can't get it from Subgraph).

src/hooks/DAO/useKeyValuePairs.ts
```diff
@@ -17,37 +17,37 @@ const getHatsTreeId = (
...
     logError({
-      message: "KVPairs 'hatsTreeIdEvent' value not a number",
+      message: "KVPairs 'topHatIdEvent' value not a number",
       network: chainId,
       args: {
-        transactionHash: hatsTreeIdEvent.transactionHash,
-        logIndex: hatsTreeIdEvent.logIndex,
+        transactionHash: topHatIdEvent.transactionHash,
+        logIndex: topHatIdEvent.logIndex,
       },
```
adamgall, line 51: Variable rename

src/hooks/DAO/useKeyValuePairs.ts
```diff
@@ -85,7 +85,14 @@ const useKeyValuePairs = () => {
...
       {
-        onLogs: logs => setHatsTreeId(getHatsTreeId(logs, chain.id)),
+        onLogs: logs => {
+          // dev: when this event is captured in realtime, give the subgraph
+          // time to index, and do that most cleanly by not even telling the rest
+          // of our code that we have the hats tree id until some time has passed.
+          setTimeout(() => {
+            setHatsTreeId(getHatsTreeId(logs, chain.id));
+          }, 10_000);
+        },
```
adamgall, line 95: Here's the good stuff. If we're getting the new Hat Tree ID from a live event, then pause 10 seconds before setting the hats tree id in state. Setting hats tree id in state is what triggers fetching the Tree from Subgraph.

src/hooks/DAO/useKeyValuePairs.ts
```diff
@@ -85,7 +85,14 @@ const useKeyValuePairs = () => {
         theAddress: node.daoAddress,
       },
       {
-        onLogs: logs => setHatsTreeId(getHatsTreeId(logs, chain.id)),
+        onLogs: logs => {
+          // dev: when this event is captured in realtime, give the subgraph
+          // time to index, and do that most cleanly by not even telling the rest
+          // of our code that we have the hats tree id until some time has passed.
+          setTimeout(() => {
+            setHatsTreeId(getHatsTreeId(logs, chain.id));
```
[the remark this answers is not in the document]
adamgall: Yeah this timeout is just for when capturing events from the live event listener. It doesn't affect loading data from past events like will be the majority of loads

src/hooks/DAO/useKeyValuePairs.ts
```diff
@@ -17,37 +17,37 @@ const getHatsTreeId = (
   }
 
   // get most recent event where `hatsTreeId` was set
-  const hatsTreeIdEvent = events
+  const topHatIdEvent = events
```
DarksightKellar, line 20: Comment above this line still reads "hatsTreeId"

src/hooks/DAO/useKeyValuePairs.ts
DarksightKellar: Nice

DarksightKellar (approved):

PR #2105, opened 2024-07-09
Replace bar loader with wave loader animation

DarksightKellar:


DarksightKellar:
> Thanks so much @DarksightKellar .
> 
> 1. one thing , on the hierarchy the loading indicator moves a bit as its loading (from lower to higher). Not urgent though but can be seen here if you reload - https://deploy-preview-2105.app.dev.decentdao.org/hierarchy?dao=eth:0xB98d45F9021D71E6Fc30b43FD37FB3b1Bf12c064
> 2. design wise, @nicolaus-sherrill  when online can you have a look at this. My take is the indicator should be a little smaller but want your take before I send kelvin down a wrong path

Hey @tomstuart123, on

1. This appears to be existing behaviour (see [same screen here on a different branch](https://deploy-preview-2106.app.dev.decentdao.org/hierarchy?dao=eth:0xB98d45F9021D71E6Fc30b43FD37FB3b1Bf12c064). Looks like different loading items that are showing, and not that the _same_ indicator jumps to the top. Regardless, this is all to be replaced by the loading indicator overhaul in the future.

2. Will wait for @nicolaus-sherrill's take here

adamgall:
I agree with @DarksightKellar on point 1, this is supposed to be a quick swap, which doesn't include custom cleanup after the fact.

On point 2, @DarksightKellar can you "make it a bit smaller" and we see what that looks like? If that would take more than like 2 minute then don't worry about it imo.

DarksightKellar:
> d we see what that looks like? If that would take more than like 2 minute then don't worry about i

It would lol. The implementation is way too sensitive to the individual boxes that make up the wave, so a size change would involve a bit of manual trial and error @adamgall

DarksightKellar:
@tomstuart123 on @adamgall's suggestion, I went ahead and experimented with a slightly smaller loader size. Take a look when you get the chance. 

@nicolaus-sherrill still need your take here when you're online. I can't use exact size numbers, but I can try to get it as close as possible to a suggested size

adamgall:
For the sake of moving things along, i'd like to merge this as-is

adamgall (approved):

nicolaus-sherrill:
Aligned, thanks @adamgall and @DarksightKellar

PR #2109, opened 2024-07-09
`Bugfix` Handle Role Detail drawer closing properly

### LINKED ISSUE #2116: Clicking "x" on Role Details drawer via New Proposal page on desktop does wrong thing
It should just close the drawer, but instead it navigates the user back to Roles Index page, losing the entire pending proposal.

Da-Colon:
Closes #2116


adamgall:
@Da-Colon does this fix the bug where clicking "x" incorrectly navigated back to Roles Index of closing the drawer and staying on Create Proposal?

adamgall (approved):

PR #2111, opened 2024-07-09
Bug fixes

### LINKED ISSUE #2100: `[roles-0.1.0]` Add customNonce component to roles form

### LINKED ISSUE #2110: New Role Proposal remove trash can fix scrolling
On New Roles Proposal page
- remove trash can
- fix scrolling

DarksightKellar:
closes #2110
closes #2100


src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -15,86 +15,90 @@ const useHatsTree = () => {
   const { setHatsTree } = useRolesState();
   const ipfsClient = useIPFSClient();
 
-  useEffect(() => {
-    async function getHatsTree() {
-      if (hatsTreeId === undefined || hatsTreeId === null) {
-        return;
-      }
+  const getHatsTree = useCallback(async () => {
```
DarksightKellar, line 18: Exporting `getHatTree` from this hook to naively re-query the entire tree whenever any of the hats contract events fire

src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -15,86 +15,90 @@ const useHatsTree = () => {
...
     }
+  }, [chain.id, hatsSubgraphClient, hatsTreeId, ipfsClient, setHatsTree]);
 
+  useEffect(() => {
     getHatsTree();
-  }, [chain.id, hatsSubgraphClient, hatsTreeId, setHatsTree, ipfsClient]);
+  }, [chain.id, hatsSubgraphClient, hatsTreeId, setHatsTree, ipfsClient, getHatsTree]);
+
+  return {
+    getHatsTree,
```
DarksightKellar, line 100: This is the only functional change in this hook (plus moving `getHatsTree` out of the `useEffect`. The rest are whitespace diffs

src/pages/daos/[daoAddress]/roles/edit/summary/index.tsx
```diff
@@ -61,7 +61,7 @@ export default function EditProposalSummary() {
       <Show above="md">
         <Portal>
           <Box
-            position="fixed"
+            position="absolute"
```
Comment (DarksightKellar) [inline src/pages/daos/[daoAddress]/roles/edit/summary/index.tsx:64]: The fix for the role create proposal screen scroll issue

src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -15,86 +15,90 @@ const useHatsTree = () => {
...
+          hatsTreeId,
+        },
+      });
     }
+  }, [chain.id, hatsSubgraphClient, hatsTreeId, ipfsClient, setHatsTree]);
 
+  useEffect(() => {
     getHatsTree();
-  }, [chain.id, hatsSubgraphClient, hatsTreeId, setHatsTree, ipfsClient]);
+  }, [chain.id, hatsSubgraphClient, hatsTreeId, setHatsTree, ipfsClient, getHatsTree]);
```
adamgall, line 97: Probably don't need all of these dependencies here?

src/pages/daos/[daoAddress]/roles/edit/index.tsx
```diff
@@ -46,6 +47,8 @@ function RolesEdit() {
   const { rolesSchema } = useRolesSchema();
   const { hatsTree, hatsTreeId, getHat } = useRolesState();
 
+  useHatsListeners();
```
Comment (adamgall) [inline src/pages/daos/[daoAddress]/roles/edit/index.tsx:50]: This needs to be instantiated somewhere higher, so that it stays mounted and in-memory throughout the context of a loaded DAO in our app. Right now it'll only be listening if the user is in the RolesEdit page here.

src/hooks/DAO/useHatsListeners.ts
```diff
@@ -0,0 +1,107 @@
...
+  useEffect(() => {
+    if (!publicClient || !safe) return;
+
+    const hatsContract = getContract({
+      abi: HatsAbi,
+      address: getAddress(hatsProtocol),
+      client: publicClient,
+    });
+
+    if (!hatsContract) return;
```
adamgall, line 76: `hatsContract` won't be null or undefined, don't need this check.

src/hooks/DAO/useHatsListeners.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    hatsContract.watchEvent.HatCreated({
```
adamgall, line 78: For all of these followingn lines:

1. These event watchers return functions, which need to be called when it's time to cleanly destroy the listener streams. Specifically, that needs to happen when the `useEffect` re-runs, because we'll be creating a new event listener, so we gotta close the first one.
    ```ts
    useEffect(() => {
      const unwatch = contract.watchEvent.Foo({ onLogs: logs => console.log(logs) })
      return () => {
        unwatch()
      }
    }, [])
   
   ```
2. Right now all of these event listeners are scoped only to the global Hats protocol contract. These listener callbacks are being fired for _every single_ time these events get fired, regardless of the Hat Tree. Can we scope these to just our (currently loaded) Hat Tree?

adamgall (changes requested):
2 of the 3 issues in this PR are good to go. I'm going to push some new code to this branch to undo the implementation of the issue that i don't think is ready, then we can merge this and keep working in a new PR for that third issue.

adamgall (approved):
This looks good now, for the two issues.

PR #2124, opened 2024-07-10
Add pluralisation to roles text

### LINKED ISSUE #2113: Proper pluralization on generated proposal titles
When I create multiple roles the language in the title says "Add 2 new role" let's get that "s" in there
Screenshot 2024-07-09 at 10 13 55 AM

_Originally posted by @azadikoff in https://github.com/decentdao/decent-interface/discussions/2108#discussioncomment-9999953_

DarksightKellar:
closes #2113

### DIFF
--- a/src/i18n/locales/en/roles.json
+++ b/src/i18n/locales/en/roles.json
@@ -20,7 +20,10 @@
   "amount": "Amount",
   "starting": "Starting",
   "ending": "Ending",
-  "addedHats": "Add {{addedHatsCount}} new role",
-  "updatedHats": "Update {{updatedHatsCount}} role",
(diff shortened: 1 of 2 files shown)


adamgall (approved):
Approving with the assumption that somewhere behind the scenes is some logic in the localization library that checks if the `count` property in the string is > 1 and if so uses the `_other` flavor of the translation.

DarksightKellar:
> Approving with the assumption that somewhere behind the scenes is some logic in the localization library that checks if the `count` property in the string is > 1 and if so uses the `_other` flavor of the translation.
> 
> That's pretty neat

Yup, [that's right!](https://www.i18next.com/translation-function/plurals)

PR #2141, opened 2024-07-12
Dictate whether mainnets are supported in-app or not based on environment variable

adamgall:
After speaking with @SpookyAction-BTC, (and probably should be speaking with @decentdao/product and @decentdao/engineering about it too), it seemed most appropriate to disable all "non-testnet" networks on the dev site.

In practice this means that _only_ Sepolia is supported on dev.

Reasoning:
- we have half-baked functionality on dev (specifically, things that will break for users if they attempt to use them)
- Therefore we shouldn't let people using DAOs on networks with Real Economic Value to do things that are potentially breaking.

Developer note: I've added the env var to dev site on Netlify already.

### DIFF
--- a/.env
+++ b/.env
@@ -48,5 +48,8 @@ VITE_APP_SHUTTER_EON_PUBKEY=0x0e6493bbb4ee8b19aa9b70367685049ff01dc9382c46aed83f
 # index.html and WalletConnect metadata
 VITE_APP_SITE_URL="https://app.dev.decentdao.org"
 
+# Disable mainnets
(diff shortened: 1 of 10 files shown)


adamgall:
As I wrote out the body of this PR, the more and more I don't like this PR. It's _good_ to be able to test mainnet things, _real DAO things_, on the dev site.

I think we should just close this and get better/smarter at not shipping half-baked and potentially breaking things to dev.

adamgall:
Not doing this

PR #2145, opened 2024-07-12
`Roles` | Various role create form bugs

### LINKED ISSUE #2143: App crashes if enter invalid ethereum address into member input and click Save
1. create new multisig dao
2. go to roles page
3. edit roles
4. add role
5. enter "a" into name and description
6. paste a valid address into member
7. delete last character
8. click save
9. app crash

### LINKED ISSUE #2144: Cannot create proposal for making initial Hats tree
on dev

1. create new multisig safe
2. go to roles page
3. click edit
4. click add role
5. enter a name, description, address
6. click save
7. click create proposal
8. click submit proposal
9. nothing happens

This bug was introduced in https://github.com/decentdao/decent-interface/pull/2133

Da-Colon:
Fixes #2143
- This is fixed by fixing the schema to work properly preventing clicking Save button when its a invalid address

---
Fixes #2144 
- There was a problem with the structure of the schema. 
  - `.default(undefined)`: Adding this stopped Yup from changing the initial value from `undefined` to `{}` which was causing the `is` conditional to be true when it should be undefined
  - `.nullable()`: allows for the `setFieldValue(undefined)` that we are using to actually clear the field.

---
## Other Fixes
- I also noticed that there was some left over 'remove' code that no longer is needed and causing a bug.
- I also re-added the `meta.touched` and got that working properly by using the `onBlur` prop

### DIFF
--- a/src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
+++ b/src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
@@ -19,7 +19,7 @@ import { RoleFormValues, RoleValue } from '../types';
 export default function RoleFormCreateProposal({ close }: { close: () => void }) {
   const [drawerViewingRole, setDrawerViewingRole] = useState<RoleValue>();
   const { t } = useTranslation(['modals', 'common', 'proposal']);
-  const { values, handleSubmit, isSubmitting } = useFormikContext<RoleFormValues>();
+  const { values, isSubmitting, submitForm } = useFormikContext<RoleFormValues>();
   const editedRoles = useMemo(() => {
     return values.hats.filter(hat => !!hat.editedRole);
   }, [values.hats]);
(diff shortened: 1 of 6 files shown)


adamgall (changes requested):
#2143 is still occurring

#2144 seems good though ✅

Da-Colon:
> #2143 is still occurring
> 
> #2144 seems good though ✅

O I see. It was properly invalidating when it was a new role, but existing roles it was failing as the test was failing with an uncaught promise. adding a try/catch in the Yup validation test fixes it.

adamgall (approved):
Lookin good!

adamgall:
good points @mudrila

tomstuart123 (approved):
Both work for me now

DarksightKellar (approved):

DarksightKellar:
Pushed a commit to address Kyrill's comments on translation -- merging.

PR #2147, opened 2024-07-13
`2146` Run netlify dev locally

### LINKED ISSUE #2146: Local Netlify Functions use
I don't know about yall but that treasury toast is getting in the way. Sooo let's get Netlify Functions working in the frontend

Da-Colon:
Closes #2146 
This will allow the netlify function to properly work.

I'll let GPT explain the setup

---
1.	Application Port (targetPort):
	•	Purpose: This is the port where your actual development server (Vite, in this case) is running. It’s where your application is served from during development.
	•	Why Separate?: By running your development server on a separate port, you isolate the development environment and allow Netlify Dev to manage and reroute requests appropriately without interfering with the direct operation of the development server.
2.	Netlify Dev Proxy Port (port):
	•	Purpose: Netlify Dev runs on this port. It acts as a proxy, handling requests to your functions, managing redirects, rewrites, and providing features like server-side logic that mirror your production environment on Netlify.
	•	Why Proxy?: Using a proxy server allows you to integrate Netlify-specific features into your local development without modifying your application’s server configuration. The proxy intercepts requests and can inject additional functionality like serverless functions, which don’t run natively on Vite.
---

Now the way I have this currently setup changes the vite server to `8888` and netlify dev to run on `3000` allowing us to continue using 3000 as normal locally. But if we want to keeep the vite server setting along I can switch it back but we would need to use `8888` locally to use netlify function (might be fine to do that too)

Thoughts? @decentdao/engineering


Da-Colon:
Convinces my self otherwise on switching the vite port, I reverted that change

adamgall (approved):
This is great!

I think it would also be cool to see an optional env var called something like SILENCE_NETLIFY_CONNECTION_ERRORS that you could set true on local. So that way we could still not run the netlify dev server and also not get those toasts.

Maybe that doesn't need to be in this PR though.

I guess I'm sayin: I like seeing this in the package json scripts, but I don't think I'll like to run the netlify dev server all the time just to silence that toast.

DarksightKellar:
Agree with Adam's comment re: `SILENCE_NETLIFY_CONNECTION_ERRORS`, but gonna merge as is cz I MUST HAVE THIS FEATURE.

DarksightKellar (approved):

PR #2169, opened 2024-07-23
Calendar Updates!

DarksightKellar:
This PR implements a couple missing/buggy features on the original date picker implementation. There's no open issue for this.

- The "today" indicator was misaligned on mobile screen. The fix is a dirty hacky one. Added a todo comment to that effect -- would love a much cleaner approach, but CSS, man. CSS.
Screenshot 2024-07-23 at 17 43 00

- The "header" area of the date picker has been added
Screenshot 2024-07-23 at 17 46 20

- Date range picker + practically all UI design elements from figma! (not gonna call out the 2 or so TINY bits that I couldn't implement)
Screenshot 2024-07-23 at 17 44 57

Screenshot 2024-07-23 at 17 45 22


xanaramoss:
Uau @DarksightKellar this looks so good!!!
From the @decentdao/design side, I think the only thing missing is the active state on the end date field (like google flights for example). That would look way more clear for the user. I can add that to the designs tomorrow. 
Screenshot 2024-07-23 at 18 27 32

On the mobile, the input fields are breaking in lines. 
Screenshot 2024-07-23 at 18 33 22

xanaramoss:
@DarksightKellar is it possible to make the day item selector pink background a bit more square instead of rectangular?
Screenshot 2024-07-23 at 18 34 53

nicolaus-sherrill (changes requested):
UI is looking sharp so far @DarksightKellar. Great work.
Couple of small things:
1. Square up the selected date. Currently rectangular.
2. On desktop, align the pop-out component with the left side of the date input fields.

Photo from handoff designs for reference:
Role List - Edit Role Vesting Fixed Dates

tomstuart123:
Thanks @DarksightKellar . I had some issues with the date picker when first opened but haven't been able to replicate it since then. 

Won't approve yet as design had raises above but will approve once reverted :)

DarksightKellar:
> Uau @DarksightKellar this looks so good!!! From the @decentdao/design side, I think the only thing missing is the active state on the end date field (like google flights for example). That would look way more clear for the user. I can add that to the designs tomorrow. Screenshot 2024-07-23 at 18 27 32
> 
> On the mobile, the input fields are breaking in lines. Screenshot 2024-07-23 at 18 33 22

@xanaramoss  The line breaking inputs were due to a sheer lack of horizontal space on smaller width screens. We've worked out a good middle ground (shoutout to @Da-Colon for working his grid magic):
Screenshot 2024-07-25 at 18 39 41

Regarding "active state on the end date field", I'm not sure what the ask is there, and I don't yet see any design updates

DarksightKellar:
> UI is looking sharp so far @DarksightKellar. Great work. Couple of small things:
> 
> 1. Square up the selected date. Currently rectangular.
> 2. On desktop, align the pop-out component with the left side of the date input fields.

@nicolaus-sherrill Both points should be addressed now. Have another look when you get the chance =)

adamgall:
@DarksightKellar Yesterday during a screenshare of this feature on standup (which I didn't call out in the moment), I saw that it was possible to set the End Date to a date _before_ the Start Date, which I don't think should be possible. Can you confirm that I'm not just making this up?

tomstuart123:
@t

> @DarksightKellar Yesterday during a screenshare of this feature on standup (which I didn't call out in the moment), I saw that it was possible to set the End Date to a date _before_ the Start Date, which I don't think should be possible. Can you confirm that I'm not just making this up?

@adamgall if I remember my test two days ago, it is possible to click the end date before the start date. However, all that will happen is the start date will revert to that click and you'll still be forced to re-click an end date. I didn't hate that UX so I didn't raise it. 

I'll bring back up the netflify during standup to confirm

adamgall:
@tomstuart123 has confirmed that i'm full of shit

adamgall (approved):

DarksightKellar:
@nicolaus-sherrill your re-review on this plz 🥹

tomstuart123 (approved):

PR #2182, opened 2024-07-26
Disallow edit roles navigation if user cannot create proposals

### LINKED ISSUE #2168: Create roles proposal doesn't check if user can create a proposal
On an Azorius safe, we need to be checking if the user can indeed create proposals (see `canUserCreateProposal` in `useCanUserCreateProposal` hook in codebase), _before_ allowing them to click that button to create any roles edit proposal.

DarksightKellar:
Closes #2168 
Connected, delegated wallet:
Screenshot 2024-07-26 at 15 26 07

Unconnected wallet, or connect wallet but not yet delegated:
Screenshot 2024-07-26 at 15 26 21
Also blocked from role details:
Screenshot 2024-07-26 at 15 32 56

### DIFF
--- a/src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
+++ b/src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
@@ -12,8 +12,6 @@ import { useNetworkConfig } from '../../../../providers/NetworkConfig/NetworkCon
 import { CustomNonceInput } from '../../../ui/forms/CustomNonceInput';
 import { InputComponent, TextareaComponent } from '../../../ui/forms/InputComponent';
 import LabelWrapper from '../../../ui/forms/LabelWrapper';
-import { ModalType } from '../../../ui/modals/ModalProvider';
-import { useFractalModal } from '../../../ui/modals/useFractalModal';
 import { RoleCardEdit } from '../RoleCard';
 import RolesDetailsDrawer from '../RolesDetailsDrawer';
 import RolesDetailsDrawerMobile from '../RolesDetailsDrawerMobile';
--- a/src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
+++ b/src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
@@ -6,6 +6,7 @@ import { useNavigate } from 'react-router-dom';
 import { Hex } from 'viem';
 import { CARD_SHADOW } from '../../../../constants/common';
 import { DAO_ROUTES } from '../../../../constants/routes';
+import { useCanUserCreateProposal } from '../../../../hooks/utils/useCanUserSubmitProposal';
 import { useFractal } from '../../../../providers/App/AppProvider';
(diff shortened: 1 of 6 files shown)


src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
```diff
@@ -39,9 +39,7 @@ export default function RoleFormCreateProposal({ close }: { close: () => void })
     [navigate, addressPrefix, daoAddress],
   );
 
-  const handleCloseDrawer = () => {
-    setDrawerViewingRole(undefined);
-  };
+  const handleCloseDrawer = () => setDrawerViewingRole(undefined);
```
adamgall, line 42: Why this change? I think the original was correct: we probably don't need to return the result of `setDrawerViewingRole()` from `handleCloseDrawer`.

src/pages/daos/[daoAddress]/roles/index.tsx
```diff
@@ -40,7 +44,7 @@ function Roles() {
           },
         ]}
         buttonVariant="secondary"
-        buttonText={t('editRoles')}
+        buttonText={canUserCreateProposal ? t('editRoles') : undefined}
```
Comment (adamgall) [inline src/pages/daos/[daoAddress]/roles/index.tsx:47]: I see how this works (in PageHeader), I just don't like it very much.

adamgall (approved):
Approving, but it would be great to see some changes in PageHeader that allowed consumers to be more explicit with regards to deciding whether or not a button should be there.

tomstuart123:
@DarksightKellar / @adamgall - Assuming this PR just removes the button if not connected with Token. Just so you're aware, this does mean that if you connect wallet, click 'edit roles' and then disconnect and reload the page, you can still go through the flow even when disconnected. 

see video -- https://decentdao.slack.com/archives/C06QZTUTKRC/p1722006635219679?thread_ts=1722006582.990389&cid=C06QZTUTKRC

I don't think this is the end of the world as proposal still isn't submittable. It also would mean that the user would have to have a token at that session at some point. But just an fyi, we probably should align on before we approve

Let me know what you both think. Worth fixing or leave it for now?

tomstuart123 (approved):
approved as @adamgall and I discussed the 'disconnect and carry on editing' flow in a future issue that I'll create now

src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
DarksightKellar: Eh, it was more visual (code-wise) than functional, so not because I thought we'd need the result of `setDrawerViewingRole()`. If you _passionately_ prefer the former I'm happy to revert 👀

adamgall: I don't know about _passionately_, but I do prefer the code to do what we want and "nothing else".

However now that I actually look at the return types, this change _doesn't_ affect the return type of `handleCloseDrawer`, so this is all good with me 👍

PR #2200, opened 2024-08-02
Reset Safe state when daoAddress is goes null

DarksightKellar:
This PR moves the calls `action.resetSafeState()` and `resetHatsStore()` to the `DAOController` component, to reset the safe data if `daoAddress` becomes null.

Notes:

- The unmount route did not work. For some reason the callback kept getting called, and state kept getting reset, so safes were never loading. Resetting under `if (!daoAddress)` is what seems to work instead.

- `if (!daoName)` was similarly unsuccessful (I was hoping to simply reuse it in the new `useEffect`) -- it was getting unset too many times for some reason, causing safe data to never really load.

- Through console logging, I've noticed that `daoAddress` _doesn't actually get unset_ when navigating to the home page. Instead, when switching safes via any method currently possible in-app, it gets set to `null` first, _then_ it gets set to the new safe address. This is what triggers the reset in the end.

- I'm honestly not confident this PR 100% works. If it does, great, but I did hit the same bug a couple times, although it seemed it was related to a "too many requests" error on the alchemy key. When I waited a while and retested, all was fine. Buuut both scenarios could well have been flukes -- a false negative _and_ a false positive in succession.

Lol ok I just discovered some wonky crap on this PR when switching safes. If I'm on a safe's proposals page and change safes using the favourites menu, then quickly go the proposals page, the _old_ safe's proposals load, and the _old_ safe's address is in the page header's breadcrumbs. Maybe this isn't the way to go after all...

Or maybe I'm switching safes too quickly -- "too many requests" coming up again, and just before this, the issue disappeared with no code changes.

Dunno man lol.


adamgall:
@DarksightKellar to confirm if the wonky behavior is because of this code, or existed before this code.

DarksightKellar:
> @DarksightKellar to confirm if the wonky behavior is because of this code, or existed before this code.

Well well well. This wonkiness exists on `develop` too.

PR #2215, opened 2024-08-05
`1857` allow 0 custom nonce

### LINKED ISSUE #1857: Nonce 0? Do we need it
@Da-Colon tested this but the proposal to modify governance won't execute (note) on a 'Safe created in Safe.global's UI'; not within Decent. DAO link [here](https://deploy-preview-1854.app.dev.decentdao.org/home?dao=sep:0xf0415067c0a385146C3CCB45f5E92a8530065C38)

I did it a second time but with a 'Decent Safe' and modifying governance this time worked. DAO [link here](https://deploy-preview-1854.app.dev.decentdao.org/proposals/0xcf57d6a0572e0cc700f825f68b4afc37112a5b17524c0ff47fe23ce2f4a07e65?dao=sep:0xB4c83a6D1cb389f4087584EDefD640192FeCcd67)

I have a hypothesis but please challenge it @adamgall / @Da-Colon 

So we set the nonce by default on new Safes to be 1 and we don't allow changing it to zero in the UI. This is because we do an initial Tx on the safe to add some stuff for a Decent Safe. This takes up nonce 0 so it makes sense for our created Safes.

However, on a Multisig created from the Safe UI, we don't do that initial Tx. So the nonce defaults to 1 and skips without executing nonce 0. This isn't changeable by the user in the UI (for modify governance + adjust signers) either so the user is blocked doing stuff if its the first Tx they do on an imported safe in the Decent UI.

Could this be it? Or is it something else? 

If it is it, it still begs the question of whether we want people to write nonce zero just to fix this edge case instance? I personally, would like to give the user the option to set nonce zero in all areas of the app and just show the message in screenshot if they do click it
Screenshot 2024-05-22 at 10 06 40

_Originally posted by @tomstuart123 in https://github.com/decentdao/fractal-interface/issues/1854#issuecomment-2124265563_

Da-Colon:
Fixes #1857

### DIFF
--- a/src/components/ui/forms/CustomNonceInput.tsx
+++ b/src/components/ui/forms/CustomNonceInput.tsx
@@ -57,7 +57,7 @@ export function CustomNonceInput({
         helper={renderTrimmed ? '' : t('customNonceTooltip', { ns: 'proposal' })}
         isRequired={false}
         value={nonce?.toString() || ''}
-        onChange={e => onChange(e.target.value ? Number(e.target.value) : undefined)}
+        onChange={e => onChange(e.target.value !== undefined ? Number(e.target.value) : undefined)}
         disabled={disabled}
         subLabel={
           renderTrimmed ? null : (
--- a/src/components/ui/forms/CustomNonceInput.tsx
+++ b/src/components/ui/forms/CustomNonceInput.tsx
@@ -15,7 +15,7 @@ export function CustomNonceInput({
 }: {
   align?: 'start' | 'end';
   nonce: number | undefined;
-  onChange: (nonce?: number) => void;
+  onChange: (nonce?: string) => void;
   disabled?: boolean;
   renderTrimmed?: boolean;
 }) {
@@ -57,7 +57,12 @@ export function CustomNonceInput({
         helper={renderTrimmed ? '' : t('customNonceTooltip', { ns: 'proposal' })}
         isRequired={false}
         value={nonce?.toString() || ''}
-        onChange={e => onChange(e.target.value !== undefined ? Number(e.target.value) : undefined)}
+        onChange={e => {
+          const value = e.target.value;
+          if (/^\d*$/.test(value)) {
+            onChange(value);
+          }
+        }}
         disabled={disabled}
         subLabel={
           renderTrimmed ? null : (
--- a/src/components/ui/forms/CustomNonceInput.tsx
+++ b/src/components/ui/forms/CustomNonceInput.tsx
@@ -59,6 +59,8 @@ export function CustomNonceInput({
         value={nonce?.toString() || ''}
         onChange={e => {
           const value = e.target.value;
+
+          // @dev This regex /^\d*$/ ensures the input contains only digits (0-9).
           if (/^\d*$/.test(value)) {
             onChange(value);
           }
--- a/src/components/DaoCreator/formComponents/AzoriusGovernance.tsx
+++ b/src/components/DaoCreator/formComponents/AzoriusGovernance.tsx
@@ -24,7 +24,7 @@ export function AzoriusGovernance(props: ICreationStepProps) {
   const minutes = t('minutes', { ns: 'common' });
 
   const handleNonceChange = useCallback(
-    (nonce?: number) => {
+    (nonce?: string) => {
       setFieldValue('multisig.customNonce', nonce);
     },
     [setFieldValue],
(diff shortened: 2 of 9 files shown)


adamgall:
@Da-Colon @tomstuart123 hmm... I am testing this by first trying to recreate the issue on `dev`, but I can't.

What are the steps to recreate the bug?

Da-Colon:
Ah sorry create a Safe via Safe's Website/app. Then load up that safe in Decent UI. on dev. if you go to say create proposal builder and try to change the nonce to 0 it won't let you. even though `nonce` and `nextNonce` are both 0

adamgall:
After speaking with @Da-Colon , the issue that was repeatable for me, happens when _after deleting_ the "0" that appears by default in the Custom Nonce component. Not able to type a "0" into that component after deleting it.

adamgall (approved):

DarksightKellar (approved):

tomstuart123:
@Da-Colon this implementation unblocks the issue (i.e. nonce zero is now possible to be input manually)... but not as I expected. 

i.e. if I try and delete the input field to have no number in it (I.e. not zero but nothing) then a zero still remains. If you press backspace on any number, zero takes its place.

Was that intentional ? If so, any reason you chose that as a fix rather than just letting the input field work as normal

Da-Colon:
> @Da-Colon this implementation unblocks the issue (i.e. nonce zero is now possible to be input manually)... but not as I expected. 
> 
> i.e. if I try and delete the input field to have no number in it (I.e. not zero but nothing) then a zero still remains. If you press backspace on any number, zero takes its place.
> 
> Was that intentional ? If so, any reason you chose that as a fix rather than just letting the input field work as normal

Internationally setting 0? Not on my part. Only thing I updated really is a conditional that makes checks for undefined versus 0. You are experiencing  what was already there, just unblocked. 

To go through to make that change I'll need to make more changes

adamgall:
@tomstuart123 great find :pray:

Da-Colon:
> Gotcha. So I just played with it some more with app.dev and this PR side by side.
> 
> [Video'd here ](https://decentdao.slack.com/archives/C06QZTUTKRC/p1722952285441039?thread_ts=1722952266.007179&cid=C06QZTUTKRC)the experience
> 
> Takeaways:
> 
> * _**Solved**_ - This PR does solve the issue of not being able to type zero into an input field
> * _**Big Issue**_ - It reveals a new issue that didn't exist before. If you type anything that is 'NaN' then you cannot alter the field without a hard refresh (...even clicking 'back to proposal description' in our UI, doesn't solve it).
> * _**Not a problem**_It also reveals a new thing where you cannot leave the input field empty
> 
> The bottom one I can live with but not the second one, (i.e. one tap of the wrong key when typing nonce could lose them all the proposal settings they've done via a hard refresh), I don't think we should merge even though it solves the issue.
> 
> @Da-Colon how hard would it be to tackle the NaN issue? The third issue I can live with

oof and a nan issue too. yeah I agree @tomstuart123 not to long. just got make this code sync across all forms better and remove that default and let it be undefined

Da-Colon:
@tomstuart123 I'll need to do more testing. But I got it quicker than I thought. (and easier). lmk if you see it any where. (Also unblocks the error messages that we had setup already ha.

image
image

tomstuart123 (approved):
awesome , this worked in create proposal, send assets, add subsafe and add role. Approved

src/components/ui/forms/CustomNonceInput.tsx
```diff
@@ -57,7 +57,12 @@ export function CustomNonceInput({
         helper={renderTrimmed ? '' : t('customNonceTooltip', { ns: 'proposal' })}
         isRequired={false}
         value={nonce?.toString() || ''}
-        onChange={e => onChange(e.target.value ? Number(e.target.value) : undefined)}
+        onChange={e => {
+          const value = e.target.value;
+          if (/^\d*$/.test(value)) {
```
adamgall, line 62: @Da-Colon just because im a simpleton, can you leave a code comment here explaining what pattern this is testing for?

adamgall (approved):
Approved, but with small request for comment

src/components/ui/forms/CustomNonceInput.tsx
Da-Colon: 100%

Issue #2084, opened 2024-07-06
`[roles-0.1.0]` Bug | Roles Data doesn't refresh


Da-Colon:
When switching DAOs, Roles data is not refreshed and remains the Roles of the previous DAO until page refresh

adamgall:
@Da-Colon I'm unable to replicate this on the current tip of `roles-0.1.1/desktop-and-other-updates`, at commit `bcdf9eddc455d444b02a4f0b9c22c9389b18df2f`.

@Da-Colon @mudrila @DarksightKellar are you?

adamgall:
On a call with @DarksightKellar he experienced this during some unknown situation and/or scenario.

Still unsure how to replicate, but it happens...

adamgall:
@DarksightKellar you mentioned that you know how to reproduce it. Please share those steps here

DarksightKellar:
> @DarksightKellar you mentioned that you know how to reproduce it. Please share those steps here

Most reliable way I've found to repro:

- Be in a safe that has some roles
- Go back to homepage, create a new safe
- After safe is deployed, go to its roles page
- See that the roles from the previous safe shows here until a refresh

adamgall:
Another way to replicate:

https://github.com/decentdao/decent-interface/issues/2179#issue-2430324329

adamgall:
Closed via https://github.com/decentdao/decent-interface/pull/2188

PR #2219, opened 2024-08-08
Revert safe reset refactor

DarksightKellar:
In a previous commit, we moved the calls to `resetSafeState` and `resetHatsStore` from the home page to the dao controller component, as that felt like the more logical place to put it. 

While the argument is still valid, this just led to more bugs due to the brittle nature of the flow of safe state changes.

We have tried moving the `reset`s higher up the hierarchy into `Layout/index.tsx` with slightly better result, but far from good enough. 

That said, the decision was made to revert this move for now. This fixes the original bug described in #2084, with the caveat that there are edge cases around switching safes instantly using the search bar, when already on the roles page of the current safe.

This will have to do for now as the most common UX flows to end up on different safes should be fine. I do plan to continue to dig into the original safe state issues in the meantime until more pressing issues get on the radar.

### DIFF
--- a/src/pages/DAOController.tsx
+++ b/src/pages/DAOController.tsx
@@ -5,7 +5,6 @@ import { toast } from 'react-toastify';
 import useDAOController from '../hooks/DAO/useDAOController';
 import { useUpdateSafeData } from '../hooks/utils/useUpdateSafeData';
 import { useFractal } from '../providers/App/AppProvider';
-import { useRolesStore } from '../store/roles';
 import LoadingProblem from './LoadingProblem';
 
 const useTemporaryProposals = () => {
@@ -34,12 +33,9 @@ const useTemporaryProposals = () => {
 
 export default function DAOController() {
   const { errorLoading, wrongNetwork, invalidQuery } = useDAOController();
-  const { resetHatsStore } = useRolesStore();
-
   useUpdateSafeData();
   const {
-    node: { daoName, daoAddress },
-    action,
+    node: { daoName },
   } = useFractal();
 
   useTemporaryProposals();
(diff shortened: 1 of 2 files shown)


DarksightKellar:
@tomstuart123 Could you checkout the description on this PR please and test to confirm the fix on your end?

tomstuart123 (approved):

tomstuart123:
@DarksightKellar thanks man, very clear description 

I just jumped between Safe's via the homepage-->DAOhome-->DAO roles--> homepage --> etc. This worked 100% of the time on 6 tries.

I was able - as you mentioned to replicate the edge case you mentioned with the search bar. However, I can confirm that this is fine. To replicate it you need to move super fast in the UI (i.e. power users) and even then it does reload automatically eventually

Approved

tomstuart123:
Question @DarksightKellar / @decentdao/engineering - 

will this reverted change of 'call' location have any other user impact? Or was it just a general 'code efficiency' / 'logical place' reason ?

DarksightKellar:
> Question @DarksightKellar / @decentdao/engineering -
> 
> will this reverted change of 'call' location have any other user impact? Or was it just a general 'code efficiency' / 'logical place' reason ?

@tomstuart123 Yeah this revert just sent back logic to the way it was already on develop. Plus the "fix" on top of that. The move (that ended up not working out) was meant to patch up edge case resetting logic flaws that do exist right now regardless and were discovered while fixing the original issue.

For example, having the reset logic exist in the homepage means that going to the homepage is the only to _reliably_ reset safe state. If the "homepage" changes at some point in the future, that affects this guarantee.

PR #2221, opened 2024-08-08
`2218` BugFix: Queued Transactions not showing up

### LINKED ISSUE #2218: Bugfix: `allTransactions` request doesn't include queued transactions anymore
Multisig Safes will not be able to see any queued/created proposal in our UI. Safe has removed queued transactions from endpoint in lastest release of API.

Release:
https://github.com/safe-global/safe-transaction-service/releases/tag/v5.7.0

PR:
https://github.com/safe-global/safe-transaction-service/pull/2166

There is another endpoint/sdk method that returns all the transactions, `getMultisigTransactions`.  But we lose out on `transfers` that is included in `getAllTransactions`. 

See `useSafeTransactions` hook for where it starts to effect things (haven't deep dive on what exactly this effects). This doesn't effect the treasury page as that is a different request. I believe we just do some parsing on the transfer to detect what will be transfer to who and only display it if there is no Proposal Title/Description. But maybe this isn't needed here anymore?

Da-Colon:
Fixes #2218

This PR replaces the `getAllTransaction` API request with `getMultisigTransactions` request. Most of the properties are the same except for the `transfers` property.

There are two spots where this is changes somethings.

Dashboard:
Will no longer show Treasury activities. These would be either a Card itself or added text to the bottom of the description that would display any transfer information.

Treasury:
~In Progress: But this also used `getAllTransactions`. will be need to be updated to `getIncomingTransactions`, but we lose the easily asseible `tokenInfo` in the transfer data.

~Update: Was able to use a combination of `getIncomingTransactions` and `getToken` from Safe's `api-kit` to retrieve the same data and put Treasury back together.


Da-Colon:
Closing for new branch based off main

PR #2222, opened 2024-08-08
Debug reset safe state

DarksightKellar:
### UPDATE:

@decentdao/engineering has decided it is time to take a deeper look at the application's state management. 

To that end, this PR will now serve as the first of a series of PRs with the end goal of migrating to Zustand, AND optimising our state management. 

The logs and guard checks in this PR may or may not change after the work is complete.

In any case we're leaving the changes in - to be merged as _part_ of the state overhaul work - so we can validate that the roles loading issues described below have in fact been fixed. Then we can cleanup whatever needs cleaning up, if any, from the changes on this PR.

---

ORIGINAL DESCRIPTION:

I’m making some progress with the safe state stuff. While there still is the issue of repeated calls to the same set of functions (maybe this already existed and only became visible to me because of debug logging), I think I’ve got to the point where safe changes via the search box do correctly reload its roles.

However.

There a small issue where, only _sometimes_, roles loading fails. We have seen this issue before during pair programming earlier, I think — that was when the roles simply didn’t load for no reason. This is that reason. The hats tree do load, but `sanitize` throws, so roles never get set in state. Now, every time this error has happened, a simple hot reload fixed it. Not even a browser reload — just a “save random VSCode file” reload.

So… the contract call fails randomly, but only when changing safes, BUT it will inevitably work the very next time, implying that if we put it in a retry loop of 2 max calls it’ll work 100% of the time.

Is there any reason that you can think of for this to be the case? I’m stumped atp. Is this worth digging further into?

### DIFF
--- a/src/components/ui/page/Layout/index.tsx
+++ b/src/components/ui/page/Layout/index.tsx
@@ -1,6 +1,6 @@
 import { Box, Container, Grid, GridItem, Show } from '@chakra-ui/react';
-import { useRef } from 'react';
-import { Outlet } from 'react-router-dom';
+import { useEffect, useRef } from 'react';
+import { Outlet, useSearchParams } from 'react-router-dom';
 import {
   useContentHeight,
   useHeaderHeight,
@@ -9,6 +9,8 @@ import {
   useFooterHeight,
 } from '../../../../constants/common';
 import useNavigationScrollReset from '../../../../hooks/utils/useNavigationScrollReset';
+import { useFractal } from '../../../../providers/App/AppProvider';
+import { useRolesStore } from '../../../../store/roles';
 import { ErrorBoundary } from '../../utils/ErrorBoundary';
 import { TopErrorFallback } from '../../utils/TopErrorFallback';
 import { Footer } from '../Footer';
@@ -17,6 +19,21 @@ import { NavigationLinks } from '../Navigation/NavigationLinks';
 
 export default function Layout() {
   const headerContainerRef = useRef<HTMLDivElement>(null);
+
+  const [searchParams] = useSearchParams();
+  const safeParam = searchParams.get('dao');
+
+  const { resetHatsStore } = useRolesStore();
+
+  const { action } = useFractal();
+
+  useEffect(() => {
+    if (!safeParam) {
+      console.log('reset safe state from layout');
+      action.resetSafeState();
+      resetHatsStore();
+    }
+  }, [action, resetHatsStore, safeParam]);
 
   const HEADER_HEIGHT = useHeaderHeight();
   const CONTENT_HEIGHT = useContentHeight();
--- a/src/hooks/DAO/loaders/useFractalNode.ts
+++ b/src/hooks/DAO/loaders/useFractalNode.ts
@@ -141,7 +141,7 @@ export const useFractalNode = (
       addressPrefix + daoAddress !== currentValidSafe.current
     ) {
       console.count('reset hats store from useFractalNode because:');
-      console.log({ skip, currentValidSafe: currentValidSafe.current, addressPrefix, daoAddress });
+      console.log({ skip, currentValidSafe: currentValidSafe.current, daoAddress });
       reset({ error: false });
       if (addressPrefix && daoAddress) {
         setDAO(addressPrefix, daoAddress);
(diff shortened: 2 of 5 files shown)


adamgall:
@DarksightKellar please deal with these build errors, and then we could probably mark this PR as Ready for review.

src/components/ui/page/Layout/index.tsx
```diff
@@ -18,6 +20,21 @@ import { NavigationLinks } from '../Navigation/NavigationLinks';
...
+
+  const { resetHatsStore } = useRolesStore();
+
+  const { action } = useFractal();
+
+  useEffect(() => {
+    if (!safeParam) {
+      console.log('reset safe state from layout');
+      action.resetSafeState();
+      resetHatsStore();
```
DarksightKellar, line 34: We would ideally like to have the "safe state reset" logic right here, so that whenever `stateParam` is undefined, the reset happens.

`stateParam` will go undefined when the user performs any navigation that takes them outside of any safe's context, for example, navigating to the homepage, or the create safe page(s).

src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -119,6 +126,20 @@ const useHatsTree = () => {
...
+          const fetchedHatsTreeId = !!firstHatId
+            ? hatIdToTreeId(hexToBigInt(firstHatId))
+            : undefined;
+
+          console.log('in tree load');
+          console.log({ fetchedHatsTreeId, hatsTreeId, treeWithFetchedDetails });
+
+          if (fetchedHatsTreeId !== hatsTreeId) {
+            console.log('Hats Tree ID does not match the fetched Hats Tree ID');
+            return;
```
DarksightKellar, line 139: Sometimes, probably due to super-quick navigations combined with network requests that have already gone out, `fetchedHatsTreeId`, which is simply the hat tree id of the first (or any) hat in `treeWithFetchedDetails`, would have a different value from `hatsTreeId`, which comes from app state. 

`hatsTreeId` will (should, at least) have the actual, proper hat id of the newly-navigated-to context safe.

So if there's that disconnect, don't call `setHatsTree` to poison the roles state. 

This logic may or may not be necessary after the state overhaul. Hopefully not. We'll see.

src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -170,6 +192,21 @@ const useHatsTree = () => {
...
+
+      const firstHatId = hatsTree?.roleHats?.at(0)?.id;
+      const fetchedHatsTreeId = !!firstHatId ? hatIdToTreeId(hexToBigInt(firstHatId)) : undefined;
+
+      console.log('in payments');
+      console.log({ fetchedHatsTreeId, hatsTreeId });
+
+      if (fetchedHatsTreeId !== hatsTreeId) {
+        console.log('Hats Tree ID does not match the fetched Hats Tree ID');
+        return;
```
DarksightKellar, line 207: Same story here

src/hooks/DAO/loaders/useHatsTree.ts
adamgall: yep

src/hooks/DAO/loaders/useHatsTree.ts
adamgall: yep

adamgall (approved):

adamgall:
@DarksightKellar feel free to merge at your discretion

src/components/ui/page/Layout/index.tsx
DarksightKellar: > do u mean `safeParam`?
> 
> also sorry just to be clear, when you say "we would ideally like to have the 'safe state reset' logic right here...", you mean like how it is currently implemented here in this PR right? i'm just questioning your use of future tense.

@adamgall Yep. Edit my comment to be clearer.

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -92,13 +94,16 @@ export const useFractalNode = (
...
+      resetHatsStore();
       setErrorLoading(error);
     },
-    [action],
+    [action, resetHatsStore],
   );
 
   const setDAO = useCallback(
     async (_addressPrefix: string, _daoAddress: string) => {
+      console.log('setDAO', _daoAddress);
```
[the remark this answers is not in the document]
DarksightKellar: Yep all these logs are meant to remain right until the state overhaul project is complete and verified working. I think I'm gonna add tags to them so they're easy to delete later

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -135,6 +140,8 @@ export const useFractalNode = (
       daoAddress === undefined ||
       addressPrefix + daoAddress !== currentValidSafe.current
     ) {
+      console.count('reset hats store from useFractalNode because:');
+      console.log({ skip, currentValidSafe: currentValidSafe.current, daoAddress });
```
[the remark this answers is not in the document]
DarksightKellar: Yep

PR #2233, opened 2024-08-12
Sablier proposal builder

### LINKED ISSUE #2220: Shutter Sablier custom proposal
Reviving a request from Shutter to build custom Sablier proposal since our Proposal Builder can't handle nested tuples in arrays 😿

mudrila:
In order to use `Tranched Sablier Stream Builder` - please visit DAO's Sablier proposal creation link. 
It should look like this:

https://deploy-preview-2233.app.dev.decentdao.org/proposals/new/sablier/metadata?dao=sep:0x7d0E8e93B5785CB4d5c76f65BB798B3d0a3e89Cb

For instance, for `Shutter Mainnet DAO` - the link will be following:

https://deploy-preview-2233.app.dev.decentdao.org/proposals/new/sablier/metadata?dao=eth:0x36bD3044ab68f600f6d3e081056F34f2a58432c4

Fill in metadata - proposal title and description (markdown supported) and proceed to the next step - configuring streams.
You can create multiple streams, though note it only supports **Tranched** streams creation.

Paste in following information for each stream you would like to create:

1. Token address - after pasting Token Contract address - you should see your DAO balance of this token, it's symbol and name. Make sure it matches the token you'd like to stream from the Safe
2. Recipient address - the receiver address. Only this address will be able to execute `withdraw` transaction through Sablier UI
3. Total amount - the total amount to be streamed.
4. Cancelable - whether this stream can be cancelled by DAO through further proposal
5. Transferable - whether recipient should be able to transfer stream receiving rights to some other recipient (original recipient will completely loose access to withdrawing funds from this stream)
6. Tranches
    - Amount: the amount to stream per tranche
    - Duration: the duration of the tranche streaming **in seconds** (the tranche will be available for withdraw only after passing this duration)
    - Please note - the sum of tranches amounts **needs to be exactly the same as stream total amount**
    - First tranche will start streaming right after proposal execution - this is technical detail enforced from Sablier side. If you'd like to emulate postponed start - put `0` into very first tranche "Amount" field and desired duration of delay in seconds. Take into account that this duration will start only after executing proposal - so you'll need to account for proposal voting period

After configuring streams - proceed to submitting proposal. This should result in creating proposal with a set of transaction that will look like on the screenshot below

Screenshot 2024-08-12 at 16 33 20

If you'll be creating multiple streams **with the same token streamed** - that should result in 2 transactions with correct amount for `approve` function and with proper amount of streams created for `createWithDurationsLT` function.

If you'll be creating multiple streams **with different tokens streamed** - that will result in multiple invocations of `approve` and `createWithDurationsLT`, one per token respectively.

Nevertheless functionality is tested on `Sepolia` - please, perform your own round of testing with desired number of streams, token to be streamed and streams configuration - then verify through `Sablier` UI, that stream is created with proper parameters and distribution curve.

Closes #2220


adamgall:
@mudrila can we try to get this merged into `develop`? Let's put access to this feature behind some conditionals to narrow it down to one or two people. Something like

```
if (current_user == 0xabc && current_safe == 0x36b && network == mainnet)
  show_sablier_feature
```

adamgall:
Oh, ok, I think that's fine then. Will merge

PR #2247, opened 2024-08-14
Normalise `daoSnapshotENS` to lower case

### LINKED ISSUE #2234: Don't allow Snapshot ENS names with capitalization
Our app allows users to set (and save, ONCHAIN) ENS names which include capital letters.

Although this is "allowed", it's against convention.

`resolve(foo.eth) !== resolve(Foo.eth)`, but _every ENS tool out there_ (except for Decent) normalizes a user typing "Foo.eth" into "foo.eth". 

To see what I mean in practice, check out the official ENS app:

Image
Image

From the point of view of the app, _there are no capital letters in ENS names_.

The Decent app should "normalize" user input into all lowercase.

mudrila:
Closes #2234

### DIFF
--- a/src/components/DaoCreator/formComponents/EstablishEssentials.tsx
+++ b/src/components/DaoCreator/formComponents/EstablishEssentials.tsx
@@ -5,7 +5,7 @@ import { URL_DOCS_GOV_TYPES } from '../../../constants/url';
 import { createAccountSubstring } from '../../../hooks/utils/useDisplayName';
 import { useFractal } from '../../../providers/App/AppProvider';
 import { useNetworkConfig } from '../../../providers/NetworkConfig/NetworkConfigProvider';
-import { ICreationStepProps, VotingStrategyType, GovernanceType } from '../../../types';
+import { GovernanceType, ICreationStepProps, VotingStrategyType } from '../../../types';
 import { InputComponent, LabelComponent } from '../../ui/forms/InputComponent';
 import LabelWrapper from '../../ui/forms/LabelWrapper';
 import { RadioWithText } from '../../ui/forms/Radio/RadioWithText';
@@ -34,7 +34,7 @@ export function EstablishEssentials(props: ICreationStepProps) {
       setFieldValue('essentials.daoName', daoName, false);
       if (createAccountSubstring(daoAddress!) !== daoName) {
         // Pre-fill the snapshot URL form field when editing
-        setFieldValue('essentials.snapshotENS', daoSnapshotENS || '', false);
+        setFieldValue('essentials.snapshotENS', daoSnapshotENS?.toLowerCase() || '', false);
       }
     }
   }, [setFieldValue, mode, daoName, daoSnapshotENS, isEdit, daoAddress]);
@@ -70,7 +70,7 @@ export function EstablishEssentials(props: ICreationStepProps) {
           helper={t('helperDAOName')}
           isRequired
           value={values.essentials.daoName}
-          id="searchEssentials-daoName"
+          id="essentials-daoName"
           onChange={cEvent => setFieldValue('essentials.daoName', cEvent.target.value, true)}
           onBlur={cEvent => setFieldValue('essentials.daoName', cEvent.target.value.trim(), true)}
           disabled={daoNameDisabled}
--- a/src/components/DaoCreator/formComponents/EstablishEssentials.tsx
+++ b/src/components/DaoCreator/formComponents/EstablishEssentials.tsx
@@ -34,7 +34,7 @@ export function EstablishEssentials(props: ICreationStepProps) {
       setFieldValue('essentials.daoName', daoName, false);
       if (createAccountSubstring(daoAddress!) !== daoName) {
         // Pre-fill the snapshot URL form field when editing
-        setFieldValue('essentials.snapshotENS', daoSnapshotENS?.toLowerCase() || '', false);
+        setFieldValue('essentials.snapshotENS', daoSnapshotENS || '', false);
       }
     }
   }, [setFieldValue, mode, daoName, daoSnapshotENS, isEdit, daoAddress]);
--- a/src/components/DaoCreator/formComponents/EstablishEssentials.tsx
+++ b/src/components/DaoCreator/formComponents/EstablishEssentials.tsx
@@ -159,7 +159,7 @@ export function EstablishEssentials(props: ICreationStepProps) {
             <Input
               value={values.essentials.snapshotENS}
               onChange={cEvent =>
-                setFieldValue('essentials.snapshotENS', cEvent.target.value, true)
+                setFieldValue('essentials.snapshotENS', cEvent.target.value.toLowerCase(), true)
               }
               isDisabled={snapshotENSDisabled}
               data-testid="essentials-snapshotENS"
--- a/src/components/Proposals/ProposalInfo.tsx
+++ b/src/components/Proposals/ProposalInfo.tsx
@@ -1,4 +1,4 @@
-import { Box, Flex, Button, Link } from '@chakra-ui/react';
+import { Box, Button, Flex, Link } from '@chakra-ui/react';
 import { ArrowUpRight } from '@phosphor-icons/react';
 import { useTranslation } from 'react-i18next';
 import { Shield } from '../../assets/theme/custom/icons/Shield';
@@ -53,7 +53,9 @@ export function ProposalInfo({
         />
         {isSnapshotProposal && (
           <>
-            <SnapshotButton snapshotENS={`${daoSnapshotENS}/proposal/${proposal.proposalId}`} />
+            <SnapshotButton
+              snapshotENS={`${daoSnapshotENS?.toLowerCase()}/proposal/${proposal.proposalId}`}
+            />
             {(proposal as ExtendedSnapshotProposal).privacy === 'shutter' && (
               <Button
                 as={Link}
--- a/src/components/Proposals/ProposalInfo.tsx
+++ b/src/components/Proposals/ProposalInfo.tsx
@@ -53,9 +53,7 @@ export function ProposalInfo({
         />
         {isSnapshotProposal && (
           <>
-            <SnapshotButton
-              snapshotENS={`${daoSnapshotENS?.toLowerCase()}/proposal/${proposal.proposalId}`}
-            />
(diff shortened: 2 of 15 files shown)


src/components/Proposals/ProposalInfo.tsx
```diff
@@ -53,7 +53,9 @@ export function ProposalInfo({
         />
         {isSnapshotProposal && (
           <>
-            <SnapshotButton snapshotENS={`${daoSnapshotENS}/proposal/${proposal.proposalId}`} />
+            <SnapshotButton
+              snapshotENS={`${daoSnapshotENS?.toLowerCase()}/proposal/${proposal.proposalId}`}
```
adamgall, line 57: Hmm, I think we shouldn't touch this one. Let's assume that the data onchain is accurate for now.

src/components/DaoCreator/formComponents/EstablishEssentials.tsx
```diff
@@ -34,7 +34,7 @@ export function EstablishEssentials(props: ICreationStepProps) {
       setFieldValue('essentials.daoName', daoName, false);
       if (createAccountSubstring(daoAddress!) !== daoName) {
         // Pre-fill the snapshot URL form field when editing
-        setFieldValue('essentials.snapshotENS', daoSnapshotENS || '', false);
+        setFieldValue('essentials.snapshotENS', daoSnapshotENS?.toLowerCase() || '', false);
```
adamgall, line 37: Again here, I think we should leave this alone.

src/components/pages/DaoSettings/components/MetadataContainer.tsx
```diff
@@ -38,7 +38,7 @@ export function MetadataContainer() {
     }
 
     if (daoSnapshotENS) {
-      setSnapshotENS(daoSnapshotENS);
+      setSnapshotENS(daoSnapshotENS.toLowerCase());
```
adamgall, line 41: This is another spot where we're reading `daoSnapshotENS` from state, which came from what was onchain. Let's leave this alone, and assume that whatever was onchain and now is in state, is valid and accurate.

I want this PR to only touch the places where users are actively entering an ENS name. Only on that input action should we be normalizing into lowercase.

adamgall (changes requested):
Didn't go through the whole PR yet, but saw enough to stop.

Please see my comments! Thank you sir

src/components/DaoCreator/formComponents/EstablishEssentials.tsx
adamgall: But my point is that if someone is putting something non-normalized onchain, they're doing that _outside_ of our UI, and therefore who are we to assume it's wrong?

src/components/pages/DaoSettings/components/MetadataContainer.tsx
adamgall: Hm i think we just inherently disagree on this one.

The root of my position is that the reach of our responsibility should include only what people actually do within our app.

A `WiErD.eth` address is technically valid in ENS. ENS tooling convention is to direct users towards all-lowercase when entering/creating ENS names. But what's on-chain is on-chain, and it's NOT WRONG if it's non-normalized.

Just because the ENS name has capital letters in it doesn't immediately mean our app will become buggy.

I'd like other's take on this @Da-Colon @DarksightKellar

Da-Colon: hmmm We are not currently creating ENS name in the app, and in IMO, seems like we shouldn't try to mess with what is coming from onchain (to Adam's point). 

As far as ENSSnapshot, I don't think have a big opinion but I always lead to standards. Since this is something that is saved to our stuff. might be a good idea to just add a quick change to the one input to prevent uppercase letters from the git go and call it a day.

DarksightKellar: Yeahh I'm on the same page as the others -- don't touch what's already onchain. If some user somewhere _intentionally_ set a `wEiRdEnS.eth` outside of conventional/standard UI, they probably want to see it everywhere.

I don't see any issue at all with showing such mixed casing in-app -- that's not our responsibility, no will it cause any bug. At the same time there's still merit in keeping with established standards when the input comes _from_ Decent.

DarksightKellar (changes requested):
In the same theme of the ongoing discussion on this PR, I'd much rather the "lower-casing" happens only when it is input from the user via the Decent app.

Thus, as long as the ENS string came from Decent, it should be _guaranteed_ to be lower case, so we should not need to lower case them anywhere else in-app. If it _happens_ to be mixed case, it is most certainly not originally from Decent (at least after this PR goes live), and I don't think we have any business insisting that is MUST be lowercase for whoever decided (or whatever) they'd like their name non-standardised.

src/components/DaoCreator/formComponents/EstablishEssentials.tsx
adamgall: Sorry ser you're outnumbered 3 to 1

src/components/Proposals/ProposalInfo.tsx
adamgall: Sorry ser you're outnumbered 3 to 1

src/components/ui/cards/DAOInfoCard.tsx
```diff
@@ -111,7 +111,7 @@ export function DAOInfoCard() {
         <AddressCopier address={displayedAddress} />
 
         {/* SNAPSHOT ICON LINK */}
-        {node.daoSnapshotENS && <SnapshotButton snapshotENS={node.daoSnapshotENS} />}
+        {node.daoSnapshotENS && <SnapshotButton snapshotENS={node.daoSnapshotENS.toLowerCase()} />}
```
adamgall, line 114: Remove this `.toLowerCase()`

src/components/ui/cards/DAONodeInfoCard.tsx
```diff
@@ -104,7 +104,9 @@ export function DAONodeInfoCard({ node, freezeGuard, guardContracts, ...rest }:
             />
 
             {/* SNAPSHOT ICON LINK */}
-            {node.daoSnapshotENS && <SnapshotButton snapshotENS={node.daoSnapshotENS} />}
+            {node.daoSnapshotENS && (
+              <SnapshotButton snapshotENS={node.daoSnapshotENS.toLowerCase()} />
```
adamgall, line 108: Remove this `.toLowerCase()`

src/hooks/DAO/loaders/snapshot/useSnapshotSpaceName.ts
```diff
@@ -5,5 +5,5 @@ export default function useSnapshotSpaceName() {
     node: { daoSnapshotENS },
   } = useFractal();
 
-  return daoSnapshotENS;
+  return daoSnapshotENS?.toLowerCase();
```
adamgall, line 8: Remove this `.toLowerCase()`

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -53,7 +53,7 @@ export const useFractalNode = (
           },
           daoName: name as string,
           daoAddress: getAddress(_daoAddress),
-          daoSnapshotENS: snapshotENS as string,
+          daoSnapshotENS: snapshotENS?.toLowerCase() as string,
```
adamgall, line 56: Remove this `.toLowerCase()`

src/hooks/DAO/loaders/useLoadDAONode.ts
```diff
@@ -42,7 +42,7 @@ export const useLoadDAONode = () => {
           },
           daoName: name as string,
           daoAddress: getAddress(_daoAddress),
-          daoSnapshotENS: snapshotENS as string,
+          daoSnapshotENS: snapshotENS?.toLowerCase() as string,
```
adamgall, line 45: Remove this `.toLowerCase()`

src/hooks/utils/useAvatar.ts
```diff
@@ -4,7 +4,7 @@ import { useNetworkConfig } from '../../providers/NetworkConfig/NetworkConfigPro
 const useAvatar = (name: string) => {
   const { chain } = useNetworkConfig();
   const { data: avatarURL } = useEnsAvatar({
-    name: name,
+    name: name?.toLowerCase(),
```
adamgall, line 7: Remove this `.toLowerCase()`

src/models/DaoTxBuilder.ts
```diff
@@ -228,7 +228,7 @@ export class DaoTxBuilder extends BaseTxBuilder {
     return buildContractCall(
       this.baseContracts.keyValuePairsContract,
       'updateValues',
-      [['snapshotENS'], [this.daoData.snapshotENS]],
+      [['snapshotENS'], [this.daoData.snapshotENS.toLowerCase()]],
```
adamgall, line 231: Remove this `.toLowerCase()`.

I want the `.toLowerCase()` to happen much closer to the actual input component.

adamgall (changes requested):

src/hooks/DAO/loaders/snapshot/useSnapshotProposal.ts
```diff
@@ -108,7 +107,7 @@ export default function useSnapshotProposal(proposal: FractalProposal | null | u
       const { choices, type, privacy } = proposalQueryResult;
 
       if (type === 'weighted') {
-        Object.keys(choices).forEach((choice: string, choiceIndex) => {
+        Object.keys(choices).forEach((_choice: string, choiceIndex) => {
```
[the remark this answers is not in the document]
DarksightKellar: The beers ma dude

DarksightKellar (changes requested):
I might be missing something, but is the snapshot input field in this screenshot covered by this PR (or is something else entirely lowercasing that value)?
Screenshot 2024-08-16 at 14 00 50

adamgall:
@DarksightKellar i would expect that these two inputs are the two places that this PR should touch. What change are you requesting?

adamgall (changes requested):
@mudrila maybe this is why @DarksightKellar requested changes... does this PR do anything anymore? It doesn't seem like any of the changes here do the thing I was looking for: lowercasing text typed into those two input fields Kelvin screenshotted

adamgall (approved):

PR #2268, opened 2024-08-19
`Roles | Update: 2267` Update DatePicker components

### LINKED ISSUE #2267: Bugfix: DecentDatePicker jankyiness
- Make sure Date is correctly time to UTC.
- Selecting Date Range isn't smooth

Da-Colon:
Closes #2267
Did quite a refactor to the DatePicker. I branched off my previous PRs (probably will have conflicts when caught up to root (but I'll fix em). But need those errors cleared to work and test

- Separated Range Versus Single Date Logic
- Removed local component state
- moved drawer operating component `PaymentDatePicker` into `DecentDatePicker`'s container component.

### DIFF
--- a/src/components/ui/utils/DecentDatePicker.tsx
+++ b/src/components/ui/utils/DecentDatePicker.tsx
@@ -243,14 +243,13 @@ export function DecentDatePickerRange({
 }) {
   const { values } = useFormikContext<RoleFormValues>();
 
-  const selectedRange = useMemo(() => {
-    if (values.roleEditing?.roleEditingPaymentIndex === undefined)
-      return [null, null] as [DateOrNull, DateOrNull];
+  const selectedRange: [DateOrNull, DateOrNull] = useMemo(() => {
+    if (values.roleEditing?.roleEditingPaymentIndex === undefined) return [null, null];
     const paymentIndex = values.roleEditing.roleEditingPaymentIndex;
     return [
       values.roleEditing?.payments?.[paymentIndex]?.startDate ?? null,
       values.roleEditing?.payments?.[paymentIndex]?.endDate ?? null,
-    ] as [DateOrNull, DateOrNull];
+    ];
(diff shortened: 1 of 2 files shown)


src/components/ui/utils/DecentDatePicker.tsx
```diff
@@ -80,77 +87,206 @@ const isToday = (date: Date) => {
...
+  type: 'startDate' | 'endDate' | 'cliffDate';
+  formIndex: number;
+}) {
+  const [isDrawerOpen, setIsDrawerOpen] = useState(false);
+  const { values } = useFormikContext<RoleFormValues>();
+  const selectedDate = values.roleEditing?.payments?.[formIndex]?.[type];
   const boxShadow = useBreakpointValue({ base: 'none', md: SEXY_BOX_SHADOW_T_T });
   const maxBoxW = useBreakpointValue({ base: '100%', md: '26.875rem' });
+  return (
+    <Field name={`roleEditing.payments[${formIndex}].${type}`}>
```
DarksightKellar, line 105: I don't think it's a good idea to couple the datepicker and the roles feature. Maybe this could be a `formFieldName` param so the date picker is reusable in the future?

DarksightKellar (changes requested):

DarksightKellar (changes requested):
Noice. Overall I dig this refactor. Just one request on field name

src/components/ui/utils/DecentDatePicker.tsx
DarksightKellar: Turns out this whole thing was already uber coupled to roles >.<

If we need a date picker for something else in the future, we could rename this component to something like `RolePaymentDatePicker` and create another picker component that uses internally resuses `Calendar` from `react-calendar`. Styles are shared also so that should be fine.

DarksightKellar (approved):

adamgall (approved):

PR #2273, opened 2024-08-19
Fix streams creation for roles with existing payments, fix editing payments

### LINKED ISSUE #2261: BugFix: Adding payment to the role that has existing payment doesn't work (seems like problems with filtering created/edited/non-edited streams)

### LINKED ISSUE #2274: Editing payment does not create new payment stream. We're missing indication of whether payment was edited at all

mudrila:
Closes #2261 
Closes #2274 

Note - edit/delete is not as much working. I'll be working on that in the next PR. With that I'll also implement more thoughtful mapping


src/store/roles/rolesStoreUtils.ts
```diff
@@ -341,28 +337,6 @@ export async function getNewRole({
...
-    streamId: payment.streamId,
-    contractAddress: payment.contractAddress,
-    asset: payment.asset,
-    amount: payment.amount,
-    startDate: payment.startDate,
-    endDate: payment.endDate,
-    cliffDate: payment.cliffDate,
-  };
-};
-
```
adamgall, line 365: What was your reason for removing this? I quite like it.

src/hooks/utils/useCreateRoles.ts
```diff
@@ -523,10 +518,10 @@ export default function useCreateRoles() {
 
       if (addedNewPaymentsHats.length) {
         const streamsData = addedNewPaymentsHats.flatMap(role =>
-          (role.payments ?? []).map(payment => normalizePaymentFormData(payment)),
+          (role.payments ?? []).filter(payment => !payment.streamId),
```
adamgall, line 521: This filter already happens higher up. Why remove the `noramlizePaymentFormData()` call, that was to make typing better.

Da-Colon:
I don't really understand these changes? @mudrila

adamgall (commented):
@mudrila Can you please go through and leave PR comments on your changes in here?

src/hooks/utils/useCreateRoles.ts
```diff
@@ -577,9 +577,11 @@ export default function useCreateRoles() {
           }),
         );
         const streamsData = editedPayrollHats.flatMap(role =>
-          (role.payments ?? []).map(payment => normalizePaymentFormData(payment)),
+          (role.payments ?? []).filter(payment => !payment.streamId),
```
DarksightKellar, line 580: Apart from the comment above about `filter` already happening higher up (this this is redundant), `editedPayrollHats` is formed from a `.filter(payment => !!payment.streamId)`, so this filter would effectively always return a 0-length array, no?

src/store/roles/rolesStoreUtils.ts
DarksightKellar: Ditto

adamgall (approved):
These changes look fine to me. I'm still a bit unsure about the removal of `normalizePaymentFormData`, but that's mostly because of my own knowledge deficiencies.

I'm working on a bunch of type updates in another branch that David started, and I think there will be some conflicts with this. But it's all good.

adamgall:
Merging this early to keep things moving

PR #2278, opened 2024-08-19
Active payments count

adamgall:
Closes https://github.com/decentdao/decent-interface/issues/2254

So this code doesn't work.

All the typing checks out, but over in `RolePaymentDetails`, on the line where this happens:

```ts
<GreenStreamingDot isStreaming={payment.isStreaming()} />
```

at runtime `payment` doesn't have an `isStreaming()` function when a new payment is being added.

Trying to figure this out on my own right now, but @Da-Colon if you have any hints plz holler

### DIFF
--- a/src/components/pages/Roles/RolesTable.tsx
+++ b/src/components/pages/Roles/RolesTable.tsx
@@ -262,7 +262,11 @@ export function RolesTable({
                 name={role.name}
                 wearerAddress={role.wearer}
                 handleRoleClick={handleRoleClick}
-                paymentsCount={role.payments?.length}
+                paymentsCount={
+                  role.payments === undefined
+                    ? undefined
+                    : role.payments.filter(p => p.isStreaming()).length
+                }
               />
             ))}
           </Tbody>
--- a/src/components/pages/Roles/types.tsx
+++ b/src/components/pages/Roles/types.tsx
@@ -22,9 +22,12 @@ export interface SablierPayment extends BaseSablierStream {
   startDate: Date;
   endDate: Date;
   cliffDate: Date | undefined;
+  isStreaming: () => boolean;
 }
 
-export interface SablierPaymentFormValues extends Partial<SablierPayment> {}
+export interface SablierPaymentFormValues extends Partial<SablierPayment> {
+  isStreaming: () => boolean;
+}
 
 export type SablierPaymentOrPartial = SablierPayment | SablierPaymentFormValues;
 export interface RoleProps {
--- a/src/hooks/DAO/loaders/useHatsTree.ts
+++ b/src/hooks/DAO/loaders/useHatsTree.ts
@@ -176,6 +176,13 @@ const useHatsTree = () => {
                     BigInt(lockupLinearStream.depositAmount),
                     lockupLinearStream.asset.decimals,
                   );
+
+                  const startDate = secondsTimestampToDate(lockupLinearStream.startTime);
+                  const endDate = secondsTimestampToDate(lockupLinearStream.endTime);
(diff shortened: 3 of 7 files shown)


src/components/pages/Roles/RolesTable.tsx
```diff
@@ -262,7 +262,11 @@ export function RolesTable({
                 name={role.name}
                 wearerAddress={role.wearer}
                 handleRoleClick={handleRoleClick}
-                paymentsCount={role.payments?.length}
+                paymentsCount={
+                  role.payments === undefined
+                    ? undefined
+                    : role.payments.filter(p => p.isStreaming()).length
```
Da-Colon, line 268: Ah good catch filtering for length

src/components/pages/Roles/types.tsx
```diff
@@ -22,9 +22,12 @@ export interface SablierPayment extends BaseSablierStream {
   startDate: Date;
   endDate: Date;
   cliffDate: Date | undefined;
+  isStreaming: () => boolean;
 }
 
-export interface SablierPaymentFormValues extends Partial<SablierPayment> {}
+export interface SablierPaymentFormValues extends Partial<SablierPayment> {
+  isStreaming: () => boolean;
```
Da-Colon, line 29: This should be on SablierPayment not the FormValues @adamgall

adamgall: Ok, thank you, let me give that a shot

adamgall: edit: lol well it _is_ on the `SablierPayment`, too. I added it here in order to override it so it's not optional.

src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -193,11 +200,24 @@ const useHatsTree = () => {
...
+                          : cliffDate !== undefined
+                            ? cliffDate.getTime()
+                            : undefined;
+                      const end = endDate ? endDate.getTime() : undefined;
+                      const cancelled = lockupLinearStream.canceled;
+                      const now = new Date().getTime();
+                      const isStreaming =
+                        !cancelled && !!start && !!end && start <= now && end > now;
+                      return isStreaming;
+                    },
```
Da-Colon, line 220: At this point I don't believe start and end times will ever be undefined. so probably don't need those conditionals. 

isStreaming coming out correct? hmmm

src/components/pages/Roles/types.tsx
adamgall: i need halp tho

src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -193,11 +200,24 @@ const useHatsTree = () => {
...
+                    endDate,
+                    cliffDate,
+                    isStreaming: () => {
+                      const start =
+                        lockupLinearStream.cliff === undefined &&
+                        lockupLinearStream.startTime !== undefined
+                          ? startDate.getTime()
+                          : cliffDate !== undefined
+                            ? cliffDate.getTime()
+                            : undefined;
```
DarksightKellar, line 213: Maybe `: cliffDate?.getTime();` for `: cliffDate.getTime() : undefined;`

src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -193,11 +200,24 @@ const useHatsTree = () => {
...
+                      const end = endDate ? endDate.getTime() : undefined;
```
DarksightKellar, line 214: `const end = endDate?.getTime()`?

PR #2291, opened 2024-08-21
Cleanup

adamgall:
Just some light cleanup extracted to own PR as I'm working on other stuff.


DarksightKellar (approved):

package.json
```diff
@@ -120,7 +120,7 @@
     "jsdom": "^24.0.0",
     "netlify-cli": "^17.33.3",
     "prettier": "^3.2.5",
-    "typescript": "^5.3.3",
+    "typescript": "^5.5.4",
```
adamgall, line 123: This update allows us to do things like

```ts
const foo = collection.filter(item => item !== null)`
```

and `foo` will be properly typed to know that it's _not_  `null`.

PR #2295, opened 2024-08-21
Load hats tree after tree id

DarksightKellar:
This PR tightens up hats tree fetching logic so that a new hats tree load is only attempted _after_ a new hats tree ID has been loaded when the chain ID (and/or safe address) changes.


adamgall (commented):
@DarksightKellar would you mind please adding more context to this PR, describing what prompted you to make these changes and the result of them?

adamgall:
@DarksightKellar @Da-Colon perhaps [this issue](https://github.com/decentdao/decent-interface/issues/2298) will be fixed from this work?

adamgall (approved):

PR #2312, opened 2024-08-23
Handle malformed metadata

DarksightKellar:
Closes #2069 

"If proposal metadata is malformed for Azorius - we completely skipping further proposal parsing and thus it's not possible to execute proposal. We should move parsing of the proposal's metadata down the line so that we can still show & execute the proposal"

Re: "move parsing down", that didn't happen in this PR, nor did it seem necessary.

A simple try-catch on just the metadata parsing and returning a default JSON on failure did the trick.

### DIFF
--- a/src/hooks/DAO/loaders/governance/useAzoriusProposals.ts
+++ b/src/hooks/DAO/loaders/governance/useAzoriusProposals.ts
@@ -6,6 +6,7 @@ import {
 import { VotedEvent as ERC20VotedEvent } from '@fractal-framework/fractal-contracts/dist/typechain-types/contracts/azorius/LinearERC20Voting';
 import { VotedEvent as ERC721VotedEvent } from '@fractal-framework/fractal-contracts/dist/typechain-types/contracts/azorius/LinearERC721Voting';
 import { useCallback, useEffect, useMemo, useRef } from 'react';
+import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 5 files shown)


src/hooks/DAO/loaders/governance/useAzoriusProposals.ts
```diff
@@ -201,27 +215,25 @@ export const useAzoriusProposals = () => {
...
         if (cachedProposal) {
-          completeProposalLoad(cachedProposal);
+          completeOneProposalLoadProcess(cachedProposal);
           continue;
         }
 
         let proposalData;
 
         if (proposalCreatedEvent.args.metadata) {
           try {
```
adamgall, line 225: do we still need and want _this_ `try/catch`, considering we moved the JSON parsing `try/catch` into a new function?

DarksightKellar: Yep, for `decodeTransactions`

adamgall: Instead of having this `try` declaration nest the one in `parseProposalMetadata`, can this be moved down to more tightly wrap just the `decodeTransactions` call?

DarksightKellar: Yeah good point. Done.

adamgall (approved):

PR #2314, opened 2024-08-27
Add Hotjar for them heatmaps + Amplitude for event analysis

DarksightKellar:
This PR will enable Hotjar tracking on the websites that will have the `VITE_APP_HOTJAR_SITE_ID` env variable on it (currently decision is to keep this to prod only).

Caveat: _"Hotjar is currently measuring your site's traffic. This requires 7 days worth of data."_ 

We'll need to wait at least a week after merging to prod and setting up the dashboard for any data to be visible.

UPDATE: This PR is being updated to also include Amplitude analytics.

### DIFF
--- a/.env
+++ b/.env
@@ -15,8 +15,10 @@ SENTRY_AUTH_TOKEN=""
 # App name displayed as Title
 VITE_APP_NAME="Decent"
 
-# Hotjar Site ID
+# Hotjar Site ID. Type is `number`
 VITE_APP_HOTJAR_SITE_ID=""
+
+# Hotjar Version. Should be retrieved from the Hotjar dashboard. Type is `number`
 VITE_APP_HOTJAR_VERSION=""
 
 # Alchemy provider API key
--- a/.env
+++ b/.env
@@ -21,6 +21,9 @@ VITE_APP_HOTJAR_SITE_ID=""
 # Hotjar Version. Should be retrieved from the Hotjar dashboard. This should be parseable to an integer.
 VITE_APP_HOTJAR_VERSION=""
 
+# API key for Amplitude analytics
+VITE_APP_AMPLITUDE_API_KEY=""
+
 # Alchemy provider API key
 VITE_APP_ALCHEMY_API_KEY=""
 
--- a/.env
+++ b/.env
@@ -15,10 +15,10 @@ SENTRY_AUTH_TOKEN=""
 # App name displayed as Title
 VITE_APP_NAME="Decent"
 
-# Hotjar Site ID. Type is `number`
+# Hotjar Site ID. This should be parseable to an integer.
 VITE_APP_HOTJAR_SITE_ID=""
 
-# Hotjar Version. Should be retrieved from the Hotjar dashboard. Type is `number`
+# Hotjar Version. Should be retrieved from the Hotjar dashboard. This should be parseable to an integer.
 VITE_APP_HOTJAR_VERSION=""
 
 # Alchemy provider API key
--- a/.env
+++ b/.env
@@ -15,6 +15,10 @@ SENTRY_AUTH_TOKEN=""
 # App name displayed as Title
 VITE_APP_NAME="Decent"
 
+# Hotjar Site ID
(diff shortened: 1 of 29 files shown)


.env
```diff
@@ -15,6 +15,10 @@ SENTRY_AUTH_TOKEN=""
 # App name displayed as Title
 VITE_APP_NAME="Decent"
 
+# Hotjar Site ID
+VITE_APP_HOTJAR_SITE_ID=""
+VITE_APP_HOTJAR_VERSION=""
+
```
adamgall, line 21: Might be good to indicate (comment) that these environment variables are `numbers`.

src/main.tsx
```diff
@@ -1,3 +1,5 @@
+import './insights/sentry';
+import './insights/hotjar';
```
adamgall, line 2: let's move these into an `index` file and just `import ./insights`

src/vite-env.d.ts
```diff
@@ -19,6 +19,9 @@ interface ImportMetaEnv {
 
   readonly VITE_APP_SENTRY_DSN_URL: string;
 
+  readonly VITE_APP_HOTJAR_SITE_ID: number;
+  readonly VITE_APP_HOTJAR_VERSION: number;
```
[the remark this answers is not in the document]
DarksightKellar: Discussed and resolved -- type will remain string due to the way .env parsing works

src/providers/App/useReadOnlyValues.ts
```diff
@@ -61,6 +62,9 @@ export const useReadOnlyValues = ({ node, governance }: Fractal, _account?: stri
 
     const address = _account ? getAddress(_account) : undefined;
     Sentry.setUser(address ? { id: address } : null);
+    if (address) {
+      amplitude.setUserId(address);
```
adamgall, line 66: where do you unset the user id?

DarksightKellar: You don't? This just IDs the "amplitude user" whose session is being tracked. Otherwise it gets a generic anon user id

adamgall: yeah i know, i'm just thinking of the case where a user connects their wallet, then disconnects their wallet.

i'd expect that we "start" an amplitude session when a user connects their wallet, and "end" and amplitude session when they disconnect their wallet.

DarksightKellar: Ah flip, yeah lemme look into that

adamgall: i mean it could be as simple as removing the `if (address)` check. similar logic is happening with the `Sentry.setUser()`

DarksightKellar: Thanks this comment helped as I was looking for a good place to do this. See updated code plz @adamgall

adamgall (approved):

adamgall (approved):

mudrila (approved):
Have a question but assuming I'm wrong - should be good to go

src/pages/daos/[daoAddress]/SettingsPage.tsx
```diff
@@ -1,12 +1,16 @@
...
 import { useTranslation } from 'react-i18next';
 import { DAOSettingsContent } from '../../../components/pages/DaoSettings/DAOSettingsContent';
 import { BarLoader } from '../../../components/ui/loaders/BarLoader';
 import PageHeader from '../../../components/ui/page/Header/PageHeader';
 import { useHeaderHeight } from '../../../constants/common';
+import { analyticsEvents } from '../../../insights/analyticsEvents';
 import { useFractal } from '../../../providers/App/AppProvider';
 
 export function SettingsPage() {
+  amplitude.track(analyticsEvents.SafeSettingsPageOpened);
```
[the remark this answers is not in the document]
Comment (adamgall) [reply src/pages/daos/[daoAddress]/SettingsPage.tsx:12]: o yeah, this is a good callout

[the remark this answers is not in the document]
Comment (DarksightKellar) [reply src/pages/daos/[daoAddress]/SettingsPage.tsx]: Man am I glad you caught this

PR #2316, opened 2024-08-27
Center TopErrorFallback in da middle of the screen

mudrila:
I'm quite sure this was the original intention and was implemented this way - but seems like we've broke it at some point a while ago cause I've seen this shitty error display but for who knows which reason thought it's intentional. Now I checked Figma - and there you go. Center as it should be. I'm not mad or something 🤣 
https://www.figma.com/design/9hUYtvCKjOyB92vt8ukYtD/Product-Handoff?node-id=2599-8113&m=dev

Before:
Screenshot 2024-08-27 at 23 42 50
Screenshot 2024-08-27 at 23 43 09

After:
Screenshot 2024-08-27 at 23 37 47
Screenshot 2024-08-27 at 23 38 02

@decentdao/engineering though I don't really like the way I implemented this - I wasn't able to come up with meaningful implementation without touching too much of surrounding layout. I'm actually OK keeping this implementation long term, but if someone feels we should implemented it better - feel free to create an issue and we'll tackle it (or plz halp to do that now 🤣 )

### DIFF
--- a/src/components/ui/utils/TopErrorFallback.tsx
+++ b/src/components/ui/utils/TopErrorFallback.tsx
@@ -1,16 +1,20 @@
-import { Button, Box, Flex, Text, Image, Hide, Show } from '@chakra-ui/react';
+import { Box, Button, Flex, Hide, Image, Show, Text } from '@chakra-ui/react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


adamgall (approved):
I'm ok with this implementation as-is for now. Thanks for picking it up.

nicolaus-sherrill (approved):
perfectly-balanced-thanos

PR #2317, opened 2024-08-27
Typing the Push -> The Withdraw Modal+

Da-Colon:
As I was looking at the `RolePaymentDetails` component noticed that the typing for the modal props were a `any`, now they are typed. 

This also updates a few other places where variables could have been passed to the Modal props as undefined.

### DIFF
--- a/src/components/ui/modals/useDecentModal.tsx
+++ b/src/components/ui/modals/useDecentModal.tsx
@@ -1,5 +1,11 @@
 import { useContext } from 'react';
-import { IModalContext, ModalContext, ModalType } from './ModalProvider';
+import {
+  CurrentModal,
+  IModalContext,
+  ModalContext,
+  ModalPropsTypes,
+  ModalType,
+} from './ModalProvider';
 
 /**
  * Returns a Function intended to be used in a click listener to open the provided ModalType.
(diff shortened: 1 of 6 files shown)


src/components/ui/modals/useDecentModal.tsx
```diff
@@ -8,9 +14,11 @@ import { IModalContext, ModalContext, ModalType } from './ModalProvider';
  * @param props optional arbitrary key:value properties to pass to the modal
  * @returns a Function that when called opens the provided ModalType modal.
  */
-export const useDecentModal = (modal: ModalType, props?: Record<string, any>) => {
+export const useDecentModal = <T extends ModalType>(modal: T, props?: ModalPropsTypes[T]) => {
   const { setCurrent } = useContext<IModalContext>(ModalContext);
   return () => {
-    setCurrent({ type: modal, props: props ? props : [] });
+    const modalObject = { type: modal, props: props ?? {} } as CurrentModal;
```
Da-Colon, line 20: I REALLY tried to avoid this `as` but alas I failed. The problem lies with `type` property. Whether `modal` is typed as `ModalType` or the new update it shows an error because of 

=== 
     Types of property 'type' are incompatible.
        Type 'ModalType' is not assignable to type 'ModalType.NONE'.ts(2322)
 ===
 
 Something about how the `union` works.

adamgall (approved):

PR #2321, opened 2024-08-29
`Bugfix` | EditStreams

Da-Colon:
Found the issues with the Edit Stream workflow. Looks like code got duplicated in a merge fix and the wrong address was being passed into the `encodeFunctionData` function for flushing the stream.


adamgall (approved):
This code is good to go from my end. There is another branch coming soon which rewrites a bunch of this anyway.

Da-Colon (commented):
ship it

PR #2332, opened 2024-09-04
Typechain quick fixes

DarksightKellar:
A couple o' quick fixes that I could implement on the fly while testing the typechain branch.


src/hooks/schemas/common/useValidationAddress.tsx
```diff
@@ -143,12 +142,22 @@ export const useValidationAddress = () => {
...
+          resolvedAddress = await publicClient.getEnsAddress({
+            name: normalize(addressOrENS),
+          });
+
+          if (!resolvedAddress) return false;
+          return !safe.owners.includes(resolvedAddress);
+        }
+
+        resolvedAddress = getAddress(addressOrENS);
+        return !safe.owners.includes(resolvedAddress);
```
DarksightKellar, line 160: Previous logic was causing all addresses to be rejected even if it was valid and not already a signer. 
`const check = await publicClient.getEnsAddress({ name: normalize(address) });` was failing with already "resolved" address strings (strings that were NOT ENS names but valid addresses), causing the whole test to return false.

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
adamgall, line 1: Do the changes in this file make any functional changes?

DarksightKellar: Nope, except the bug fix insrc/hooks/schemas/common/useValidationAddress.tsx`

adamgall (approved):

PR #2333, opened 2024-09-05
Process roles serially

Da-Colon:


src/hooks/utils/useCreateRoles.ts
```diff
@@ -460,331 +233,511 @@ export default function useCreateRoles() {
...
+          if (streamsToCancel.length || fundsToClaimStreams.length) {
+            allTxs.push({
+              calldata: encodeFunctionData({
+                abi: HatsAbi,
+                functionName: 'transferHat',
+                args: [BigInt(formHat.id), daoAddress, getAddress(formHat.wearer)],
+              }),
+              targetAddress: hatsProtocol,
+            });
+          }
```
adamgall, line 447: Do we need to transfer a "deleted" Hat back to the original wearer? We've already flushed it and cancelled the stream, so I don't think this is necessary.

But if it is necessary, then my next question is, why is the logic to do that transfer wrapped in this conditional that only does it if there are "streams to cancel" or "streams with funds to claim"?

Da-Colon: Is it necessary, good question. I guess right now only in there to 'leave things how it was' since deleting doesn't 'remove' wearer. But kinda makes sense that its wearer would be the DAO either way.

well currently the transfer to the DAO only happens if one of those two happens. but is mute as I think we should just transfer to the DAO and remove this other logic

adamgall: yeah agreed

src/hooks/utils/useCreateRoles.ts
```diff
@@ -460,331 +233,511 @@ export default function useCreateRoles() {
...
+              newStreams,
+              newPredictedHatSmartAccount,
+            );
+            allTxs.push(...newStreamTxData.preparedTokenApprovalsTransactions);
+            allTxs.push(...newStreamTxData.preparedStreamCreationTransactions);
+          }
+        } else if (formHat.editedRole.status === EditBadgeStatus.Removed) {
+          /**
+           * Removed Role
+           * Transfer hat to DAO, flush streams, cancel streams, transfer hat to back to wearer, set hat status to false
```
adamgall, line 398: This comment should be updated due to latest commit, right?

adamgall: I'm pushing a commit about it now

adamgall (approved):
LGTM

src/hooks/utils/useCreateRoles.ts
```diff
@@ -345,6 +70,50 @@ export default function useCreateRoles() {
...
+      }
+
+      return createHatStruct(role.name, role.description, getAddress(role.wearer));
+    },
+    [createHatStruct],
+  );
+
+  const createHatStructsFromRolesFormValues = useCallback(
+    async (modifiedRoles: RoleHatFormValueEdited[]) => {
+      return Promise.all(modifiedRoles.map(role => createHatStructFromRoleFormValues(role)));
```
DarksightKellar, line 112: Nitpick and def shouldn't block merge but thinking we can do `Promise.all(modifiedRoles.map(createHatStructFromRoleFormValues));` here

src/hooks/utils/useCreateRoles.ts
```diff
@@ -810,78 +818,45 @@ export default function useCreateRoles() {
...
-            uploadHatDescription,
-            getHat,
-          );
-          const editedPaymentStreams = identifyAndPrepareEditedPaymentStreams(
-            modifiedHats,
-            getHat,
-            getPayment,
-          );
+          if (modifiedHats.some(hat => hat.editedRole.status !== EditBadgeStatus.New)) {
+            throw new Error(
```
DarksightKellar, line 828: Nice

src/hooks/utils/useCreateRoles.ts
```diff
@@ -810,78 +818,45 @@ export default function useCreateRoles() {
...
-            memberChangedHats,
-            roleDetailsChangedHats,
-            editedPaymentStreams,
-            addedPaymentStreamsOnNewHats,
+            modifiedHats,
           );
         }
 
-        // All done, submit the proposal!
+        // // All done, submit the proposal!
```
DarksightKellar, line 848: 👀 
`// //` ?

DarksightKellar (approved):
LGTM! 

I didn't spend a lot of time scrutinising the EDIT code paths because
 - @adamgall spent quite a bit of time with me taking me thru the algorithm and implementation
 - We're removing editing in favour of cancelling anyways

PR #2344, opened 2024-09-11
Fix PAX price and treasury transactions

adamgall:
Two things in here:

1. Moralis API doesn't return good price data for the PAX token. We will just hardcode this to $1.
2. While implementing the PAX price override, I noticed that no treasury transactions were loading (on local, dev, or prod). Seems like the Safe API started returning `null` instead of `undefined` for certain datapoints, which was causing this error in our app. Fixed that.

### DIFF
--- a/netlify/functions/tokenBalances.mts
+++ b/netlify/functions/tokenBalances.mts
@@ -14,16 +14,24 @@ export default async function getTokenBalancesWithPrices(request: Request) {
 
   const fetchFromMoralis = async (scope: { chain: string; address: Address }) => {
     const tokensResponse = await Moralis.EvmApi.wallets.getWalletTokenBalancesPrice(scope);
-
     const mappedTokensData = tokensResponse.result
       .filter(tokenBalance => tokenBalance.balance.value.toBigInt() > 0n)
-      .map(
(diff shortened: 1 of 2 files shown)


netlify/functions/tokenBalances.mts
```diff
@@ -14,16 +14,24 @@ export default async function getTokenBalancesWithPrices(request: Request) {
...
   const fetchFromMoralis = async (scope: { chain: string; address: Address }) => {
     const tokensResponse = await Moralis.EvmApi.wallets.getWalletTokenBalancesPrice(scope);
-
     const mappedTokensData = tokensResponse.result
       .filter(tokenBalance => tokenBalance.balance.value.toBigInt() > 0n)
-      .map(
-        tokenBalance =>
-          ({
-            ...camelCaseKeys(tokenBalance.toJSON()),
-            decimals: Number(tokenBalance.decimals),
```
adamgall, line 24: I'm not sure why we were doing this line. `tokenBalance.decimals` is already a `number` so no need to cast it.

Da-Colon (approved):

DarksightKellar (approved):

PR #2356, opened 2024-09-18
Freeze updates

DarksightKellar:
Mostly bug fixes and some light refactoring from testing freeze and clawback functionality on the remove typechain branch


src/i18n/locales/en/common.json
```diff
@@ -45,7 +45,8 @@
   "errorInvalidERC20Address": "Invalid ERC-20 token address",
   "errorInvalidERC721Address": "Invalid ERC-721 token address",
   "errorInvalidBigint": "Invalid number value",
-  "errorMinimumValue": "This value can't be less then {{minValue}}",
+  "errorMinimumValue": "This value can't be less than {{minValue}}",
+  "errorMaximumValue": "This value can't be more than {{maxValue}}",
```
adamgall, line 49: Nitpick: "more" -> "greater"

src/components/DaoCreator/formComponents/GuardDetails.tsx
```diff
@@ -37,7 +33,7 @@ function GuardDetails(props: ICreationStepProps) {
 
   const handleNonceChange = useCallback(
     (nonce?: string) => {
-      setFieldValue('multisig.customNonce', nonce);
+      setFieldValue('multisig.customNonce', Number(nonce));
```
adamgall, line 36: Can you explain this change?

I don't see anywhere in this PR where the type of any `customNonce` property has changed from `string` to `number`.

src/hooks/schemas/DAOCreate/useDAOCreateSchema.ts
```diff
@@ -10,7 +10,14 @@ import { useDAOCreateTests } from './useDAOCreateTests';
...
  * @dev https://www.npmjs.com/package/yup
  */
-export const useDAOCreateSchema = ({ isSubDAO }: { isSubDAO?: boolean }) => {
+export const useDAOCreateSchema = ({
+  isSubDAO,
+  parentNumSigners,
+}: {
+  isSubDAO?: boolean;
+  // @todo: `parentNumSigners` should probably be a bigint
+  parentNumSigners?: number;
```
adamgall, line 19: It's fine that this is `undefined` for Azorius safes, right?

src/components/DaoCreator/index.tsx
```diff
@@ -25,7 +26,15 @@ function DaoCreator({
...
   mode: DAOCreateMode;
 }) {
-  const { createDAOValidation } = useDAOCreateSchema({ isSubDAO });
+  const {
+    node: { safe },
+  } = useFractal();
+
+  const { createDAOValidation } = useDAOCreateSchema({
+    isSubDAO,
+    parentNumSigners: safe?.owners.length,
```
adamgall, line 35: This will be `undefined` for Azorius safes, right?

src/components/Proposals/ProposalActions/CastVote.tsx
```diff
@@ -23,6 +23,10 @@ export function CastVote({ proposal }: { proposal: FractalProposal }) {
   const { snapshotProposal, extendedSnapshotProposal, loadSnapshotProposal } =
     useSnapshotProposal(proposal);
 
+  useEffect(() => {
+    loadSnapshotProposal();
+  }, [loadSnapshotProposal]);
```
adamgall, line 28: What was the scenario that required this to be added?

src/components/Proposals/SnapshotProposalDetails/SnapshotProposalVotes.tsx
```diff
@@ -53,12 +53,15 @@ export default function SnapshotProposalVotes({ proposal }: ISnapshotProposalVot
             {choices.map((choice, i) => {
               const votesBreakdownChoice =
                 type === 'weighted' ? votesBreakdown[i + 1] : votesBreakdown[choice];
+
               const votesBreakdownChoiceTotal =
                 votesBreakdownChoice && votesBreakdownChoice?.total
```
adamgall, line 58: This conditional is kind of ambiguous.

The type of `votesBreakdownChoice` is:
```ts
const votesBreakdownChoice: {
    total: number;
    votes: SnapshotVote[];
}
```

, so why checking for it's existence? It's an object which will never be falsey. (I know you didn't change this line, but since we're here...

src/components/Proposals/SnapshotProposalDetails/SnapshotProposalVotes.tsx
```diff
@@ -53,12 +53,15 @@ export default function SnapshotProposalVotes({ proposal }: ISnapshotProposalVot
...
                   ? votesBreakdownChoice?.total
                   : 0;
-              const choicePercentageFromTotal =
-                (votesBreakdownChoiceTotal * 100) / totalVotesCasted;
+
+              const choicePercentageFromTotal = totalVotesCasted
```
adamgall, line 62: This conditional is kind of ambiguous. `totalVotesCasted` is a `number`, so are you checking to see if it's equal to 0 here? I'd prefer to be explicit if that's the case: `totalVotesCasted !== 0 ? ...`

src/hooks/DAO/useClawBack.ts
```diff
@@ -123,7 +123,7 @@ export default function useClawBack({ childSafeInfo, parentAddress }: IUseClawBa
               pendingToastMessage: t('clawBackPendingToastMessage'),
               failedToastMessage: t('clawBackFailedToastMessage'),
               successToastMessage: t('clawBackSuccessToastMessage'),
-              safeAddress: parentAddress,
+              safeAddress: getAddress(parentAddress),
```
adamgall, line 126: This `getAddress` seems unneeded.

src/hooks/schemas/DAOCreate/useDAOCreateSchema.ts
```diff
@@ -143,32 +150,59 @@ export const useDAOCreateSchema = ({ isSubDAO }: { isSubDAO?: boolean }) => {
...
+                          value === undefined ||
+                          (parentNumSigners !== undefined && value <= parentNumSigners)
+                        ) {
+                          return true;
+                        }
+                        return false;
+                      },
+                    }),
+                })
+                .required(),
```
adamgall, line 184: Ah I didn't realize you wrote this validator. Nice.

src/pages/daos/[daoAddress]/new/index.tsx
```diff
@@ -27,16 +27,23 @@ export default function SubDaoCreate() {
...
+    daoData: SafeMultisigDAO | AzoriusGovernanceDAO | SubDAO,
+    customNonce?: number,
+  ) => {
     const subDAOData = daoData as SubDAO;
-    proposeDao(subDAOData, subDAOData.customNonce || safe?.nextNonce, successCallback);
+    proposeDao(
+      subDAOData,
+      subDAOData.customNonce || customNonce || safe?.nextNonce,
+      successCallback,
+    );
```
Comment (adamgall) [inline src/pages/daos/[daoAddress]/new/index.tsx:39]: Gah can you explain this change? Why so many places that the "custom nonce" could come from?

src/providers/App/hooks/useSafeAPI.ts
adamgall, line 1: I believe I remember why these changes are here... the Safe API requires that addresses passed into it are fully checksummed, right?

However, my preference would be that rather than checksumming this address _right before it's required to do so_, the address was just plain ol _already_ checksummed.

Meaning, where did this address come from? Loaded in from some data provider (blockchain, moralis, subgraph, idk, etc) originally, probably. Can we get this address into its proper checksummed form as it comes into the app?

Which would render these changes unnecessary.

adamgall (changes requested):

src/components/DaoCreator/formComponents/GuardDetails.tsx
DarksightKellar: Ah yes this is good. `SafeConfiguration.customNonce` in `createDAO.ts` is a `number`. Doing `setFieldValue('multisig.customNonce', nonceWhichIsAString)` skips that typing and _makes_ it a string, which is obvious when inspecting or logging `multisig.customNonce` after setting.

I don't remember if is was in this instance, but this led to a validation edge case bug during my tests, I think when comparing the set form value with another number. I don't remember exactly what, but it _was_ a problem.

Because of this callout I've realised there are plenty of `setFieldValue('multisig.customNonce', nonce)`s around the codebase. Gonna update them to set a number instead.

src/components/Proposals/ProposalActions/CastVote.tsx
DarksightKellar: `extendedSnapshotProposal` was undefined until this is called. While it's called in a parent component (`SnapshotProposalDetails`), the set value wasn't reflected here for some reason.

src/components/Proposals/SnapshotProposalDetails/SnapshotProposalVotes.tsx
DarksightKellar: Yeahhh good catch here. The whole thing seems to be unnecessary, and we should be able to just do `const votesBreakdownChoiceTotal = votesBreakdownChoice.total`, or just `votesBreakdownChoice.total` directly

src/components/Proposals/SnapshotProposalDetails/SnapshotProposalVotes.tsx
DarksightKellar: Updated to `totalVotesCasted > 0` before I saw this comment. Which should we prefer?

src/hooks/DAO/useClawBack.ts
DarksightKellar: Alas, it is.

src/pages/daos/[daoAddress]/new/index.tsx
Comment (DarksightKellar) [reply src/pages/daos/[daoAddress]/new/index.tsx:39]: > Why so many places that the "custom nonce" could come from?
That is a good question, one which I did try to figure out but didn't get much clarity into.

But, for this change, it was needed for creating a subsafe on a multisig safe whose `nextNonce` hasn't yet updated. In that case, setting a custom nonce was not being reflected in the built tx.

src/hooks/schemas/DAOCreate/useDAOCreateSchema.ts
DarksightKellar: Sigh, dammit I forgot to test the Azorius path. that'll decide what I do with this arg. So I don't have an answer now unfortunately. Lemme circle back

src/providers/App/hooks/useSafeAPI.ts
DarksightKellar: Hmm that's interesting because my line of reasoning was from the other end -- since these Safe API calls are the only things requiring a checksummed address, we co-locate that (seemingly redundant) transformation with the call that requires it, instead of hunting down the data sources. 

I do see the pros to your point though...

src/components/DaoCreator/index.tsx
adamgall: Same thing as previous comment, closing.

src/hooks/schemas/DAOCreate/useDAOCreateSchema.ts
adamgall: @DarksightKellar I see new commits. What's the update on this?

src/providers/App/hooks/useSafeAPI.ts
adamgall: @DarksightKellar I see new commits. What's the update on this?

PR #2357, opened 2024-09-19
Fix random and blocking npm install error

DarksightKellar:
Out of the blue, randomly, I started being unable to run `npm i`. 

This is what fixed it eventually, after having a chat with good ol' chatGPT. Submitting as a PR to `develop` instead of pushing to the typechain branch because:

- I want this tested and confirmed to not break anything on anyone else's machine
- I want to confirm proper behaviour on netlify builds
- The `npm i` error was present for me no matter which branch I switched to, including `develop`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,7 +69,7 @@
         "viem": "^2.13.1",
         "vite": "^5.1.0",
         "vite-plugin-checker": "^0.6.4",
-        "wagmi": "^2.9.7",
+        "wagmi": "^2.12.12",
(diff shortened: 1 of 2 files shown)


adamgall:
A whole PR when you could just do

```sh
rm -rf node_modules
nvm use
npm i
```

DarksightKellar:
> A whole PR when you could just do
> 
> ```shell
> rm -rf node_modules
> nvm use
> npm i
> ```

I tried though, it didn't work out T_T

PR #2371, opened 2024-09-24
Fix treasury page tx list inconsistencies when switching safes

### LINKED ISSUE #2370: Treasury page sometimes includes txs from previous safe when navigating to a new safe
Steps to see the problem:

- Check out [this safe's treasury page](https://app.dev.decentdao.org/treasury?dao=sep:0xa042F2b97855f72c4714B9E5ff9d603D2EE7F078)
- Navigate to the Organization tab and click into the treasury page of any of its sub safes. See that there are a similar number of txs. A refresh reveals the more accurate, much fewer number of tx -- around a couple.

Unfortunately there appears to be a race condition on safe loads, so this bug is fairly inconsistent to show. 
See: https://github.com/decentdao/decent-interface/pull/2371

DarksightKellar:
Closes #2370 

There was a race condition that was causing previous safe data to be loaded over a recently-navigated-to-safe's data. `useUpdateSafeData` and `useFractalNode.setDAO` were both loading safe data at around the same time: when the url changes and when the user changes safes respectively.

This PR fixes that by having `useUpdateSafeData` use `daoAddress` from url params instead of `node` in state, which could still hold the old safe's value before it is updated after the new safe is loaded in.


src/hooks/DAO/loaders/useDecentTreasury.ts
```diff
@@ -225,10 +226,11 @@ export const useDecentTreasury = () => {
 
     const newLoadKey = `${chain.id}${daoAddress}`;
     if (newLoadKey !== loadKey.current) {
+      // action.dispatch({ type: TreasuryAction.RESET });
```
Da-Colon, line 229: commented code.

src/hooks/utils/useUpdateSafeData.ts
```diff
@@ -1,13 +1,14 @@
...
 import { useFractal } from '../../providers/App/AppProvider';
 import { useSafeAPI } from '../../providers/App/hooks/useSafeAPI';
 import { NodeAction } from '../../providers/App/node/action';
 
-export const useUpdateSafeData = () => {
+export const useUpdateSafeData = (daoAddress?: Address) => {
   const {
     action,
-    node: { daoAddress },
+    // node: { daoAddress },
```
Da-Colon, line 11: commented code.

Da-Colon (approved):
Code lgtm. I commented on some commented code. Can these be removed?

src/hooks/DAO/loaders/useDecentTreasury.ts
adamgall: ya agree plz remove

src/hooks/DAO/loaders/useFractalNode.ts
```diff
@@ -134,9 +134,10 @@ export const useFractalNode = (
       skip ||
       addressPrefix === undefined ||
       daoAddress === undefined ||
-      addressPrefix + daoAddress !== currentValidSafe.current
+      `${addressPrefix}${daoAddress}`.toLowerCase() !== currentValidSafe.current?.toLowerCase()
```
adamgall, line 137: @DarksightKellar did you find that this change is needed? If not, plz undo it.

src/hooks/DAO/useDAOController.ts
```diff
@@ -21,7 +21,8 @@ export default function useDAOController() {
 
   const prefixAndAddress = addressWithPrefix?.split(':');
   const addressPrefix = prefixAndAddress?.[0];
-  const daoAddress = prefixAndAddress?.[1];
+  const daoAddressStr = prefixAndAddress?.[1];
+  const daoAddress = daoAddressStr ? getAddress(daoAddressStr) : undefined;
```
adamgall, line 25: Does this crash the app (in an unexpected way) if the user manually modifies the address in the url to be malformed? I think that before this change, the app recognized that and gave an appropriate error message. Does that still hold true?

src/hooks/utils/useUpdateSafeData.ts
adamgall: ya agree plz remove

src/hooks/DAO/loaders/useDecentTreasury.ts
adamgall: ya agree plz remove

src/hooks/DAO/useDAOController.ts
adamgall: just checked, no it doesn't. plz fix.

adamgall (changes requested):

src/hooks/DAO/useDAOController.ts
DarksightKellar: Should be fixed now ser, plz re-test

adamgall (approved):
nice

PR #2373, opened 2024-09-27
Fix BigIntInput

DarksightKellar:
BigIntInput was broken after (EDIT: a fix from) the typechain merge -- no input was getting registered. This PR fixes that.

### DIFF
--- a/src/components/ui/forms/BigIntInput.tsx
+++ b/src/components/ui/forms/BigIntInput.tsx
@@ -149,7 +149,7 @@ export function BigIntInput({
   useEffect(() => {
     if (!inputValue || inputValue === currentValue?.value) return;
     if (currentValue?.bigintValue === 0n) setInputValue('');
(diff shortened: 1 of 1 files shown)


adamgall:
Hmm, IIRC this component is used often... I'm surprised it's been broken all throughout the `remove-typechain` branch. Is that what you're saying? Did all of your manual testing not run into this bug?

DarksightKellar:
> Hmm, IIRC this component is used often... I'm surprised it's been broken all throughout the `remove-typechain` branch. Is that what you're saying? Did all of your manual testing not run into this bug?

@adamgall Oh no. It was a fix in the typechain branch that led to this bug. Remember the initial freeze threshold bug? The fix to that one led to this, because we didn't test usages of BigIntInput elsewhere after that update

adamgall (approved):

Da-Colon (approved):

PR #2374, opened 2024-09-27
Date range update

### LINKED ISSUE #2315: Strange behavior on date picker component.
Strange behavior on this date picker component. The start and end date inputs do not share the same component, rather they use two different date pickers, so I can select and override two different date ranges.

Additionally, the input field does not respond when I select a date and remains blank.

Lastly, I know we designed the date input fields in the date picker itself to be manually typed, but they aren't currently interactive and due to the positioning of the component the two input fields look very redundant. I think we should just remove those date fields from the date picker component.

recording

<details><summary>Browser metadata</summary>

```
Path:      /roles/edit/details?dao=sep:0xe20221A8939F9373C091262CBEF9493aca196acD&hatId=0xaac26be5e48b1859b982553e0ed9da801827d128cddd7f96627e8cdcb60eba2f
Browser:   Chrome 127.0.0.0 on Mac OS 10.15.7
Viewport:  1425 x 904 @2x
Language:  en-US
Cookies:   Enabled
```

[Open in BrowserStack](https://www.browserstack.com/user/try-live?url=https%3A%2F%2Fdeploy-preview-2309.app.dev.decentdao.org%2Froles%2Fedit%2Fdetails%3Fdao%3Dsep%3A0xe20221A8939F9373C091262CBEF9493aca196acD%26hatId%3D0xaac26be5e48b1859b982553e0ed9da801827d128cddd7f96627e8cdcb60eba2f&os=OS%20X&os_version=Catalina&browser=Chrome&browser_version=127.0&resolution=1425x904&speed=1&start=true&ref=netlify-source)

</details>

[Open Deploy Preview](https://deploy-preview-2309.app.dev.decentdao.org/roles/edit/details?dao=sep:0xe20221A8939F9373C091262CBEF9493aca196acD&hatId=0xaac26be5e48b1859b982553e0ed9da801827d128cddd7f96627e8cdcb60eba2f) · [Mark as Resolved](https://app.netlify.com/cdp/resolve?deployID=66ccb3f9866789000868b25e&commentID=66cdf6f4d7595335f5a369cc&resolution=resolved)

_Originally posted by @nicolaus-sherrill in https://github.com/decentdao/decent-interface/issues/2309#issuecomment-2312945908_

DarksightKellar:
Closes #2315

### DIFF
--- a/src/components/pages/Roles/forms/RoleFormPaymentStream.tsx
+++ b/src/components/pages/Roles/forms/RoleFormPaymentStream.tsx
@@ -28,8 +28,8 @@ function FixedDate({ formIndex }: { formIndex: number }) {
     });
   };
 
-  const setStartDate = values?.roleEditing?.payments?.[formIndex]?.startDate;
-  const setEndDate = values?.roleEditing?.payments?.[formIndex]?.endDate;
+  const selectedStartDate = values?.roleEditing?.payments?.[formIndex]?.startDate;
+  const selectedEndDate = values?.roleEditing?.payments?.[formIndex]?.endDate;
 
   return (
     <Box>
@@ -50,7 +50,7 @@ function FixedDate({ formIndex }: { formIndex: number }) {
           <GridItem area="start">
             <DecentDatePicker
               type="startDate"
-              maxDate={values?.roleEditing?.payments?.[formIndex]?.endDate}
+              maxDate={selectedEndDate ? addDays(selectedEndDate, -2) : undefined}
               formIndex={formIndex}
               onChange={date => onDateChange(date, 'startDate')}
             />
@@ -69,7 +69,7 @@ function FixedDate({ formIndex }: { formIndex: number }) {
           <GridItem area="end">
             <DecentDatePicker
               type="endDate"
-              minDate={values?.roleEditing?.payments?.[formIndex]?.startDate}
+              minDate={selectedStartDate ? addDays(selectedStartDate, 2) : undefined}
               formIndex={formIndex}
               onChange={date => onDateChange(date, 'endDate')}
             />
@@ -89,8 +89,8 @@ function FixedDate({ formIndex }: { formIndex: number }) {
         <DecentDatePicker
           type="cliffDate"
           formIndex={formIndex}
-          minDate={setStartDate ? addDays(setStartDate, 1) : undefined}
-          maxDate={setEndDate ? addDays(setEndDate, -1) : undefined}
+          minDate={selectedStartDate ? addDays(selectedStartDate, 1) : undefined}
+          maxDate={selectedEndDate ? addDays(selectedEndDate, -1) : undefined}
           onChange={(date: Date) => {
             setFieldValue(`roleEditing.payments[${formIndex}].cliffDate`, date);
           }}
--- a/src/components/pages/Roles/forms/RoleFormPaymentStream.tsx
+++ b/src/components/pages/Roles/forms/RoleFormPaymentStream.tsx
@@ -2,6 +2,7 @@ import { Box, Button, Flex, FormControl, Grid, GridItem, Icon } from '@chakra-ui
 import { ArrowLeft, ArrowRight } from '@phosphor-icons/react';
 import { addDays } from 'date-fns';
 import { FormikErrors, useFormikContext } from 'formik';
+import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { CARD_SHADOW } from '../../../../constants/common';
 import { useRolesStore } from '../../../../store/roles';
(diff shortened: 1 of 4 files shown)


adamgall (changes requested):
Screenshot 2024-09-27 at 10 42 15 AM

Seems like maybe here is a UX bug.

Can select the same date for both start and end.

The Cliff date, in that case, is unselectable (every date is disabled in the calendar). This makes sense, because the Cliff must be between start and end, exclusive.

But, the Save button is disabled.

Da-Colon (approved):
LG pending Adam's comment

adamgall:
> > Seems like maybe here is a UX bug.
> > Can select the same date for both start and end. Intuitively, I think this should be changed. End date should be at least one day after start date.
> > The Cliff date, in that case, is unselectable (every date is disabled in the calendar). This makes sense, because the Cliff must be between start and end, exclusive.
> > But, the Save button is disabled.
> 
> @adamgall Nice catch.
> 
> On "End date should be at least one day after start date", it should probably be _at least two_ days after start date, so there's an "in-between" date that will be valid for the cliff lol

From a technical point of view sure I understand what you're saying. That would allow us to always be able to set a Cliff, assuming we have this requirement for cliff to exist between start and end exclusively.

From a product point of view, is that what we want? _Maybe_ the cliff date is only available as long as there is at least one day in between start and end.

From a product point of view, I think we all would agree that a stream starting and ending on the same date doesn't make any sense. But a one day stream (with no cliff) does "make sense" imo.

DarksightKellar:
> From a product point of view, is that what we want? Maybe the cliff date is only available as long as there is at least one day in between start and end.

Hmm lemme put some thought into this

adamgall:
I'm gonna disagree with your latest commit @DarksightKellar. I would prefer that we don't require a "buffer" day between Start and End, and the Cliff day is simply unselectable if those Start and End days are sequential. The validation should still pass in this scenario though.

DarksightKellar:
> I'm gonna disagree with your latest commit @DarksightKellar. I would prefer that we don't require a "buffer" day between Start and End, and the Cliff day is simply unselectable if those Start and End days are sequential. The validation should still pass in this scenario though.

@adamgall checkout the latest latest commit

Only concern is the cliff date reset logic. While I like it, it's not super obvious this has happened to the user, and the save button remains enabled, and there's the risk they click save thinking they've got a cliff date (which would have been invalid) set, but they actually don't because it was reset.

adamgall:
@DarksightKellar good point. Let's ask @nicolaus-sherrill and @xanaramoss to chime in

xanaramoss:
> @DarksightKellar good point. Let's ask @nicolaus-sherrill and @xanaramoss to chime in

The disable button alone it's not enough, but accompanied by an alert banner may do the trick.
Something like "To select a cliff, the start and end dates must be at least 2 days apart./ Cliff selection requires at least 2 days between the start and end dates."

From a product stand a one-day stream (start on day 1, end on day 2, with no cliff) should be valid and not require a buffer.
What if the cliff date only appears after the user selects a valid start/end date? Like that, the cliff option is only available when there's enough time.

DarksightKellar:
> The disable button alone it's not enough, but accompanied by an alert banner may do the trick. Something like "To select a cliff, the start and end dates must be at least 2 days apart./ Cliff selection requires at least 2 days between the start and end dates."
> 
> From a product stand a one-day stream (start on day 1, end on day 2, with no cliff) should be valid and not require a buffer. What if the cliff date only appears after the user selects a valid start/end date? Like that, the cliff option is only available when there's enough time.

Oh I like that last suggestion, showing the cliff picker only if it makes sense to, based on selected start and end dates. Lemme dive down that route
cc  @adamgall @xanaramoss

adamgall:
I dig it

src/components/pages/Roles/forms/RoleFormPaymentStream.tsx
```diff
@@ -13,6 +14,7 @@ import { SectionTitle } from './RoleFormSectionTitle';
 function FixedDate({ formIndex }: { formIndex: number }) {
   const { t } = useTranslation(['roles']);
   const { values, setFieldValue } = useFormikContext<RoleFormValues>();
+  const [showCliffDatePicker, setShowCliffDatePicker] = useState(false);
```
adamgall, line 17: Nitpick but you don't really need to `useState` here. 

```ts
const showCliffDatePicker = !!startDate && !!endDate && addDays(startDate, 1) < endDate
```

should do

adamgall (commented):
Scenario:

User has selected start and end with > 1 day buffer, then selects a Cliff date.

Then, changes the End date to be Start+1, which hides the Cliff picker.

What happens to the cliff value in the form at that point? Is it nullified (as it should be?) Or does that Cliff date stick around in memory somewhere and end up borking the prop?

DarksightKellar:
> Scenario:
> 
> User has selected start and end with > 1 day buffer, then selects a Cliff date.
> 
> Then, changes the End date to be Start+1, which hides the Cliff picker.
> 
> What happens to the cliff value in the form at that point? Is it nullified (as it should be?) Or does that Cliff date stick around in memory somewhere and end up borking the prop?

It's "nullified" via: 
```
if (cliffDate && ((startDate && startDate >= cliffDate) || (endDate && endDate <= cliffDate))) {
      setFieldValue(`roleEditing.payments[${formIndex}].cliffDate`, undefined);
    }
```
in `onDateChange`

adamgall (approved):
Awesome, looking good sir

PR #2377, opened 2024-09-27
Use `fallbackTokenData`

### LINKED ISSUE #2369: On Treasury page, safe governance token shows as "ETH" instead of the token symbol
Steps to see the problem:

Check out [this safe's treasury page](https://app.dev.decentdao.org/treasury?dao=sep:0xa042F2b97855f72c4714B9E5ff9d603D2EE7F078). Nothing obvious amiss here, except:
- The "ETH" tx is actually meant to be [the safe's governance token](https://sepolia.etherscan.io/tx/0xd5457aac85e3171621757a461505127fc3b52f9097976e72933c36128414cf32)

Likely culprit is this exception from safe api:
Image

DarksightKellar:
Closes #2369 (~kinda~ fr)

This PR implements a fallback for transferred token data when `safeAPI.getToken` throws.

The cause of the "no token matches given query" exception is still unclear, as the address is quite clearly on etherscan (and Moralis returns its balance if the safe has some). ~While this fallback works for safes that have _some_ balance of the problem token, it doesn't for safes that have a 0-balance even if they at one time had some.~ The fallback retrieves token info from the blockchain directly.

Also the `setTimeout` retry in case of rate limited error was removed because I'm not sure about the end result of this PR yet and didn't want to commit to a lengthy refactor at this point. If needed, this functionality can be restored in some form during review.

### DIFF
--- a/src/hooks/DAO/loaders/useDecentTreasury.ts
+++ b/src/hooks/DAO/loaders/useDecentTreasury.ts
@@ -115,12 +115,12 @@ export const useDecentTreasury = () => {
       assetsDeFi,
       assetsNonFungible,
       totalUsdValue,
-      transfers: null, // transfers not yet loaded. these are setup hereafter
+      transfers: null, // transfers not yet loaded. these are setup below
     };
 
     action.dispatch({ type: TreasuryAction.UPDATE_TREASURY, payload: treasuryData });
 
-    const tokenAddresses = flattenedTransfers
+    const tokenAddressesOfTransfers = flattenedTransfers
       // map down to just the addresses, with a type of `string | undefined`
       .map(transfer => transfer.tokenAddress)
       // no undefined or null addresses
@@ -132,17 +132,30 @@ export const useDecentTreasury = () => {
 
     // Instead of this block of code, check the commented out snippet
     // below this for a half-implemented alternative.
-    const tokenData = await Promise.all(
-      tokenAddresses.map(async addr => {
+    const transfersTokenInfo = await Promise.all(
+      tokenAddressesOfTransfers.map(async addr => {
         try {
           return await safeAPI.getToken(addr);
         } catch (e) {
-          setTimeout(async () => {
-            // @note retry fetching token data in case of rate limit
-            if (addr) {
-              return safeAPI.getToken(addr);
-            }
-          }, 300);
+          const fallbackTokenData = tokenBalances?.find(
+            tokenBalanceData => getAddress(tokenBalanceData.tokenAddress) === addr,
+          );
+          if (!fallbackTokenData) {
+            return {
+              address: addr,
+              name: 'Unknown',
+              symbol: '???',
+              decimals: 18,
+            };
+          }
+
+          return {
+            address: addr,
+            name: fallbackTokenData?.name,
+            symbol: fallbackTokenData?.symbol,
+            decimals: fallbackTokenData?.decimals,
+            logoUri: fallbackTokenData?.logo,
+          };
         }
       }),
     );
@@ -185,11 +198,11 @@ export const useDecentTreasury = () => {
         };
         const transferTokenAddress = transfer.tokenAddress;
         if (transferTokenAddress) {
-          const token = tokenData.find(
+          const tokenData = transfersTokenInfo.find(
             _token => _token && getAddress(_token.address) === getAddress(transferTokenAddress),
           );
-          if (token) {
-            tokenInfo = token;
+          if (tokenData) {
+            tokenInfo = tokenData;
           }
         }
 
--- a/src/hooks/DAO/loaders/useDecentTreasury.ts
+++ b/src/hooks/DAO/loaders/useDecentTreasury.ts
@@ -133,16 +133,16 @@ export const useDecentTreasury = () => {
     // Instead of this block of code, check the commented out snippet
     // below this for a half-implemented alternative.
     const transfersTokenInfo = await Promise.all(
-      tokenAddressesOfTransfers.map(async addr => {
+      tokenAddressesOfTransfers.map(async address => {
         try {
-          return await safeAPI.getToken(addr);
+          return await safeAPI.getToken(address);
         } catch (e) {
           const fallbackTokenData = tokenBalances?.find(
-            tokenBalanceData => getAddress(tokenBalanceData.tokenAddress) === addr,
+            tokenBalanceData => getAddress(tokenBalanceData.tokenAddress) === address,
           );
           if (!fallbackTokenData) {
             return {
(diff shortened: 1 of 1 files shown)


src/hooks/DAO/loaders/useDecentTreasury.ts
```diff
@@ -132,17 +132,30 @@ export const useDecentTreasury = () => {
 
     // Instead of this block of code, check the commented out snippet
     // below this for a half-implemented alternative.
-    const tokenData = await Promise.all(
-      tokenAddresses.map(async addr => {
+    const transfersTokenInfo = await Promise.all(
+      tokenAddressesOfTransfers.map(async addr => {
```
mudrila, line 136: Is there other `address` variable in the scope? Why just not naming it `address`?

src/hooks/DAO/loaders/useDecentTreasury.ts
```diff
@@ -132,17 +132,30 @@ export const useDecentTreasury = () => {
...
+              address: addr,
+              name: 'Unknown',
+              symbol: '???',
+              decimals: 18,
+            };
+          }
+
+          return {
+            address: addr,
+            name: fallbackTokenData?.name,
```
mudrila, line 154: Seems like this optional chaining is redundant - we already doing early return

src/hooks/DAO/loaders/useDecentTreasury.ts
```diff
@@ -132,17 +132,30 @@ export const useDecentTreasury = () => {
...
-              return safeAPI.getToken(addr);
-            }
-          }, 300);
+          const fallbackTokenData = tokenBalances?.find(
+            tokenBalanceData => getAddress(tokenBalanceData.tokenAddress) === addr,
+          );
+          if (!fallbackTokenData) {
+            return {
+              address: addr,
+              name: 'Unknown',
```
mudrila, line 146: Wondering if that makes sense to also add some sort of tooltip in place where we're displaying to outline that we weren't able to load token metadata

src/hooks/DAO/loaders/useDecentTreasury.ts
```diff
@@ -132,17 +132,30 @@ export const useDecentTreasury = () => {
...
+              symbol: '???',
```
mudrila, line 147: I'd suggest using `---` instead - but that's just "my personal user preference" - kinda makes more sense to me personally that `---` means we weren't able to fetch something in comparison to `???` which for me means "some unexpected shit is going on"

mudrila (commented):
Some comments sir - but mostly code styling / "my user feedback" - so feel free to reason your vision and I'll approve :)

src/hooks/DAO/loaders/useDecentTreasury.ts
DarksightKellar: Because `address` on its own doesn't say anything about what and where these addresses came from

src/hooks/DAO/loaders/useDecentTreasury.ts
DarksightKellar: True

src/hooks/DAO/loaders/useDecentTreasury.ts
DarksightKellar: Wholeheartedly agree

src/hooks/DAO/loaders/useDecentTreasury.ts
DarksightKellar: idk tbh. And I also don't know if this is worth pursuing if this issue is potentially a testnet only problem

DarksightKellar:
@decentdao/engineering Looking for feedback re:
> Also the `setTimeout` retry in case of rate limited error was removed because I'm not sure about the end result of this PR yet and didn't want to commit to a lengthy refactor at this point. If needed, this functionality can be restored in some form during review

See diff on `140`

src/hooks/DAO/loaders/useDecentTreasury.ts
mudrila: I'd say neither `addr` - that's why either suggesting not shortcutting or putting proper variable naming
But this one is super minor - I just personally don't line shortcut variable naming - feel free to ignore :)

DarksightKellar: oh lmaooo you meant the `addr` _in the callback_. That wasn't diff from this PR so didn't realise that's what you were referring to. Thought you meant `tokenAddressesOfTransfers` which is what I'd updated.

I mean, I don't mind it but sure I can update it too.

src/hooks/DAO/loaders/useDecentTreasury.ts
```diff
@@ -132,17 +132,30 @@ export const useDecentTreasury = () => {
...
-              return safeAPI.getToken(addr);
-            }
-          }, 300);
+          const fallbackTokenData = tokenBalances?.find(
+            tokenBalanceData => getAddress(tokenBalanceData.tokenAddress) === address,
+          );
+          if (!fallbackTokenData) {
+            return {
+              address,
+              name: 'Unknown',
```
adamgall, line 146: Nitpicky but let's put this "Unknown" string into a localization file

adamgall (changes requested):

src/hooks/DAO/loaders/useDecentTreasury.ts
```diff
@@ -132,17 +132,30 @@ export const useDecentTreasury = () => {
...
+              symbol: '---',
+              decimals: 18,
```
adamgall, line 148: Also, we don't know that this unknown token is actually 18 decimal places, right?

I'm almost thinking that maybe we just straight up _ignore_ any transfer records where attempting to get the token data from Safe API fails, instead of attempting to backfill those records with dummy data.

adamgall (approved):

src/hooks/DAO/loaders/useDecentTreasury.ts
DarksightKellar: No longer an issue after [this commit](https://github.com/decentdao/decent-interface/pull/2377/commits/c74d5ce11cd8f4bec83fab412917886e6ecde2a9)

src/hooks/DAO/loaders/useDecentTreasury.ts
DarksightKellar: Resolved in [this commit](https://github.com/decentdao/decent-interface/pull/2377/commits/c74d5ce11cd8f4bec83fab412917886e6ecde2a9). And, whew.

mudrila (approved):
Lezzgooo

PR #2380, opened 2024-09-27
Attempt switching chain automatically when on wrong network

mudrila:
## The problem

When user navigates directly to the DAO, which is not on mainnet (base - for instance) - app default to mainnet anyway and displays "wrong network" error.

## The solution

This PR implements attempt to automatically switch network. Far from being perfect - wrong network screen will be displayed as long as network is not switched. Automatic attempt might fail due to intermittent RPC errors. If wallet is not connected - it pretty much always works. If wallet is connected - some issues might occur depending on specific wallet used (if wallet does not support adding chain alongside with switching at single request).

This is kinda hot-fix and we'll need longer term solution - temporary patch to improve UX for MCON.

## Testing 
1. Try visiting https://app.decentdao.org/home?dao=base:0x0BcC8861d36B610f19492C8E512Ebb9E99BB7654 - you will just see error screen about wrong network, if you weren't connected to `Base`.
2. Try visiting same DAO on the deployed preview https://deploy-preview-2380.app.decentdao.org/home?dao=base:0x0BcC8861d36B610f19492C8E512Ebb9E99BB7654. 
3. If you weren't connected - app should just switch automatically, error might flash for a moment but then DAO dashboard should appear after page reload. 
4. If you were connected - it should prompt you to switch chain in your wallet and then DAO will load

### DIFF
--- a/src/hooks/DAO/useDAOController.ts
+++ b/src/hooks/DAO/useDAOController.ts
@@ -56,5 +56,5 @@ export default function useDAOController() {
   useKeyValuePairs();
   useHatsTree();
 
-  return { invalidQuery, wrongNetwork, errorLoading };
+  return { invalidQuery, wrongNetwork, errorLoading, urlAddressPrefix: addressPrefix };
 }
--- a/src/pages/DAOController.tsx
+++ b/src/pages/DAOController.tsx
@@ -48,6 +48,7 @@ export default function DAOController() {
     if (urlAddressPrefix && wrongNetwork) {
       try {
         switchChain({ chainId: getChainIdFromPrefix(urlAddressPrefix) });
+        window.location.reload();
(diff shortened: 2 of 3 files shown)


Da-Colon (commented):
Def works, does the loaders breaking affect normal navigation or just covers us in case of direct url changes?

mudrila:
> Def works, does the loaders breaking affect normal navigation or just covers us in case of direct url changes?

Only for direct navigation

DarksightKellar (approved):

DarksightKellar:
Changes look good to me. Not at my desk though so untested manually

Da-Colon (approved):
I'll approve, can you create an issues to circle back to loaders later?

PR #2386, opened 2024-10-01
Configurable fractal module attachment

### LINKED ISSUE #2385: Allow creating subDAO without attaching `FractalModule`
In order to meet requirements for Coordinape - we need to adjust subDAO creation flow, so that attaching `FractalModule` is configurable from UX.
The idea behind it is to allow Parent DAO to "Veto" Child DAO through freezing, but eliminate backdoor that `FractalModule` provides to the Parent DAO.

mudrila:
## Changes

- Add Checkbox into Child Safe creation flow to allow user control whether `FractalModule` should be deployed & enabled on Child Safe or not
- Adjust Child Safe creation logic accordingly to that setting
- Do not display "Initiate clawback" option from `ManageDAOMenu` when `FractalModule` is not attached to the Child Safe

Closes #2385 

## Testing

1. Create Parent Safe (Multisig or any other type - doesn't really matter)
2. Create Child Safe with "Enable Clawback" checkbox being checked (it is checked by default)
3. Navigate to Child Safe settings page
4. Make sure `FractalModule` appears in the list of Safe Modules
5. Freeze Child Safe
6. Make sure "Initiate clawback" option appears in the `ManageDAOMenu` once Child Safe is frozen
7. Create Child Safe with "Enable Clawback" checkbox being unchecked (make sure to hit this buddy)
8. Navigate to 2nd Child Safe settings page
9. Make sure `FractalModule` does not appear in the list of Safe Modules
10. Freeze 2nd Child Safe
11. Make sure "Initiate clawback" option does not appear in the `ManageDAOMenu` once Child Safe is frozen
12. Celebrate

## Updated UI

Screenshot 2024-10-01 at 17 37 21


tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "preserve",
+    "jsx": "react-jsx",
```
mudrila, line 17: My IDE was yelling at me about `React refers to UMD global...`

`"jsx": "preserve"` was set quite a while ago during migration from `CRA` to `Next.js` - so not as much valid any longer after migrating to Vite (but even latest `Next.js` configuration sets `jsx` option to something else, not `preserve`).

@decentdao/engineering Please verify this setting on your local - make sure it does not conflict with your local setup. Otherwise - I'll be looking into alternative ways to make my IDE happy about `jsx: preserve` 🤣

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "ESNext",
     "lib": ["dom", "dom.iterable", "esnext"],
-    "allowJs": true,
+    "allowJs": false,
```
mudrila, line 5: We don't wanna vanilla JS I believe :)

src/models/DaoTxBuilder.ts
```diff
@@ -155,8 +164,8 @@ export class DaoTxBuilder extends BaseTxBuilder {
     }
 
     // If subDAO, deploy Fractal Module
-    if (this.parentAddress) {
-      txs.push(this.deployFractalModuleTx!);
+    if (this.parentAddress && this.deployFractalModuleTx) {
+      txs.push(this.deployFractalModuleTx);
```
mudrila, line 168: @adamgall You gonna like this killed `!` thingy 😄

DarksightKellar (dismissed):
Couple 'o suggestions, then LGTM! 
Not tested yet

src/models/DaoTxBuilder.ts
```diff
@@ -112,9 +116,14 @@ export class DaoTxBuilder extends BaseTxBuilder {
         this.parentStrategyAddress,
       );
 
+      if (this.enableFractalModuleTx) {
+        this.internalTxs = this.internalTxs.concat([
+          // Enable Fractal Module b/c this a subDAO
+          this.enableFractalModuleTx,
+        ]);
```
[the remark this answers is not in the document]
mudrila: Actually - almost never has to be an array, only in cases where we spread another array into args conditionally.
Gonna update rest of usages whenever it's reasonable

src/components/DaoCreator/index.tsx
```diff
@@ -45,7 +46,7 @@ function DaoCreator({
           const choosenGovernance = values.essentials.governance;
           const freezeGuard = isSubDAO ? values.freeze : undefined;
 
-          let daoData: SafeMultisigDAO | AzoriusERC20DAO | AzoriusERC721DAO | undefined;
+          let daoData: SafeMultisigDAO | AzoriusERC20DAO | AzoriusERC721DAO | SubDAO | undefined;
```
adamgall, line 49: What is this SubDAO type and why is it needed here now?

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -92,24 +93,22 @@ const useBuildDAOTx = () => {
...
+          parentStrategyAddress = linearVotingErc721Address;
         }
       }
 
-      const daoTxBuilder = txBuilderFactory.createDaoTxBuilder(
-        parentVotingStrategyType,
-        parentVotingStrategyAddress,
-      );
+      const daoTxBuilder = txBuilderFactory.createDaoTxBuilder({
+        attachFractalModule: (daoData as SubDAO).attachFractalModule,
```
adamgall, line 108: is the `useBuildDAOTx` hook used when creating a fresh parent DAO? if so, will casting this `daoData` as a `SubDAO` type work correctly here?

src/models/DaoTxBuilder.ts
```diff
@@ -211,8 +225,8 @@ export class DaoTxBuilder extends BaseTxBuilder {
     ];
 
     // If subDAO, deploy Fractal Module.
```
adamgall, line 227: Nitpick but update this comment plz

src/hooks/DAO/useDeployAzorius.ts
```diff
@@ -91,7 +91,7 @@ const useDeployAzorius = () => {
 
       txBuilderFactory.setSafeContract(daoAddress);
 
-      const daoTxBuilder = txBuilderFactory.createDaoTxBuilder();
+      const daoTxBuilder = txBuilderFactory.createDaoTxBuilder({ attachFractalModule: false });
```
adamgall, line 94: `useDeployAzorius` hook is only used when creating new top level safes?

adamgall (commented):
Just some questions to start

src/hooks/DAO/useDeployAzorius.ts
mudrila: Good point - actually, might be used in scope of subDAO as well, but I don't think it was ever tested - subDAOs were implemented/finalized already after this `useDeployAzorius`.
`useDeployAzorius` is used to deploy / attach Azorius to an existing Safe - which in theory might be subDAO.

I think it worth implementing / figuring this out in scope of separate PR

src/hooks/DAO/useBuildDAOTx.ts
mudrila: Potentially - it will evaluate to `undefined` in case of parent DAO deployment - but the code where this parameter would be used won't be reached because of not passed `parentAddress` - so we should be safe.
I can make it more explicit through optionality of `attachFractalModule` parameter on `createDaoTxBuilder` - WDYT?

adamgall: IMO explicit is always better

src/hooks/DAO/useDeployAzorius.ts
adamgall: I'm a bit confused then. Is it accurate to always pass `false` in here if this might be used from a sub dao? Will the toggle selection be utilized?

adamgall (approved):

src/hooks/DAO/useDeployAzorius.ts
mudrila: Not accurate - but that's the thing. This hook was never actually being tested on subDAOs - so it pretty much won't work for subDAOs at the moment at all

mudrila: I added `@todo` to outline this problem on a code level - gonna create an issue to track it down.
However, I don't think this is a priority atm - but yeah, definitely we don't want to get it lost

xanaramoss (approved):

PR #2388, opened 2024-10-02
Fix streams race conditions

### LINKED ISSUE #2331: What Are We Gonna Do When Blockchain State Changes During "Roles Edits" Proposals?
unaffected user stories

- add hats
- role details change
- add stream

affected user stories

ASSUMPTION:

we (the app, at proposal creation time) assume that the state of the blockchain at proposal creation will stay unchanged (except for time passing) by the time proposal execution eventually happens.

the following scenarios will render one or more of the above set of assumptions invalid:

- role member changed
    - IF any streams end streaming OR any streams are cancelled DURING the proposal time
        - AND the original recipient CLAIMS their funds after the stream ends or is cancelled during this proposal time
            - THEN the proposal will attempt to withdraw funds from an empty Sablier stream
                - RESULTING IN A REVERTED TRANSACTION
    - IF any streams start streaming funds DURING the proposal time
        - THEN the proposal will not include a call to claim and withdraw from the Sablier stream at execution
            - RESULTING IN OUTGOING MEMBER FUNDS GOING TO THE INCOMING MEMBER
    - IF any streams are not actively streaming but DO have funds to withdraw from Sablier
        - AND the original recipient CLAIMS their funds from the stream DURING the proposal
            - THEN the proposal will attempt to withdraw funds from an empty Sablier stream
                - RESULTING IN A REVERTED TRANSACTION
- role deleted
    - IF any streams end streaming OR any streams are cancelled DURING the proposal time
        - THEN the proposal will attempt to cancel a stream which is already over or cancelled
            - RESULTING IN A REVERTED TRANSACTION
    - IF any streams start streaming funds DURING the proposal time
        - THEN the proposal will not include a transaction to withdraw any streamed funds from the newly cancelled Sablier stream
            - RESULTING IN LOST MEMBER FUNDS
    - IF any streams are not actively streaming but DO have funds to withdraw from Sablier
        - AND the recipient CLAIMS their funds from the stream DURING the proposal
            - THEN the proposal will attempt to withdraw funds from an empty Sablier stream
                - RESULTING IN A REVERTED TRANSACTION
- stream edited (not past end date and not cancelled)
    - IF stream ends OR stream is cancelled DURING the proposal time
        - THEN the proposal will attempt to cancel the stream which is already over or cancelled
            - RESULTING IN A REVERTED TRANSACTION
    - IF the stream start streaming funds DURING the proposal time
        - THEN the proposal will not include a transaction to withdraw any streamed funds from the newly cancelled Sablier stream
            - RESULTING IN LOST MEMBER FUNDS
    - IF the stream is not actively streaming but DOES have funds to withdraw from Sablier
        - AND the recipient CLAIMS their funds from the stream DURING the proposal
            - THEN the proposal will attempt to withdraw funds from an empty Sablier stream
                - RESULTING IN A REVERTED TRANSACTION
- stream cancelled
    - IF stream ends OR stream is cancelled DURING the proposal time
        - THEN the proposal will attempt to cancel the stream which is already over or cancelled
            - RESULTING IN A REVERTED TRANSACTION
    - IF the stream starts streaming funds DURING the proposal time
        - THEN the proposal will not include a transaction to withdraw any streamed funds from the newly cancelled Sablier stream
            - RESULTING IN LOST MEMBER FUNDS
    - IF the stream is not actively streaming but DOES have funds to withdraw from Sablier
        - AND the recipient CLAIMS their funds from the stream DURING the proposal
            - THEN the proposal will attempt to withdraw funds from an empty Sablier stream
                - RESULTING IN A REVERTED TRANSACTION

DarksightKellar:
Closes #2331

Reproducing the original issue: See [this comment](https://github.com/decentdao/decent-interface/issues/2331#issuecomment-2388939136). Test on develop. These issues should not exist on this PR


src/components/pages/Roles/forms/RoleFormAssetSelector.tsx
```diff
@@ -304,6 +304,8 @@ export function AssetSelector({ formIndex, disabled }: { formIndex: number; disa
                 <BigIntInput
                   isDisabled={inputDisabled}
                   value={field.value?.bigintValue}
+                  // @todo: line below is temporary -- remove
+                  parentFormikValue={values?.roleEditing?.payments?.[formIndex]?.amount}
```
adamgall, line 308: When does this todo need to be addressed?

src/components/ui/forms/EthAddressInput.tsx
```diff
@@ -24,13 +25,18 @@ export function AddressInput({ value, onChange, ...rest }: InputProps) {
     };
   }, [debounceValue]);
 
+  useEffect(() => {
+    setLocalValue(value);
+  }, [value]);
+
```
adamgall, line 31: Does this need to be in a useEffect? My gut says no.

DarksightKellar: Yup

adamgall (approved):

src/components/pages/Roles/RoleCard.tsx
```diff
@@ -62,7 +61,7 @@ export function AvatarAndRoleName({
         >
           {wearerAddress ? accountDisplayName : t('unassigned')}
         </Text>
-        {isFeatureEnabled('STREAMS') && paymentsCount !== undefined && (
+        {paymentsCount !== undefined && (
```
mudrila, line 64: Does this mean we're fully enabling payments on prod? Is this intentional?

src/components/pages/Roles/forms/RoleFormPaymentStream.tsx
```diff
@@ -201,6 +202,25 @@ export default function RoleFormPaymentStream({ formIndex }: { formIndex: number
                 {t('save')}
               </Button>
             )}
+            {isDevMode() && (
```
mudrila, line 205: Maybe we should remove it before merging?

src/components/ui/forms/BigIntInput.tsx
```diff
@@ -19,29 +19,30 @@ export interface BigIntInputProps
...
 export function BigIntInput({
+  // @todo `value` can most likely either be removed, or be used for what `parentFormikValue` is currently being used for. Currently it works as nothing more than an initial value.
   value,
   onChange,
   decimalPlaces = 18,
   min,
   max,
   maxValue,
-  currentValue,
+  parentFormikValue = { value: '', bigintValue: undefined },
```
mudrila, line 45: I don't really like the fact we're bounding prop name to library name while component is not supposed to be used only in that context and also this prop is not used for anything specific to Formik within the component itself. Would suggest you either rename it and note todo to refactor it to make use of formik context, if we're coming from the fact that this input should be used only in formik context

mudrila (approved):
Few non-blocking comments, don't wanna hold this PR - I'm cool with merging as-is but wanna make sure we're intentional about merging dev mode code

src/components/pages/Roles/RoleCard.tsx
adamgall: Yes it does, and yes it is!

src/components/pages/Roles/forms/RoleFormPaymentStream.tsx
adamgall: We talked about this on a huddle today... decided to leave it as-is. We can use this special mode to enable more "helper UX" for us in our development environments.

src/components/ui/forms/BigIntInput.tsx
adamgall: Ah, I do agree with this. Will create an issue from this comment.

src/components/ui/modals/SendAssetsModal.tsx
```diff
@@ -136,7 +136,8 @@ export function SendAssetsModal({
                           iconSize="1.5rem"
                           icon={<CaretDown />}
                           onChange={e => {
-                            setFieldValue('inputAmount', { value: '0', bigintValue: 0n });
+                            // New asset selected. First reset the form input amount
+                            setFieldValue('inputAmount', undefined);
```
Da-Colon, line 140: ah good call

src/components/pages/Roles/forms/RoleFormPaymentStream.tsx
Da-Colon, line 205: As I mentioned on the call. Happy to see this as I was just thinking about this for us. Buttons, informational overlays, stuff that will make QA easier for us. @mudrila opportunity to maybe add a easter egg to override the ENV and enable it on prod/dev

Da-Colon (approved):
I like all the added comments. belated approval

PR #2391, opened 2024-10-03
Fix redirect loaders

### LINKED ISSUE #2381: Redirect loaders in `router.tsx` aren't working (`daoAddress` param is `undefined` which leads to wrong redirect)

mudrila:
Closes #2381


src/main.tsx
```diff
@@ -1,26 +1,32 @@
...
 
-function FractalRouterProvider() {
+function DecentRouterProvider() {
   const { addressPrefix } = useNetworkConfig();
-  return <RouterProvider router={router(addressPrefix)} />;
+  const urlParams = new URLSearchParams(window.location.search);
+  const addressWithPrefix = urlParams.get('dao');
+
+  const prefixAndAddress = addressWithPrefix?.split(':');
+  const daoAddressStr = prefixAndAddress?.[1];
```
adamgall, line 19: Considering this is all classified as "user input", do we need to do any sanitization on this data?

adamgall (approved):

src/main.tsx
mudrila: Well, sanitization kinda gonna happen anyway after the redirect through the logic of fetching DAO - so I feel like doing it here as well to prevent redirect / redirect to app homepage is kinda redundant - we'll be taking from the user ability to see that there's something wrong with the address

PR #2394, opened 2024-10-03
Prevent fetching DeFi positions when Moralis does not support it

### LINKED ISSUE #2390: Do not fetch `defiBalances` on unsupported chains (like `Sepolia`)
We're trying to fetch `defiBalances` on `Sepolia`, which is became unsupported very recently. 
This is causing redundant requests and error toast appearance - which we should avoid.

mudrila:
Closes #2390


Da-Colon (approved):
🤘

adamgall (approved):
Nice and easy

PR #2397, opened 2024-10-04
Fix original wearer bug

DarksightKellar:
The dev team discovered a few bugs that will cause the wrong role member to be set in some scenarios, and in one instance, for an edited role to fail to transfer the hat back from the safe to the original/intended wearer. 

This PR addresses those issues.

### DIFF
--- a/src/hooks/utils/useCreateRoles.ts
+++ b/src/hooks/utils/useCreateRoles.ts
@@ -713,12 +713,16 @@ export default function useCreateRoles() {
                 // Cancel the stream
                 allTxs.push(prepareCancelStreamTx(stream.streamId, stream.contractAddress));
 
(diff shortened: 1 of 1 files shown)


mudrila (approved):
Jeez, I feel like "why da hell it was implemented this way before" 🤣

DarksightKellar:
> @DarksightKellar can you talk about the testing you did on these changes?

I didn't get the chance to arrange for testing these specific changes yet, especially as they were originally part of the larger fix streams race conditions effort which I've been focused on.

One of the bugs is ID'd [here (bottom comment)](https://github.com/decentdao/decent-interface/issues/2331#issuecomment-2388939136) though, and is expected to be fixed now with this update.

PR #2414, opened 2024-10-16
Fix bug causing ENS input in role member edits to crash

### LINKED ISSUE #2411: Verify ENS string doesn't break formHat.address
[For all instances of `getAddress(formHat.wearer)` we should confirm that at this point, `formHat.wearer` is definitely an `Address` type.](https://github.com/decentdao/decent-interface/pull/2388/files#:~:text=//%20%40todo%3A%20For%20all%20instances%20of%20%60getAddress(formHat.wearer)%60%20we%20should%20confirm%20that%20at%20this%20point%2C%20%60formHat.wearer%60%20is%20definitely%20an%20%60Address%60%20type.)

DarksightKellar:
Closes #2411

This PR is probably messy TBVH (but fixes the bug and side effect bugs that were exposed after that, so should be complete) so looking for some callouts here.

To summarise: 

- The role form values are not actually processed until at or near proposal creation time
- ~Until then, whatever the user input, as long as it passes ENS validation, should be what the form value holds. This means we should only attempt to grab the address behind the ENS string _at the time_ when `getAddress` was previously being used and would've failed.~
- ~Since all role edit objects are being processed in parallel in `useCreateRoles` before actual proposal preparation begins, there's one point at which we can transform all strings to `Address` types.~
- In `RoleFormCreateProposal.tsx:100,101`, `roleHat.wearer` as a form value is both `undefined`-able and may hold an ENS string. `getAddress` removed.
- EDIT: The solution this PR currently implements is: as the user types in the member input field, as long as the input may be a valid ENS name, we are resolving the input to an address at the same time, on the same role form value. This new prop is what _should be_ used for `Address`-related logic.
- I've renamed some references to `wearerAddress` to `wearer` (and vice versa) to make it more obvious that we're dealing with a string that is possiby _not_ an `Address` (or to make it obvious that we're dealing with an `Address` type strictly).

Half the diffs in here are from a rename of `useDisplayName` to `useGetAccountName`


src/components/pages/Roles/RolesDetailsDrawer.tsx
```diff
@@ -63,12 +63,22 @@ export default function RolesDetailsDrawer({
   const {
     node: { daoAddress },
   } = useFractal();
-  const { chain } = useNetworkConfig();
+
+  const { address: roleHatWearer } = useAddress(roleHat.wearer);
+
+  const roleHatWearerAddress = useMemo(() => roleHatWearer ?? zeroAddress, [roleHatWearer]);
+
+  const { getDAOName } = useGetDAONameDeferred();
```
adamgall, line 71: I know this is somewhat legacy code from before changes, but since you're here no need to continue using the wrong hook. This `useGetDaoName[Deferred]` hook is for getting DAO names, specifically, not really for simply resolving ENS names.

src/components/pages/Roles/RolesDetailsDrawerMobile.tsx
```diff
@@ -50,6 +50,13 @@ export default function RolesDetailsDrawerMobile({
 
   if (!daoAddress || !hatsTree) return null;
 
+  let roleHatWearerAddress: Address;
+  try {
+    roleHatWearerAddress = getAddress(roleHat.wearer);
+  } catch (error) {
+    throw new Error('Invalid wearer address');
+  }
+
```
adamgall, line 59: I'm surprised that this code isn't duplicated (or shared) with the non-mobile version of this component. I'd expect that the differences between `RolesDetailsDrawer` and `RolesDetailsDrawerMobile` are only visual. Should they behave differently too?

cc: @Da-Colon

src/components/pages/Roles/forms/RoleFormInfo.tsx
```diff
@@ -1,14 +1,29 @@
...
+    useAddress(roleWearerString);
+
+  const { setFieldValue } = useFormikContext<RoleFormValues>();
+
+  useEffect(() => {
+    if (isValidWearerAddress) {
+      setFieldValue('roleEditing.resolvedWearer', resolvedWearerAddress);
+    }
+  }, [isValidWearerAddress, resolvedWearerAddress, setFieldValue]);
+
```
adamgall, line 26: Yeah I dig this

src/components/pages/Roles/forms/RoleFormInfo.tsx
```diff
@@ -103,6 +106,8 @@ export default function RoleFormInfo() {
                   setFieldTouched(field.name, true);
                 }}
                 onChange={e => {
+                  const inputWearer = e.target.value;
+                  setRoleWearerString(inputWearer);
```
adamgall, line 110: Super nitpick but why create a `const` for `inputWearer` when just using it once?

adamgall (changes requested):

src/components/pages/Roles/RolesDetailsDrawerMobile.tsx
mudrila: Not from the validation point - that's certainly. Yet, there's somewhat good place to generally refactor / restructure how modals are done - it seems like we're shifting towards "modal on desktop, drawer on mobile" pattern, and the amount of modals we're having is kinda getting bigger - so our "global modal" component growing too much (I knew it will become a problem from the very beginning 🤷🏼‍♂️ ).

I'd like to get that refactored - but think we can keep it "as is" in this PR - I believe that's out of scope here

src/components/pages/Roles/RolesDetailsDrawer.tsx
```diff
@@ -63,12 +63,22 @@ export default function RolesDetailsDrawer({
...
+  const [accountDisplayName, setAccountDisplayName] = useState(roleHat.wearer);
+  useEffect(() => {
+    if (!!roleHatWearer) {
+      getDAOName({ address: roleHatWearer }).then(displayName =>
```
mudrila, line 75: Ooof, that's definitely smells 🤣 
I know that under the hood `getDAOName` resolves ENS name.
Yet, using this hook specifically for that purpose makes code quite confusing.

I'd suggest either extracting that logic of ENS name / account display name showing (BTW - seems like we're usually showing truncated version, maybe same should apply here and link to the Etherscan? @nicolaus-sherrill @adamgall @Da-Colon @xanaramoss WDYT?)

src/components/pages/Roles/RolesDetailsDrawerMobile.tsx
mudrila: Ooooh, so this logic was just added - sorry, haven't recognized that before actually going into the code. Yeah, I think it would rather be better to remove it here. I assume issues are popping up when going directly to role details page on mobile and while tree / role initially loaded - so the drawer is empty or some errors due to ENS resolving are thrown - then I think it is better to adjust resolving logic, not this component. Maybe we need loading state here 

cc @nicolaus-sherrill 

I'd assume our bar loader should work

src/components/pages/Roles/types.tsx
```diff
@@ -119,6 +120,9 @@ export interface RoleHatFormValue
   extends Partial<Omit<DecentRoleHat, 'id' | 'wearer' | 'payments'>> {
   id: Hex;
   wearer?: string;
+  // Not a user-input field.
+  // `resolvedWearer` is auto-populated from the resolved address of `wearer` in case it's an ENS name.
```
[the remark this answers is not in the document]
mudrila: At least that's what I see through the code - so I think it's all good here

src/hooks/utils/useCreateRoles.ts
```diff
@@ -771,14 +771,20 @@ export default function useCreateRoles() {
...
+              return null;
+            }
+            return {
+              ...hat,
+              editedRole: hat.editedRole,
+              wearer: hat.resolvedWearer,
+            };
+          }),
+        )
+      ).filter(hat => hat !== null);
```
mudrila, line 787: Hmm, doesn't `filter` first works here? Will be less mapping then
Nitpick though, feel free to ignore

mudrila (changes requested):
Bunch of comments sir

src/components/pages/Roles/forms/RoleFormInfo.tsx
DarksightKellar: It was meant to be used below as well in `setFieldValue`, whoops. Will update

src/components/pages/Roles/types.tsx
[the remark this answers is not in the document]
DarksightKellar: Yep

src/hooks/utils/useCreateRoles.ts
DarksightKellar: Nah I don't think it will

src/components/pages/Roles/RolesDetailsDrawer.tsx
DarksightKellar: I've done a good bit of refactoring here now. Just trying to track down a potential render loop (that's not crashing the app, just almost 150 `Status: 429` logs as soon as you open the safes drawer

src/hooks/utils/useGetAccountName.ts
```diff
@@ -2,20 +2,51 @@ import { abis } from '@fractal-framework/fractal-contracts';
 import { useCallback, useEffect, useState } from 'react';
 import { Address, PublicClient, getContract } from 'viem';
 import { usePublicClient } from 'wagmi';
-import { useNetworkConfig } from '../../providers/NetworkConfig/NetworkConfigProvider';
-import { createAccountSubstring } from '../utils/useDisplayName';
-import { demoData } from './loaders/loadDemoData';
+import { demoData } from '../DAO/loaders/loadDemoData';
+import { createAccountSubstring } from './useDisplayName';
+
+export const getSafeNameFallback = async (
```
DarksightKellar, line 8: Separated out safe-specific logic so that the newly named `getAccountName` can be used in other contexts

src/components/pages/Roles/RolesDetailsDrawerMobile.tsx
DarksightKellar: Ok, so I've made a bunch of updates after these comments. Plz see recent changes.

src/components/pages/Roles/types.tsx
```diff
@@ -182,3 +190,11 @@ export type PreparedEditedStreamData = PreparedNewStreamData & {
   roleHatSmartAddress: Address;
   streamContractAddress: Address;
 };
+
+export interface RoleDetailsDrawerProps {
```
DarksightKellar, line 194: Extracted this interface out since it was exactly the same as `RoleDetailsDrawerMobileProps`

src/pages/daos/[daoAddress]/roles/edit/index.tsx
```diff
@@ -72,7 +72,10 @@ function RolesEdit() {
         title: '',
         description: '',
       },
-      hats,
+      hats: hats.map(hat => ({
+        ...hat,
+        resolvedWearer: hat.wearerAddress,
```
Comment (DarksightKellar) [inline src/pages/daos/[daoAddress]/roles/edit/index.tsx:77]: `initialValues` is a `RoleFormValues[]`, but `hats` is a `DecentRoleHat[]`. This makes roles that already existed won't have an empty `RoleFormValues.resolvedWearer`, which led to undefined states when switching to edit views.

src/components/pages/Roles/RolesDetailsDrawerMobile.tsx
```diff
@@ -48,6 +40,10 @@ export default function RolesDetailsDrawerMobile({
     [roleHat.payments],
   );
 
+  let roleHatWearer = 'wearer' in roleHat ? roleHat.wearer : roleHat.wearerAddress;
```
mudrila, line 43: Why `let` though?
Also, is it a trick for simplifying type compatibility constraints? Aka why `roleHat.wearer ?? roleHat.wearerAddress` doesn't work?

src/hooks/utils/useCreateRoles.ts
mudrila: Yeah I've meant 
```
const modifiedHats: RoleHatFormValueEdited[] = (
        await Promise.all(
          values.hats.filter(hat => !!hat.editedRole).map(async hat => ({
              ...hat,
              editedRole: hat.editedRole,
              wearer: hat.resolvedWearer,
            })
          ),
        )
      );
```
This way filtering out `hat` with existing `editedRole` would happen before doing `map` - thus less iterations during `map`
Also it feels like `Promise.all` and `async hat` are redundant here 🤔 

But yeah, don't worry about this one - rather micro-optimization

mudrila (approved):
Couple questions / comments, but LGTM :shipit:

src/components/pages/Roles/RolesDetailsDrawerMobile.tsx
DarksightKellar: Gahh `let` was not intentional. Nice catch.

And yes the line is a trick so I don't have to do `(roleHat as Type1WhoseNameIDontRemember).wearer ?? roleHat.wearerAddress`

src/pages/home/SafeDisplayRow.tsx
```diff
@@ -24,9 +26,12 @@ export function SafeDisplayRow({ address, network, onClick, showAddress }: SafeM
...
 
-  const { daoName } = useGetDAOName({
-    address: getAddress(address),
+  const publicClient = usePublicClient();
+
+  const { accountName: safeName } = useGetAccountName({
+    address,
     chainId: getChainIdFromPrefix(network),
+    getAccountNameFallback: () => getSafeNameFallback(address, fractalRegistry, publicClient),
   });
```
adamgall, line 35: Why is this "fallback" function needed here?

src/hooks/utils/useGetAccountName.ts
adamgall, line 1: Idk, the functions in this still seem pretty redundant and convoluted. 

Three exported functions: `useGetAccountName`, `useGetAccountNameDeferred`, `getSafeNameFallback`. Not clear to me at all when one should be used over the other.

adamgall (commented):
I think this is a great start, but IMO the core of these changes in `useGetAccountName.ts` could still use some cleanup. I don't know quite how to vocalize it, other than "it's still too messy and confusing imo"

src/pages/home/SafeDisplayRow.tsx
adamgall: Is this "fallback" so that if the DAO doesn't have an ENS name set, it will run this other function which grabs the DAO name from the FractalRegistry contract? Seems like that's what's going on here.

src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
adamgall, line 1: I think the changes in here aren't working correctly. Notice how in the Favorites menu the DAO names from the non-currently-connected network aren't loading correctly.

Screenshot 2024-10-17 at 4 08 32 PM

src/pages/home/SafeDisplayRow.tsx
adamgall, line 1: Same bugs are happening in the Home component, too. DAO Names from daos on networks which are not currently connected, are not loading.

adamgall (changes requested):
There are issues with loading dao names on networks which are not currently connected.

This bug exists on dev and prod right now, too.

Screenshot 2024-10-17 at 4 12 16 PM
Screenshot 2024-10-17 at 4 12 26 PM

adamgall (approved):
Approving, as long as #2417 is merged shortly after this branch, or even directly into this branch.

src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
adamgall: I've fixed this in #2417

src/pages/home/SafeDisplayRow.tsx
adamgall: I've cleaned this up in #2417

adamgall (changes requested):
@DarksightKellar , it seems like the scope of this PR maybe grew a bit further than it needed to, but then not far enough.

Is it possible to "undo" the changes in the PR which are _not_ related to simply allowing users to _input_ ENS names into the various Role forms?

The PR #2417 I opened up finishes up the work that you attempted (but was still buggy) in this PR, which managed _displaying_ the proper text for addresses.

DarksightKellar:
> changes in the PR which are _not_ related to simply allowing users to _input_ ENS names into

@adamgall Isn't that what your PR is doing though? I don't see the point of "undoing" the extra stuff, when #2417 also builds on that extra stuff. #2417 itself is more of a revert of the most recent `useGetAccountName` implementation in favour of the existing `useDisplayName`. 

I'd think we just merge it into this, and go from there. Otherwise not sure what the ask is here.

adamgall (approved):

PR #2415, opened 2024-10-16
Reduce ethers

adamgall:
Reduce and consolidate the usage of ethers to the bare minimum. Only thing blocking us from removing it entirely is our usage of `@snapshot-labs/snapshot.js`, which requires an instance of `ethers.Providers.Web3Provider` into its `vote` function.


src/hooks/DAO/proposal/useCastSnapshotVote.ts
adamgall, line 1: Ok so another thing I did in this PR was separate the `useCastSnapshotVote` stuff out of the `useCastVote` file. Didn't really see a need for those to be combined into one hook, they're pretty independent.

src/hooks/DAO/proposal/useCastVote.ts
adamgall, line 1: See, all the Snapshot stuff was pulled out of this file.

src/hooks/stake/lido/useLidoStaking.ts
adamgall, line 1: Thankfully it was pretty easy to remove our usage of `ethers` from the Lido hooks. We were only using the `lido` sdk for getting back TypeChain instance of Lido contracts, then all we were doing with those was encoding some function data for our proposals. So it was nice and easy to replace that function encoding with standard `viem` functions by simply importing the contract ABIs manually into the project, then I was able to remove the entire `lido-sdk` dependency.

src/components/pages/DAOTreasury/hooks/useTreasuryLidoInteractions.ts
adamgall, line 1: Same as the other Lido comment.

src/components/ui/modals/SendAssetsModal.tsx
```diff
@@ -73,8 +73,12 @@ export function SendAssetsModal({
...
   const handleSendAssetsSubmit = async (values: SendAssetsFormValues) => {
     let destAddress = values.destinationAddress;
-    if (validateENSName(values.destinationAddress) && signer) {
-      destAddress = await signer.resolveName(values.destinationAddress);
+    if (validateENSName(values.destinationAddress) && publicClient) {
+      const ensAddress = await publicClient.getEnsAddress({ name: values.destinationAddress });
+      if (ensAddress === null) {
+        throw new Error('Invalid ENS name');
+      }
+      destAddress = ensAddress;
```
adamgall, line 81: We were using an ethers instance to do an ENS name lookup here, but there's a similar function on viem's public client

src/providers/Ethers/hooks/useEthersSigner.ts
adamgall, line 1: In here, instead of pulling the signer data from the Provider and Context which wrapped the whole app, just create an instance of ethers signer on demand. We only use it in one hook now, so no need to provide it to the whole dang app. You'll see that there's only one instance of / reference to `useEthersSigner` in the codebase now.

src/providers/Providers.tsx
```diff
@@ -30,19 +29,17 @@ export default function Providers({ children }: { children: ReactNode }) {
...
+                  <Toaster
+                    position="bottom-center"
+                    richColors
+                    pauseWhenPageIsHidden
+                    theme="dark"
+                    closeButton
+                    toastOptions={{ className: 'sonner-toast' }}
+                  />
+                  {children}
+                </AppProvider>
```
adamgall, line 42: Yay don't wrap the whole app in EthersContextProvider

src/components/Proposals/ProposalActions/CastVote.tsx
```diff
@@ -29,15 +30,15 @@ export function CastVote({ proposal }: { proposal: FractalProposal }) {
...
   const {
-    castVote,
     castSnapshotVote,
     handleChangeSnapshotWeightedChoice,
     handleSelectSnapshotChoice,
     selectedChoice,
     snapshotWeightedChoice,
-    castVotePending,
-  } = useCastVote(proposal.proposalId, extendedSnapshotProposal);
+  } = useCastSnapshotVote(extendedSnapshotProposal);
```
adamgall, line 41: This is the only change related to separating useCastVote and useCastSnapshotVote

src/hooks/stake/lido/useLidoStaking.ts
mudrila: Actually, `lido-sdk` dependency is deprecated and our approach of getting Lido contract addresses is kinda invalid now - I've created an issue for that: https://github.com/decentdao/decent-interface/issues/2413

So, if you really wanna make this "right" - I'd suggest migrating to the mentioned new version of SDK, where contract addresses are part of the package - thus we can clean up our network config.

On the other hand - that thing still could be done in a separate scope, so feel free to ignore this comment. I would say current way of refactoring clearly aligns with the PR goal

src/providers/Ethers/hooks/useEthersSigner.ts
mudrila: Do we need the hook at all then? It is pretty small - let's just move the logic into the place where it's been used and call it a day

src/components/Proposals/ProposalActions/CastVote.tsx
mudrila: Wondering if it makes sense to separate `CastVote` and `CastSnapshotVote` then - wdyt?

src/assets/abi/LidoStEthAbi.ts
mudrila, line 1: So if you'll decide to tackle https://github.com/decentdao/decent-interface/issues/2413 in scope of this PR - we won't need these Abis either

src/components/pages/DAOTreasury/hooks/useTreasuryLidoInteractions.ts
```diff
@@ -36,35 +39,38 @@ export default function useTreasuryLidoInteractions() {
...
+        abi: LidoWithdrawalQueueAbi,
+        client: publicClient,
+      });
+
+      // Since we're checking for the single NFT - we can grab first array element
       const claimableStatus = (
-        await withdrawalQueueContract.getWithdrawalStatus([lidoWithdrawelNFT!.tokenId])
-      )[0]; // Since we're checking for the single NFT - we can grab first array element
+        await withdrawalQueueContract.read.getWithdrawalStatus([
+          [BigInt(lidoWithdrawelNFT!.tokenId)],
```
mudrila, line 63: Oopsie - seems like there's my typo from the time I've implemented this 😄 
Could you fix `withdrawel` -> `withdrawal`

mudrila: Also, seems like `!` is redundant

src/hooks/stake/lido/useLidoStaking.ts
adamgall: I did see that it was deprecated, and that's why I was happy to remove it!

However, I don't really see why we'd need to pull in a whole other new lido sdk dependency _just_ to get their (static) contract addresses. What's wrong with our current approach of simply hardcoding those into our various network config files?

src/providers/Ethers/hooks/useEthersSigner.ts
adamgall: I did consider this. Sure, I'll implement it.

src/hooks/DAO/proposal/useCastSnapshotVote.ts
```diff
@@ -0,0 +1,134 @@
...
+  const [snapshotWeightedChoice, setSnapshotWeightedChoice] = useState<number[]>([]);
+
+  const {
+    node: { daoSnapshotENS },
+    readOnly: {
+      user: { address },
+    },
+  } = useFractal();
+
+  const signer = useEthersSigner();
```
mudrila: So yep - I think we can kill `useEthersSigner` and move that logic just here

src/providers/Ethers/hooks/useEthersSigner.ts
adamgall: done

src/hooks/stake/lido/useLidoStaking.ts
mudrila: Not that it is completely wrong - but also that kinda a pattern we're following with Safe dependency.
This will unlock us from enabling Lido for larger set of chains (which we weren't supporting at that time, but we can support now!)
And we could remove Abis from our assets as well

mudrila: But again - that's kinda out of the scope here - so I'm cool with keeping it "as is", but I think we should be doing that eventually.

src/hooks/utils/useSignerOrProvider.ts
mudrila, line 1: Oh my gosh...
It's happening!

src/providers/Ethers/hooks/useEthersSigner.ts
```diff
@@ -1,7 +1,19 @@
-import { useContext } from 'react';
-import { EthersContext } from '..';
+import { providers } from 'ethers';
+import { useMemo } from 'react';
+import { useWalletClient } from 'wagmi';
 
 export function useEthersSigner() {
```
mudrila: DYE AHAHAHAHAHAHAHAHA
BYBY `useEthersSigner` - you won't be missed buddy!

src/components/Proposals/ProposalActions/CastVote.tsx
adamgall: Sure, let me explore that

mudrila (approved):
Almost there...
I feel so hyped sir!
Thank you for this satisfying PR!

src/components/pages/DAOTreasury/hooks/useTreasuryLidoInteractions.ts
adamgall: done

src/utils/snapshotVote.ts
```diff
@@ -0,0 +1,251 @@
...
+}
+
+interface VoteForSigning extends Omit<Vote, 'type' | 'privacy'> {
+  timestamp: bigint;
+  reason: string;
+  app: string;
+  metadata: string;
+}
+
+// Add these type definitions at the top of the file or in a separate types file
```
Da-Colon, line 132: or this one

src/utils/snapshotVote.ts
```diff
@@ -0,0 +1,251 @@
...
+    { name: 'timestamp', type: 'uint64' },
+    { name: 'proposal', type: 'string' },
+    { name: 'choice', type: 'string' },
+    { name: 'reason', type: 'string' },
+    { name: 'app', type: 'string' },
+    { name: 'metadata', type: 'string' },
+  ],
+} as const;
+
+// Now, let's create a union type for all possible vote types
```
Da-Colon, line 98: Probably don't need this comment

src/utils/snapshotVote.ts
Da-Colon, line 1: I think I follow everything here. Is some of this from Snapshot sdk?

adamgall: yes! However, I think i'm going to remove those commits from this PR and push them into a new one. I'm still doing testing and there are enough changes that are _only_ related to this Snapshot stuff that warrant a new PR.

src/utils/snapshotVote.ts
adamgall: Good call

src/utils/snapshotVote.ts
adamgall: @Da-Colon ok done. Please re-review. I'll get the snapshot-sdk removal into a new PR.

src/components/Proposals/ProposalActions/CastVote.tsx
adamgall: This was going to be a bigger effort than I was hoping. Will save for a later PR.

src/utils/snapshotVote.ts
adamgall: Ended up removing this new utils file from this PR. implemented in #2416 instead.

src/utils/snapshotVote.ts
adamgall: Ended up removing this new utils file from this PR. implemented in #2416 instead.

adamgall:
@Da-Colon please re-review.

Da-Colon (approved):

PR #2417, opened 2024-10-17
2411 fix wearer ens continued

adamgall:
Fixes on top of the base PR #2414 that @DarksightKellar is working on:

- Renames `useGetAccountName` to `useGetSafeName`. This hook is _only for getting Safe / DAO names_. We have a different hook, `useDisplayName`, for getting ENS names of accounts.
- This `useGetSafeName` hook will, in order, attempt to get and return the name for a safe: 1) by ENS name of the Safe Address, 2) by FractalRegistry name of the Safe address, 3) by truncating the address
- This `useGetSafeName.ts` file now only exports one function instead of three. Much cleaner. We can use it everywhere we need to get a Safe name, whether on the current network or not (as in the case of Favorites which might be on any network).
- Small cleanups to the `useDisplayName` hook, to remove unused parameters
- Small improvement to `useNetworkConfig`, to allow passing in an optional `networkId` to get contract information of any network (used in `useGetAccountName` to get the correct `FractalRegistry` address if the network wanted is not currently connected network, as in the case of Favorites).

### DIFF
--- a/src/components/DaoCreator/formComponents/AzoriusNFTDetail.tsx
+++ b/src/components/DaoCreator/formComponents/AzoriusNFTDetail.tsx
@@ -27,7 +27,7 @@ export default function AzoriusNFTDetail({
   const publicClient = usePublicClient();
   const { data: walletClient } = useWalletClient();
 
-  const { displayName } = useDisplayName(tokenDetails?.address, true);
+  const { displayName } = useDisplayName(tokenDetails?.address);
 
   useEffect(() => {
     const loadNFTDetails = async () => {
--- a/src/components/pages/Roles/RoleCard.tsx
+++ b/src/components/pages/Roles/RoleCard.tsx
@@ -1,13 +1,10 @@
 import { Box, Flex, Icon, Image, Text } from '@chakra-ui/react';
 import { CaretCircleRight, CaretRight } from '@phosphor-icons/react';
 import { formatDuration, intervalToDuration } from 'date-fns';
-import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Address, getAddress, zeroAddress } from 'viem';
 import useAvatar from '../../../hooks/utils/useAvatar';
-import { useGetSafeName } from '../../../hooks/utils/useGetSafeName';
-import { useNetworkConfig } from '../../../providers/NetworkConfig/NetworkConfigProvider';
-import { getChainIdFromPrefix } from '../../../utils/url';
+import useDisplayName from '../../../hooks/utils/useDisplayName';
 import { Card } from '../../ui/cards/Card';
 import EtherscanLink from '../../ui/links/EtherscanLink';
 import Avatar from '../../ui/page/Header/Avatar';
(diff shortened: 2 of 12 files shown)


src/components/DaoCreator/formComponents/AzoriusNFTDetail.tsx
```diff
@@ -27,7 +27,7 @@ export default function AzoriusNFTDetail({
   const publicClient = usePublicClient();
   const { data: walletClient } = useWalletClient();
 
-  const { displayName } = useDisplayName(tokenDetails?.address, true);
+  const { displayName } = useDisplayName(tokenDetails?.address);
```
adamgall, line 30: `truncate` defaults to `true`, don't need to specify it.

src/components/pages/Roles/RoleCard.tsx
```diff
@@ -22,11 +20,8 @@ export function AvatarAndRoleName({
   name?: string;
   paymentsCount?: number;
 }) {
-  const { addressPrefix } = useNetworkConfig();
-  const { accountName } = useGetAccountName({
-    address: getAddress(wearerAddress || zeroAddress),
-    chainId: getChainIdFromPrefix(addressPrefix),
-  });
+  const { displayName } = useDisplayName(wearerAddress);
+
```
adamgall, line 24: In the `RoleCard` we are loading end user address names, not Safe names. So, use the `useDisplayName` hook here.

src/components/pages/Roles/RolesDetailsDrawer.tsx
```diff
@@ -61,13 +61,7 @@ export default function RolesDetailsDrawer({
...
     useAddress(roleHatWearer);
 
-  const { getAccountName } = useGetAccountNameDeferred();
-  const [accountDisplayName, setAccountDisplayName] = useState(roleHatWearer);
-  useEffect(() => {
-    if (!!roleHatWearerAddress) {
-      getAccountName(roleHatWearerAddress).then(setAccountDisplayName);
-    }
-  }, [getAccountName, roleHatWearerAddress]);
+  const { displayName } = useDisplayName(roleHatWearerAddress);
```
adamgall, line 64: In `RolesDetailsDrawer` we are getting user names, not Safe names.

src/components/pages/Roles/RolesTable.tsx
```diff
@@ -94,9 +93,8 @@ function RoleNameEditColumn({
 }
 
 function MemberColumn({ wearer }: { wearer: string | undefined }) {
-  const { chain } = useNetworkConfig();
   const { address } = useAddress(wearer || zeroAddress);
-  const { displayName: accountDisplayName } = useDisplayName(address || null, true, chain.id);
+  const { displayName: accountDisplayName } = useDisplayName(address);
```
adamgall, line 97: All of these params were just unnecessary

src/components/ui/menus/SafesMenu/SafeMenuItem.tsx
```diff
@@ -38,23 +30,15 @@ export function SafeMenuItem({ address, network }: SafeMenuItemProps) {
     },
   });
 
-  const { getAccountName } = useGetAccountNameDeferred();
+  const { getSafeName } = useGetSafeName(getChainIdFromPrefix(network));
```
adamgall, line 33: This is a Favorites item, so we should specify the network

src/pages/home/SafeDisplayRow.tsx
```diff
@@ -26,22 +24,17 @@ export function SafeDisplayRow({ address, network, onClick, showAddress }: SafeM
     },
   });
 
-  const publicClient = usePublicClient();
+  const { getSafeName } = useGetSafeName(getChainIdFromPrefix(network));
```
adamgall, line 27: Same here.. This component is used for displaying Favorites, so need to specify the networkId

src/hooks/utils/useGetSafeName.ts
DarksightKellar, line 1: From my experience testing, the `-deferred` variant of this hook (which has been removed in this PR) was the thing to use when the regular version led to unnecessary multiple re-renders. I wonder if that is still the case here

DarksightKellar (approved):

PR #2419, opened 2024-10-18
…

adamgall:
Don't even attempt to show "display name" in member field when editing role with non-ENS member


DarksightKellar (approved):
lgtm

PR #2420, opened 2024-10-20
Roles types tightening

adamgall:


adamgall:
It's going to be terrible to deal with these merge conflicts, and with the new Termed Role functionality it's going to get even worse. Closing because of those reasons plus also this was kind of experimental in the first place. Will re-approach at a later time.

PR #2431, opened 2024-10-27
`Whitelisting` Multiple voting strategies support

mudrila:
- Support multiple voting strategies
- Read whether role is whitelisted for proposals creation
- Add UI control to toggle role whitelisting (only UI element so far)

### DIFF
--- a/src/components/pages/Roles/forms/RoleFormInfo.tsx
+++ b/src/components/pages/Roles/forms/RoleFormInfo.tsx
@@ -1,5 +1,4 @@
 import { Box, Flex, FormControl, Switch, Text } from '@chakra-ui/react';
-import { Field, FieldProps, useFormikContext } from 'formik';
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { DETAILS_BOX_SHADOW } from '../../../../constants/common';
@@ -48,11 +47,11 @@ export default function RoleFormInfo() {
       >
         <FormControl>
           <Field name="roleEditing.name">
-            {({ field, form: { setFieldTouched }, meta }: FieldProps<string, RoleFormValues>) => (
+            {({ field, form: { setFieldTouched }, meta }) => (
               <InputComponent
-                value={field.value}
+                value={field.value || ''}
                 onChange={e => {
-                  setFieldValue(field.name, e.target.value);
+                  setFieldValue('roleEditing.wearer', e.target.value);
                 }}
                 onBlur={() => {
                   setFieldTouched('roleEditing.wearer', true);
@@ -74,11 +73,11 @@ export default function RoleFormInfo() {
         </FormControl>
         <FormControl>
           <Field name="roleEditing.description">
-            {({ field, form: { setFieldTouched }, meta }: FieldProps<string, RoleFormValues>) => (
+            {({ field, form: { setFieldTouched }, meta }) => (
               <TextareaComponent
-                value={field.value}
+                value={field.value || ''}
                 onChange={e => {
-                  setFieldValue(field.name, e.target.value);
+                  setFieldValue('roleEditing.description', e.target.value);
                 }}
                 isRequired
                 gridContainerProps={{
@@ -90,7 +89,7 @@ export default function RoleFormInfo() {
                 textAreaProps={{
                   h: '12rem',
                   onBlur: () => {
-                    setFieldTouched(field.name, true);
+                    setFieldTouched('roleEditing.description', true);
                   },
                 }}
                 label={t('roleDescription')}
@@ -101,7 +100,7 @@ export default function RoleFormInfo() {
         </FormControl>
         <FormControl>
           <Field name="roleEditing.wearer">
-            {({ field, form: { setFieldTouched }, meta }: FieldProps<string, RoleFormValues>) => (
+            {({ field, form: { setFieldTouched }, meta }) => (
               <LabelWrapper
                 label={t('member')}
                 errorMessage={meta.touched && meta.error ? meta.error : undefined}
@@ -111,12 +110,12 @@ export default function RoleFormInfo() {
                 <AddressInput
                   value={displayName ?? field.value}
                   onBlur={() => {
-                    setFieldTouched(field.name, true);
+                    setFieldTouched('roleEditing.wearer', true);
                   }}
                   onChange={e => {
                     const inputWearer = e.target.value;
                     setRoleWearerString(inputWearer);
-                    setFieldValue(field.name, inputWearer);
+                    setFieldValue('roleEditing.wearer', inputWearer);
                   }}
                 />
               </LabelWrapper>
@@ -126,7 +125,7 @@ export default function RoleFormInfo() {
       </Box>
       <FormControl mt={4}>
         <Field name="roleEditing.canCreateProposals">
-          {({ field }: FieldProps<string, RoleFormValues>) => (
+          {() => (
             <Flex
               justifyContent="space-between"
               alignItems="center"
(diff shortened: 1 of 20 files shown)


DarksightKellar (approved):

src/components/pages/Roles/forms/RoleFormInfo.tsx
```diff
@@ -31,96 +31,126 @@ export default function RoleFormInfo() {
...
+                  }}
+                />
+              </LabelWrapper>
+            )}
+          </Field>
+        </FormControl>
+      </Box>
+      <FormControl mt={4}>
+        <Field name="roleEditing.canCreateProposals">
+          {() => (
```
mudrila, line 128: Thought this rendering wouldn't make sense - but then ts complains 🤷🏼

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
```diff
@@ -71,13 +71,18 @@ export function ManageDAOMenu({ parentAddress, freezeGuard, guardContracts }: IM
...
         if (node?.fractalModules) {
           let result = GovernanceType.MULTISIG;
-          const votingContractAddress = await getVotingStrategyAddress();
-          if (votingContractAddress) {
-            const masterCopyData = await getZodiacModuleProxyMasterCopyData(votingContractAddress);
-
-            if (masterCopyData.isLinearVotingErc20) {
+          const votingStrategies = await getVotingStrategies();
+          if (votingStrategies) {
+            const latestVotingStrategy = votingStrategies[votingStrategies.length - 1];
```
adamgall, line 76: Why do we grab the "latest" voting strategy here? It seems like our business requirements are changing so that we can expect an azorious contract to potentially have multiple strategies enabled, but why are we just grabbing the "last" one here?

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
adamgall, line 1: In this file, the only place we actually ever read `governanceType` is we check if it's a multisig or not. I think you can simplify the code in this file by changing the state 

from:
`const [governanceType, setGovernanceType] = useState(GovernanceType.MULTISIG);`

to:
`const [isMultisig, setIsMultisig] = useState(true);`

src/hooks/utils/useMasterCopy.ts
```diff
@@ -79,31 +82,13 @@ export function useMasterCopy() {
...
           }
-
-          // setValue(
-          //   {
-          //     cacheName: CacheKeys.MASTER_COPY,
-          //     chainId: chain.id,
-          //     proxyAddress,
-          //   },
-          //   masterCopyAddress,
-          // );
```
adamgall, line 106: Actually I think... we can uncomment this code instead of deleting it. @Da-Colon thoughts?

src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -107,10 +116,13 @@ const useHatsTree = () => {
...
             hatsTree: treeWithFetchedDetails,
             chainId: BigInt(contextChainId),
-            hatsProtocol: hatsProtocol,
+            hatsProtocol,
             erc6551Registry,
             hatsAccountImplementation,
             publicClient,
+            whitelistingVotingStrategy:
+              linearVotingErc20WithHatsWhitelistingAddress ||
+              linearVotingErc721WithHatsWhitelistingAddress,
```
adamgall, line 125: what if both of these are undefined/null? Is that cool?

src/hooks/DAO/proposal/useSubmitProposal.ts
```diff
@@ -60,13 +60,21 @@ export default function useSubmitProposal() {
...
+    governanceContracts: {
+      linearVotingErc20Address,
+      linearVotingErc20WithHatsWhitelistingAddress,
+      linearVotingErc721Address,
+      linearVotingErc721WithHatsWhitelistingAddress,
+    },
     action,
+    readOnly: {
+      user: { address },
+    },
```
adamgall, line 77: Instead of this silly piece of state can we just pull the address from `useAccount`?

src/components/pages/Roles/forms/RoleFormInfo.tsx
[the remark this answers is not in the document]
mudrila: Tried removing that as TS yells at me 🥲

src/components/pages/Roles/forms/RoleFormInfo.tsx
```diff
@@ -31,96 +31,126 @@ export default function RoleFormInfo() {
...
+                  setFieldValue(
+                    'roleEditing.canCreateProposals',
+                    !values.roleEditing?.canCreateProposals,
+                  )
+                }
+                isChecked={values.roleEditing?.canCreateProposals}
+                isDisabled={!values.roleEditing}
               />
-            </LabelWrapper>
+            </Flex>
```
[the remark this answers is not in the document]
mudrila: There're no stupid questions sir 🙃

So, the reason why we're collecting these for all the roles is because this single switch is the actual place where we're entering the whole complex logic behind deploying, enabling and managing whitelisting strategy. 

There's no UI to configure that new voting strategy in any other places - only this single switch on each role. 

This decision simplifies UX for the user while we can implement more complex UI / UX later on.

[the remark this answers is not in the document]
mudrila: And if the DAO is not using whitelisting yet - we just not running the request but simply falling back to "false"

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
mudrila: It is sort of "what was installed latest - that is most important" vision. It could be first voting strategy - doesn't really matter at this stage of "multiple voting strategies" implementation

src/hooks/DAO/loaders/useHatsTree.ts
mudrila: Yep - just not running the requests to check whether hat is whitelisted but simply putting switch to off

src/hooks/DAO/proposal/useSubmitProposal.ts
mudrila: Yes - and at some point I've started removing this readOnly thingy as a whole - but then realised it will be better to do so in a scope of separate PR

mudrila: Removed this usage to at least not spread plague 🤣

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
mudrila: There was so much leftovers from times when `ManageDAOMenu` was actually used on hierarchy page - now that I've got rid of it it all became so simpler. Check that out sir

src/components/ui/menus/ManageDAO/ManageDAOMenu.tsx
mudrila: Now this code is removed completely

src/hooks/utils/useMasterCopy.ts
mudrila: At least I'll add commented code back and we could decide later on

mudrila: Done ser

src/components/pages/Roles/forms/RoleFormInfo.tsx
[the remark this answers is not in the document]
adamgall: ok makes sense, thanks for clarifying!

adamgall (approved):

PR #2433, opened 2024-10-28
`Whitelisting` voting strategy deployment

mudrila:


DarksightKellar (approved):
Code LGTM!

src/store/roles/rolesStoreUtils.ts
```diff
@@ -137,6 +137,19 @@ export const initialHatsStore: RolesStoreData = {
...
+
+  // Each next level is next **16 bits**
+  // Since we're operating only with direct child of top level admin - we don't care about nested levels
+  // @dev At least for now?
+  const newSiblingId = (hatsCount + 1).toString(16).padStart(4, '0');
+
+  // Total length of Hat ID is **32 bytes** + 2 bytes for 0x
+  return BigInt(`${adminLevelBinary}${newSiblingId}`.padEnd(66, '0'));
+};
+
```
adamgall, line 152: Hey wait... if we're adding this back, does that mean that there are places in the new code where we are reintroducing race conditions, which should actually go into contract logic?

cc: @DarksightKellar

src/hooks/utils/useCreateRoles.ts
```diff
@@ -64,6 +90,128 @@ export default function useCreateRoles() {
...
+          parseAbiParameters(
+            'address, address, address, uint32, uint256, uint256, address, uint256[]',
+          ),
+          [
+            daoAddress, // owner
+            votesToken.address, // governance token
+            moduleAzoriusAddress, // Azorius module
+            Number(votingStrategy.votingPeriod.value),
+            (votingStrategy.quorumPercentage.value * quorumDenominator) / 100n, // quorom numerator, denominator is 1,000,000, so quorum percentage is quorumNumerator * 100 / quorumDenominator
+            500000n, // basis numerator, denominator is 1,000,000, so basis percentage is 50% (simple majority)
```
adamgall, line 124: Are we making this call for the users? Does this need to be captured in a form?

src/hooks/utils/useCreateRoles.ts
```diff
@@ -64,6 +90,128 @@ export default function useCreateRoles() {
...
+            'address, address[], uint256[], address, uint32, uint256, uint256, address, uint256[]',
+          ),
+          [
+            daoAddress, // owner
+            erc721Tokens.map(token => token.address), // governance tokens addresses
+            erc721Tokens.map(token => token.votingWeight), // governance tokens weights
+            moduleAzoriusAddress, // Azorius module
+            Number(votingStrategy.votingPeriod.value),
+            (votingStrategy.quorumThreshold.value * quorumDenominator) / 100n, // quorom numerator, denominator is 1,000,000, so quorum percentage is quorumNumerator * 100 / quorumDenominator
+            500000n, // basis numerator, denominator is 1,000,000, so basis percentage is 50% (simple majority)
```
adamgall, line 173: Same question

src/hooks/utils/useCreateRoles.ts
```diff
@@ -490,6 +649,14 @@ export default function useCreateRoles() {
...
         if (formHat.editedRole.status === EditBadgeStatus.New) {
           allTxs.push(...(await prepareNewHatTxs(formHat)));
+          newHatsCount++;
+          if (formHat.canCreateProposals) {
+            const newHatId = predictHatId({
+              adminHatId: hatsTree.adminHat.id,
+              hatsCount: hatsTree.roleHats.length + newHatsCount,
+            });
+            whitelistingPermissionAddedHats.push(newHatId);
+          }
```
adamgall, line 659: Yeah this gives me pause, because it re-introduces some race condition bugs that @DarksightKellar _just_ finished removing.

I think some stuff here needs to move into a smart contract. Not sure what though...

adamgall (commented):
For the most part looks good, but I want to talk more about the re-introduction of "predicting hat IDs". Seems like that would be re-introducing the race condition issues when there are multiple proposals active at the same time that add Roles.

src/store/roles/rolesStoreUtils.ts
mudrila: Yes, that's right - and I don't think we really-really wanna address this for this use case

src/hooks/utils/useCreateRoles.ts
mudrila: At least - we do while creating Safe, it's just the same as we have it rn in `DAOCreator` flow

src/hooks/utils/useCreateRoles.ts
mudrila: Same here - we already doing the same, think that's fine and we shouldn't overcomplicate UI / UX with relatively straightforward decision

src/hooks/utils/useCreateRoles.ts
mudrila: My feeling about it is that it is probably "niche" use-case
How frequently we're gonna have a case when there's multiple proposals for initial roles setup with whitelisting involved? I think this is kinda `WONTFIX` here, as fixing that will involve significantly complicating smart contracts and FE logic to handle this case, though I don't see this case happening for our "target audience". WDYT?

Da-Colon: Tough one. All of it is niche. We could kick the bucket if its alot of work. but since we have the contracts for this purpose in flux might be a good idea to do it now versus later? 

I guess the question would be if this case were to happen, whats the result?

A reverted transaction or incorrect execution?

Da-Colon (commented):
Ooof `useCreateRoles` is going to need a trimming down when all this work gets done. Saving approval for resolved convo

src/hooks/utils/useCreateRoles.ts
mudrila: AFAIK - reverted transaction, since we'll try to whitelist hats that are already whitelisted due to previous proposal being executed

adamgall: How will this significantly complicate the smart contracts?

mudrila: @adamgall For covering all the use-cases on smart contracts side, this will require from us following changes:

1. When Safe doesn't have voting strategy attached - we'll need to deploy voting strategy with initially whitelisted hats through smart contract. Also it should be wrapped in a way where Safe is the owner of the voting strategy
2. When Safe does have voting strategy attached - hat creation and whitelisting should be happening through the module, so that caller of the Voting Strategy contract is actually Safe - coz only Safe will have permissions to whitelist / un-whitelist hats

mudrila: As per discussion - we'll be dealing with this problem in parallel with getting rest of whitelisting merged.
Created an issue for that: https://github.com/decentdao/decent-contracts/issues/122

PR #2437, opened 2024-10-31
`Termed roles` roles form clean up

Da-Colon:
I set out for this work to cover viewing the Role Terms in the Role Details, starting with a refactor of when 'current' and 'next' terms were decided. on refactoring this and I found a few bugs, and then addressed @mudrila comments from the first PR (since I what I was adding was touching some of those anyway)

I had the goal in mind to rework all the PRs and commits but that didn't happen. I would like to work with (at least 2) folks to get these open ones closed today so I can continue from a solid point on.


src/components/pages/Roles/RoleTerm.tsx
```diff
@@ -192,24 +194,26 @@ function RoleTermMemberAddress({ memberAddress }: { memberAddress: Address }) {
           address={memberAddress}
           url={avatarURL}
         />
-        <Flex
-          alignItems="center"
-          gap={2}
+        <Button
```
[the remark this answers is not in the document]
mudrila: But if that'll require complexity adjustments to `AddressCopier` but this specific case needs this "simple as is" implementation - I'm cool keeping it "as is"

src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
```diff
@@ -103,7 +103,13 @@ export default function RoleFormCreateProposal({ close }: { close: () => void })
               termNumber: term.termNumber,
             };
           }) || [];
-
+        const activeTerms = allRoleTerms.filter(term => term.termEndDate > new Date());
```
mudrila, line 106: Can reuse `new Date()` via `const now = new Date();`

src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
```diff
@@ -103,7 +103,13 @@ export default function RoleFormCreateProposal({ close }: { close: () => void })
...
+        const roleTerms = {
+          allTerms: allRoleTerms,
+          currentTerm: activeTerms[0],
+          nextTerm: activeTerms[1],
+          expiredTerms: allRoleTerms.filter(term => term.termEndDate <= new Date()),
```
mudrila, line 111: And then use it here

src/components/pages/Roles/forms/RoleFormTerms.tsx
```diff
@@ -13,21 +13,23 @@ import {
...
+import RoleTerm from '../RoleTerm';
+import { RoleFormTermStatus, RoleFormValues } from '../types';
 
-function RoleTermEndDateInput({ termIndex }: { termIndex: number }) {
+function RoleTermEndDateInput() {
   const { t } = useTranslation('roles');
   return (
     <FormControl>
-      <Field name={`roleEditing.roleTerms[${termIndex}].termEndDate`}>
+      <Field name="newRoleTerm.termEndDate">
```
mudrila, line 32: Oh yeah, I love it!

src/components/pages/Roles/forms/RoleFormTerms.tsx
```diff
@@ -84,7 +86,37 @@ function RoleTermMemberInput({ termIndex }: { termIndex: number }) {
...
+        nomineeAddress = ens;
+      }
+    }
+
+    setFieldValue('roleEditing.roleTerms', [
+      ...(values?.roleEditing?.roleTerms || []),
+      {
+        nominee: nomineeAddress,
+        termEndDate: values.newRoleTerm.termEndDate,
+        termNumber: termIndex + 1,
```
mudrila, line 114: I'd suggest renaming `termNumber` to `termIndex`, but that's rather personal styling preference - feel free to keep `termNumber` if you prefer it

src/components/pages/Roles/forms/RoleFormTerms.tsx
```diff
@@ -243,74 +270,82 @@ function RoleTermExpiredTerms({
...
 export default function RoleFormTerms() {
-  const [newTermIndex, setNewTermIndex] = useState<number>();
   const { t } = useTranslation('roles');
-  const { values } = useFormikContext<RoleFormValues>();
+  const { values, setFieldValue } = useFormikContext<RoleFormValues>();
+  const { getHat } = useRolesStore();
+
+  const roleHatTerms = useMemo(() => {
+    const roleFormHatId = values.roleEditing?.id;
+    const hat = getHat(roleFormHatId || '0x');
```
mudrila, line 279: I feel like we usually trying to be more explicit about returning `undefined` and to not use these mock values like `0x` or `zeroAddress`
Though I don't have strong stand here - I'm personally ok-ish with approach of implicit `undefined` return

mudrila (approved):
Bunch of minor comments, but otherwise looks good!

DarksightKellar (approved):
lgtm

src/components/pages/Roles/forms/RoleFormTerms.tsx
DarksightKellar: I thought the "number" in `termNumber` was significant, as in, we can't have a term number 0, but we can have a 0-index?

src/components/pages/Roles/forms/RoleFormTerms.tsx
DarksightKellar: If possible it'd better to handle the "problem" that `roleFormHatId` is undefined (ie, throw, make sure it's defined before this execution point, etc), instead of relying on defaults, since we're currently tightening up our typing app-wide. Not sure what the context of this particular line is though

src/components/pages/Roles/RoleTerm.tsx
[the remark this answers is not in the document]
Da-Colon: yeah its currently tied to the `CeleryButtonWithIcon` component

src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
Da-Colon: Changed this code. So only one use now

src/components/pages/Roles/forms/RoleFormCreateProposal.tsx
Da-Colon: Changed this code. So only one use Date()

src/components/pages/Roles/forms/RoleFormTerms.tsx
Da-Colon: yeah exactly. instead of having to constantly add one I do it at the beginning, plus its right there on the current/next objects so I can access it without having to look back at which index it is.

src/components/pages/Roles/forms/RoleFormTerms.tsx
Da-Colon: Current. I have since updated this. and is handled slightly differently. Sorry hadn't pushed up the latest and left over pseudo code.

PR #2441, opened 2024-11-01
Implement tooltip design update

DarksightKellar:
All tooltips across the app should now look like this:

Screenshot 2024-11-01 at 17 33 46

@decentdao/engineering 
I couldn't use the tooltip theme to add the tooltip arrow to all the tooltips -- it simply refused to work. Also, even when I manually set `hasArrow`, for some reason I _also_ have to set `backgroundColor` _inline_ for the arrow to show. Using `backgroundColor` on the theme alone caused the arrow NOT to show.

So I've elected to create a custom `DecentTooltip` component and have all files import from it instead.

I don't like it, so hoping there's a much less invasive way to do this.

### DIFF
--- a/src/assets/theme/components/tooltip/index.ts
+++ b/src/assets/theme/components/tooltip/index.ts
@@ -4,10 +4,11 @@ const $arrowBg = cssVar('popper-arrow-bg');
 
 export default {
   baseStyle: {
+    maxW: '20rem',
     borderRadius: '4px',
-    backgroundColor: 'black-0',
+    backgroundColor: 'neutral-9',
     padding: '0.25rem 0.5rem',
-    color: 'white-0',
+    color: 'black-0',
     [$arrowBg.variable]: 'black-0',
   },
 };
--- a/src/components/Proposals/MultisigProposalDetails/TxActions.tsx
+++ b/src/components/Proposals/MultisigProposalDetails/TxActions.tsx
@@ -1,4 +1,4 @@
-import { Box, Button, Text, Flex, Tooltip } from '@chakra-ui/react';
+import { Box, Button, Text, Flex } from '@chakra-ui/react';
 import { abis } from '@fractal-framework/fractal-contracts';
 import { useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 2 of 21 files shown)


mudrila (approved):

src/assets/theme/components/tooltip/index.ts
```diff
@@ -4,10 +4,11 @@ const $arrowBg = cssVar('popper-arrow-bg');
 
 export default {
   baseStyle: {
+    maxW: '20rem',
     borderRadius: '4px',
-    backgroundColor: 'black-0',
+    backgroundColor: 'neutral-9',
     padding: '0.25rem 0.5rem',
-    color: 'white-0',
+    color: 'black-0',
```
adamgall, line 11: Are these necessary? Looks like you also set them in the DecentTooltip component.

src/assets/theme/components/tooltip/index.ts
```diff
@@ -4,10 +4,11 @@ const $arrowBg = cssVar('popper-arrow-bg');
...
     [$arrowBg.variable]: 'black-0',
```
adamgall, line 12: What is this property?

adamgall:
@DarksightKellar i just spent a few minutes also digging into the weirdness you described, and I experienced the same. Couldn't make any progress doing much tinkering, so yeah, seems like this is the answer from my PoV unless @Da-Colon or @mudrila can figure it out.

But in regards to this PR, can probably remove some redundant "base styles" now.

adamgall (changes requested):
New review request:

just delete the Tooltip theme index file entirely

adamgall:
@DarksightKellar is there an associated issue with this PR?

DarksightKellar:
> @DarksightKellar is there an associated issue with this PR?

Not one that I can find. cc @xanaramoss

src/assets/theme/components/tooltip/index.ts
DarksightKellar: Removed

src/assets/theme/components/tooltip/index.ts
DarksightKellar: Nope wasn't needed. Removed.

adamgall:
> > @DarksightKellar is there an associated issue with this PR?
> 
> Not one that I can find. cc @xanaramoss

Ah ok, i thought you picked this up because of an outstanding issue that you remembered was assigned to you.

xanaramoss:
> > @DarksightKellar is there an associated issue with this PR?
> 
> Not one that I can find. cc @xanaramoss


I didn't create any issue related to this as far as I remember. 
The one I mentioned on Stantup was. [issue 2432](https://github.com/decentdao/decent-interface/issues/2432)

adamgall (changes requested):
Need to delete the Tooltip import which no longer exists

adamgall (approved):

adamgall (approved):

PR #2447, opened 2024-11-02
`[BUGFIX]` rename getStreamsWithFundsToClaimFromFromHat -> getStreamsWithFundsToClaimNotCancelling

### LINKED ISSUE #2446: `getStreamsWithFundsToClaimFromFormHat` and `getStreamsWithFundsToClaimFromFromHat`
Currently on develop, `useCreateRoles` has two functions defined:

`getStreamsWithFundsToClaimFromFormHat`
`getStreamsWithFundsToClaimFromFromHat`

If you'll look closely, one of them seems to be misspelled, which is why they're both able to be defined.

Even more unfortunately, they're both being used in the code.

Even more unfortunately, there is a slight difference in their implementation.

Da-Colon:
Closes #2446 

So looking into it. 

- `getStreamsWithFundsToClaimFromFromHat`: 
  - is used during `member change` action. 
  - filters for `withdrawableAmount` and `isCancelling` (which indicates the user has set this payment to be cancelled)

- `getStreamsWithFundsToClaimFromFormHat`:
  - is used during `role removed` action
  
  Looking into it. When the role is removed. all streams are cancelled and flushed. so the user can't add in additionally to cancel individual streams. In the case of `member change` we track that if a stream is marked for cancelling that it is ignored  when looking at what streams to flush during that section and handled in the `payments` section

### DIFF
--- a/src/hooks/utils/useCreateRoles.ts
+++ b/src/hooks/utils/useCreateRoles.ts
@@ -411,7 +411,7 @@ export default function useCreateRoles() {
     [prepareBatchLinearStreamCreation],
   );
 
-  const getStreamsWithFundsToClaimFromFromHat = useCallback((formHat: RoleHatFormValueEdited) => {
+  const getStreamsWithFundsToClaimNotCancelling = useCallback((formHat: RoleHatFormValueEdited) => {
     return (formHat.payments ?? []).filter(
       payment => (payment?.withdrawableAmount ?? 0n) > 0n && !payment.isCancelling,
     );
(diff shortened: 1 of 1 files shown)


adamgall (changes requested):
Thanks for looking into this.

Can we make both names a bit more descriptive then? Maybe

- getRoleChangedStreamsWithFundsToClaim
- getRoleRemovedStreamsWithFundsToClaim

Da-Colon:
> Thanks for looking into this.
> 
> Can we make both names a bit more descriptive then? Maybe
> 
> * getRoleChangedStreamsWithFundsToClaim
> * getRoleRemovedStreamsWithFundsToClaim
OoO good call.

adamgall (approved):

mudrila (approved):

DarksightKellar (approved):

PR #2450, opened 2024-11-05
BigInt serialization fix

mudrila:
- Fix serializing `BigInt`
- Re-map `proposalTemplates` `ethValue` from serialized string to unserialized `BigIntValuePair`


src/main.tsx
```diff
@@ -10,6 +10,17 @@ import { useNetworkConfig } from './providers/NetworkConfig/NetworkConfigProvide
...
+  interface BigInt {
+    /** Convert BigInt to string form in JSON.stringify */
+    toJSON: () => string;
+  }
+}
+
+BigInt.prototype.toJSON = function () {
+  return this.toString();
+};
+
```
adamgall, line 23: Is there a better or more appropriate spot for this code?

src/hooks/DAO/loaders/useFractalGovernance.ts
adamgall, line 1: @mudrila can you just explain the changes in here? Why so much new code?

adamgall:
Also, do we need to target `main` for this one?

src/main.tsx
mudrila: Yeah, I think there should be better place - let me play around

src/hooks/DAO/loaders/useFractalGovernance.ts
mudrila: Little bit DRY-ing setting empty proposal templates and adding a mapping to convert string-based `ethValue` into `bigint`-based `BigIntValuePair`

src/main.tsx
mudrila: Check dat out pls

src/hooks/DAO/loaders/useFractalGovernance.ts
mudrila: Although it might be done with less `if..else...if...else` - check out updated implementation sir :)

DarksightKellar (approved):

mudrila:
> @mudrila I added a couple of commits to this branch. Let me know if you agree with them please

Yes, looks good - though I have had it kinda in the same way but reverse order and more if... else... - but I definitely like this more.

PR #2451, opened 2024-11-05
Update fractal contracts package

adamgall:
Brings in the new version of `@fractal-framework/fractal-contracts`, which includes a number of breaking changes, which are fixed in this PR:

- ZodiacModuleProxyFactory (neither address nor ABI) is no longer exported from `fractal-contracts`
  - This PR adds the ABI as an import in `src/assets/abis`, and removes the `zodiacModuleProxyFactoryOld` property from Network Config
- Cleanup: removes an unused reference from a `useEffect` dependency
- Adds the `@hatsprotocol/modules-sdk` package, to get access to the "module factory" deployment address
- Updates names of some existing contract addresses in NetworkConfig, and add a few more
  - `decentHatsMasterCopy` variable has been broken up into `decentHatsCreationModule` and `decentHatsModificationModule` variables
    - !!! It's not a "Master Copy". "MasterCopy" nomenclature is reserved for contracts which we create proxy copies of !!!
  - `decentSablierMasterCopy` variable has been renamed to `decentSablierStreamManagementModule`
    - !!! It's not a "Master Copy". "MasterCopy" nomenclature is reserved for contracts which we create proxy copies of !!!
  - New contract addresses: `decentAutonomousAdminV1MasterCopy`, `hatsElectionsEligibilityMasterCopy`

### DIFF
--- a/src/providers/NetworkConfig/networks/mainnet.ts
+++ b/src/providers/NetworkConfig/networks/mainnet.ts
@@ -78,14 +78,19 @@ export const mainnetConfig: NetworkConfig = {
 
     claimErc20MasterCopy: getAddress(a.ERC20Claim),
 
+    decentAutonomousAdminV1MasterCopy: getAddress(a.DecentAutonomousAdminV1),
+
     fractalRegistry: getAddress(a.FractalRegistry),
     keyValuePairs: getAddress(a.KeyValuePairs),
-    decentHatsMasterCopy: getAddress(a.DecentHats_0_1_0),
-    decentSablierMasterCopy: getAddress(a.DecentSablierStreamManagement),
+
+    decentHatsCreationModule: getAddress(a.DecentHatsCreationModule),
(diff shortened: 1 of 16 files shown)


src/providers/NetworkConfig/networks/mainnet.ts
```diff
@@ -58,7 +58,6 @@ export const mainnetConfig: NetworkConfig = {
     ),
 
     zodiacModuleProxyFactory: '0x000000000000aDdB49795b0f9bA5BC298cDda236',
-    zodiacModuleProxyFactoryOld: getAddress(a.ModuleProxyFactory),
```
Da-Colon, line 61: This is okay to remove now?

Da-Colon (approved):
look good

src/providers/NetworkConfig/networks/mainnet.ts
adamgall: Um yep that's why I did it

PR #2454, opened 2024-11-06
Fix Shutter not loading on mainnet

adamgall:
Remove DOMAIN_SEPARATOR from `isVotesErc20` and `isVotesErc20Wrapper` fingerprint.

I ran this change through my test repo, and all tests on sepolia still passed. So, this `DOMAIN_SEPARATOR` part of the fingerprint that I removed wasn't required to produce a unique fingerprint.


Da-Colon (approved):
O I see

PR #2462, opened 2024-11-08
Ensure that migrations run before app is mounted

adamgall:
Currently, migrations run "at the same time" that the app is mounting. We actually want them to run before the app mounts, so that the new code in the app which relies on those migrations, actually has that migrated data.


src/main.tsx
```diff
@@ -21,13 +22,19 @@ function DecentRouterProvider() {
...
-if (root !== null) {
-  ReactDOM.createRoot(root).render(
-    <React.StrictMode>
-      <Providers>
-        <DecentRouterProvider />
-      </Providers>
-    </React.StrictMode>,
-  );
+async function initializeApp() {
+  await runMigrations();
```
mudrila, line 26: It does nothing when migration shouldn't be happening, right?

mudrila (approved):
Approved with one question

src/main.tsx
adamgall: I don't understand the question. Migrations should always be ran, if they need to be ran.

DarksightKellar (approved):

PR #2475, opened 2024-11-12
remove no wearer throw error

Da-Colon:
Quick little bugfix when creating a termed role proposal. 

There will be reckoning for this section of code but not today. in the meantime this removes this error. when creating a new role.


adamgall (approved):

DarksightKellar (approved):
"there will be reckoning"
for sure

PR #2483, opened 2024-11-13
No more directly reading anything from FractalRegsistry

adamgall:
Closes #2440 

This PR does two things:

1. Removes unused props (and unused data loaders) from a couple of components.
2. Removes all (only one) direct log reads from FractalRegistry contract.
3. Bumps subgraph version to [v0.1.2](https://github.com/decentdao/decent-subgraph/tree/v0.1.2)

PR #2487, opened 2024-11-13
Fix Safe Name fetching

adamgall:
Screenshot 2024-11-13 at 2 53 26 PM

Fix Safe Name resolution, turns out we were reading those FractalRegistry logs on the home page.

This fixes a bug introduced in #2483


mudrila (approved):

Da-Colon (approved):
works great

PR #2492, opened 2024-11-14
`Termed Roles` Regular Role Conversion to Termed

### LINKED ISSUE #2481: Toggle Existing Role into Termed Role
Currently the code path for converting regular roles into termed roles is incomplete.

Still some work left to do putting together the transaction. If it takes more than tomorrow I'll push up a PR blocking that path.

Da-Colon:
Closes #2481


mudrila (approved):

src/hooks/utils/useCreateRoles.ts
```diff
@@ -941,6 +946,213 @@ export default function useCreateRoles() {
...
+            );
+          }
+
+          const flushStreamTxCalldata = prepareFlushStreamTxs({
+            streamId: stream.streamId,
+            to: formHatCurrentWearer,
+            smartAccount: formHat.smartAddress,
+          });
+
+          streamsWithdrawTx.push(...flushStreamTxCalldata);
```
DarksightKellar, line 1050: Also just wondering if the UI is clearly communicating that changing a role type necessarily leads to flushed funds to the current wearer (whether the nominee is the current wearer or not)

DarksightKellar (approved):
Just a couple UX musings, which I'll now go test myself tbh

src/hooks/utils/useCreateRoles.ts
Da-Colon: O I see...hmmmmm. definitely need to flush since we wouldn't be reading the `SmartAccount` for that hat anymore after the switch maybe something @decentdao/design can think about as users begin to use the feature.

PR #2503, opened 2024-11-15
Mark nominee as pending wearer for the termed role

mudrila:
Screenshot 2024-11-15 at 03 29 56

Note - there's a good chance this logic doesn't cover all the cases of presenting nominee in a pending state for existing roles / edited termed roles.


Da-Colon (changes requested):
We will also need to update the wearerAddress on line 100 in `SafeRolesPage` for mobile

adamgall (approved):

Da-Colon (approved):

PR #2510, opened 2024-11-18
Remove two GovernanceTypes from enum, doesn't seem like they were being used anywhere

adamgall:
@mudrila would especially like your eyes on this one.

The reason why I think this PR is valid is because, the at only places where we're doing:

```ts
action.dispatch({
  type: FractalGovernanceAction.SET_GOVERNANCE_TYPE,
  payload: ...,
});
```

(which are in `useFractalGovernance.ts`)

the only `payload`s that are being used are
- `GovernanceType.AZORIUS_ERC20`
- `GovernanceType.AZORIUS_ERC721`
- `GovernanceType.MULTISIG`

So it seems like the `type` property on the `FractalGovernance` state is never going to be two enum values that I removed in this PR.

### DIFF
--- a/src/components/DaoCreator/hooks/useParentSafeVotingWeight.ts
+++ b/src/components/DaoCreator/hooks/useParentSafeVotingWeight.ts
@@ -21,9 +21,7 @@ export const useParentSafeVotingWeight = () => {
 
     switch (governance.type) {
       case GovernanceType.AZORIUS_ERC20:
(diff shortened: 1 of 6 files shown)


mudrila (approved):
Yeah, that's right - probably some "dead code" I've introduced while working on whitelisting / supporting multiple voting strategies

DarksightKellar (approved):

PR #2511, opened 2024-11-18
Move "isAzorius" from Read Only Values to "governance" state

adamgall:
This PR removes the `dao` object from `readOnlyValues` state. The `dao` object only had one property on it, called `isAzorius`. I've moved that over into global state into the `governance` state, and calculated its value in the reducer action that take the `type` of DAO. if the `type` is one of the Azorius types then this is `true`, else `false`.


mudrila (approved):
Oh my god...
This day has come...
So as I (if ya know what I mean 🤣 😎 )

adamgall:
Ok that's better

src/providers/App/useReadOnlyValues.ts
DarksightKellar, line 1: This guy's gotta be shaking at this point. How much longer will `useReadOnlyValues` stay relevant? Tune in same time next week to find out!

DarksightKellar (approved):

PR #2513, opened 2024-11-18
Move analytics user tracking setup to Global component

adamgall:
Create new `Global` component which does global level stuff -- like setup/teardown user analytics tracking, and set up the Layout.

### DIFF
--- a/src/components/ui/page/Global/index.tsx
+++ b/src/components/ui/page/Global/index.tsx
@@ -1,0 +1,23 @@
+import * as amplitude from '@amplitude/analytics-browser';
+import * as Sentry from '@sentry/react';
+import { useEffect } from 'react';
+import { useAccount } from 'wagmi';
+import { Layout } from '../Layout';
+
+const useUserTracking = () => {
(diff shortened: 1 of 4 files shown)


src/providers/App/useReadOnlyValues.ts
DarksightKellar, line 1: tick.

tock.

src/components/ui/page/Global/index.tsx
DarksightKellar, line 1: Instead of a component (because this is really just a `Layout` but with global level setup), how about something like a `useGlobalSetup.ts` that will contain all global setup code, `useUserTracking` being one of them (albeit out of 1 currently), and imported and called once at or above `Layout`?

DarksightKellar (changes requested):

src/components/ui/page/Global/index.tsx
DarksightKellar: @adamgall  Implemented this, checked for render count, got 20, vs 2 in this approach. Won't update lol.

PR #2555, opened 2024-11-20
Update multisig config copy

### LINKED ISSUE #2539: Create Safe/Configure Safe copy updates
https://www.notion.so/hulking-cosmos-3e8/Q4-2024-QA-Requirements-13e2ee9c5bc280a88458e7a935d17488?pvs=4#1432ee9c5bc2805fa179c40692511069

DarksightKellar:
closes #2539 

## Screenshots
Screenshot 2024-11-20 at 18 41 43

Error copy update here wasn't in the doc (and hopefully not clashing with some other part of the doc), but:
Screenshot 2024-11-20 at 21 55 51


src/i18n/locales/en/daoCreate.json
```diff
@@ -2,19 +2,19 @@
...
   "addNFTButton": "Import another token",
   "labelCreateSubDAOProposal": "Create a child Safe",
-  "errorMinSigners": "Number of signers must be greater than 0",
+  "errorMinSigners": "Number of owners must be greater than 0",
   "errorLowSignerThreshold": "Threshold must be greater than 0",
   "errorHighSignerThreshold": "Threshold is too high",
   "labelSigners": "Total signers",
-  "helperSigners": "The total number of possible signers (max 99)",
-  "labelSigThreshold": "Signature Threshold",
+  "helperSigners": "The addresses of owners to initialize the multisig with.",
```
adamgall, line 9: There is still code which limits the number to <= 99. I assume that should be removed if we're no longer talking about it.

adamgall (changes requested):

src/i18n/locales/en/daoCreate.json
DarksightKellar: Good call

adamgall (approved):

src/i18n/locales/en/daoCreate.json
parkermccurley: Part of this PRD includes removing the text-entry for "Number of Owners" and instead making the process of adding / removing addresses manual. This means we really don't need the upper bound IMO.

mudrila: > practically infinite scrolling

Guess that's not our concern - yeah, our UI is not optimized for such a large number of signers, but welp - it can render that large amount of inputs and you (probably) can create safe with whatever number of `owners` you want

So I'm rather leaning towards removing this logic / copy - but not holding this opinion strong, don't wanna start a debate 🤣

DarksightKellar: > Part of this PRD includes removing the text-entry for "Number of Owners" and instead making the process of adding / removing addresses manual. This means we really don't need the upper bound IMO.

Gotcha. Since this UX is gonna go away anyway, gonna leave as is. 

Also, this will basically mean, "if you can manually add hundreds of owners, then be our guest" no? @parkermccurley

parkermccurley:
Hey @DarksightKellar - we need this to say "transactions" not "proposals". Multisigs don't have proposals :)

PR #2557, opened 2024-11-20
Don't infinite rerender when Safe is not a DAO

adamgall:
Would rather not infinite rerender when trying to load a Safe which is not a DAO.

This is obviously not a good solution, because it doesn't fix the actual issue. The actual issue is that we _should_ be able to load up a Safe which is not yet a DAO. But I guess we recently merged some code which broke that ability. Currently when trying to load a Safe which is not a DAO, it results in the browser tab crashing. This PR simply stops the browser tab from crashing, but it still doesn't load the Safe.


adamgall:
Yikes, nevermind. This change stops _all_ DAOs from loading. Closing.

PR #2562, opened 2024-11-21
`[Issue #2479]` Tie streams to hat

### LINKED ISSUE #2479: BugFix: Termed Payments
So uncovered an issue with tying termed roles to the payments.

Currently when we retrieve regular role payments, we do this via the recipient, its smart account address, which is unique to the role.

For termed roles its the nominee who is the recipient. So what does this mean.

Currently when retrieving the payments, I'm creating a set of recipients created from all the terms nominees of a hat. The problem that it will now pull ANY Sablier payments of that address. This includes those created in other termed hats that nominee has worn and streams that happen outside of our UI.

Da-Colon:
Closes #2479 

Alrighty, Goes with https://github.com/decentdao/decent-contracts/pull/128 over in the contracts.


src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -271,12 +272,20 @@ const useHatsTree = () => {
...
+                .filter(item => item.hatId === BigInt(hat.id))
+                .map(({ streamId }) => {
+                  return streamId;
+                });
               const recipients = hat.roleTerms.allTerms.map(term => term.nominee);
               const uniqueRecipients = [...new Set(recipients)];
               for (const recipient of uniqueRecipients) {
-                payments.push(...(await getPaymentStreams(recipient)));
+                payments.push(
+                  ...(await getPaymentStreams(recipient)).filter(stream =>
```
DarksightKellar, line 285: Can this filtering happen earlier, within getPaymentStreams?

DarksightKellar (dismissed):

DarksightKellar:
Whoops did not means to approve just yet since "no QA yet". Meant to leave a comment

adamgall:
@Da-Colon `@fractal-framework/fractal-contracts@v1.4.2` is being published right now and should be available for consumption in ~3 mins from the time of this comment.

adamgall:
Also @Da-Colon just to be clear

> Until the contracts are published, did not commit the tempAbis and network addresses that I use for testing.

Why did you have new ABIs? The ABIs shouldn't have changed.

Da-Colon:
> Also @Da-Colon just to be clear
> 
> > Until the contracts are published, did not commit the tempAbis and network addresses that I use for testing.
> 
> Why did you have new ABIs? The ABIs shouldn't have changed.

Because I'm sleep deprived

src/hooks/DAO/loaders/useHatsTree.ts
Da-Colon: Hows that?

Da-Colon: Since this is specific to Termed Roles, I'd like to keep the `getPaymentStreams` pure more or less

DarksightKellar: Ah I see

src/components/ui/forms/DatePicker.tsx
```diff
@@ -195,7 +195,15 @@ export function DatePicker({
...
-      onChange?.(new Date(e.setHours(0, 0, 0, 0)));
+      // get the current time and then set selected date to it.
+      const currentHours = new Date().getHours();
+      const currentMinutes = new Date().getMinutes();
+      const currentSeconds = new Date().getSeconds();
+      const currentMilliseconds = new Date().getMilliseconds();
+
+      onChange?.(
+        new Date(e.setHours(currentHours, currentMinutes, currentSeconds, currentMilliseconds)),
+      );
```
adamgall, line 206: @Da-Colon can you explain what this is doing? What's the purpose of this?

src/types/roles.tsx
```diff
@@ -238,12 +238,14 @@ export interface RolesStoreData {
   hatsTree: undefined | null | DecentTree;
   streamsFetched: boolean;
   contextChainId: number | null;
+  hatIdsToStreamIds: { hatId: BigInt; streamId: string }[];
```
adamgall, line 241: I don't love that this data exists on the `RolesStoreData` interface.

IMO this is "internal" data that should only be used within the store to help make connections between other data (Hats and Sablier streams), and shouldn't be exposed outside of the Store itself.

src/store/roles/useRolesStore.ts
```diff
@@ -129,6 +129,9 @@ const useRolesStore = create<RolesStore>()((set, get) => ({
       },
     }));
   },
+  setHatIdsToStreamIds: hatIdsToStreamIds => {
+    set(() => ({ hatIdsToStreamIds }));
+  },
```
adamgall, line 134: It would be nice if, when calling this "setter", instead of just naively setting this property on the store, we merged this data into the existing hats (if they do exist yet). If they're not yet loaded, maybe we could "hold onto this mapping" for when the Hats are loaded in, then merge these streams into the hats at that time. 

Idk, that seems kinda complicated, but as to my earlier point, it seems wrong that we need to expose this mapping of hat ids to stream ids to the rest of the application.

src/components/ui/forms/DatePicker.tsx
Da-Colon: Its easier to remove it, that had a purpose that had changed, but had thought it wouldn't be a bad change anyway, but have since talked myself out. 

The Date object coming out of the calendar will be the start of the day selected. This makes sense.

src/types/roles.tsx
Da-Colon: 🤔 Let me see what I can do

src/store/roles/useRolesStore.ts
Da-Colon: Check out the latest commit, see if thats more or less what you are looking for. Maybe another commit if I go back and change some names

src/store/roles/useRolesStore.ts
```diff
@@ -5,6 +5,15 @@ import { convertStreamIdToBigInt } from '../../hooks/streams/useCreateSablierStr
...
 import { initialHatsStore, sanitize } from './rolesStoreUtils';
 
+const hatToStreamIdMap = new Map<BigInt, string>();
+const getHatToStreamIdMap = () => {
+  return Array.from(hatToStreamIdMap.entries()).map(([hatId, streamId]) => ({
+    hatId,
+    streamId,
+  }));
+};
+const setHatToStreamId = (key: BigInt, value: string) => hatToStreamIdMap.set(key, value);
```
adamgall, line 15: Do you need this `const` extracted out like this? Why not just `hatToStreamIdMap.set(key, value)` on line 154 directly?

adamgall (commented):
In the latest commit, is there a race condition here? Seems like this code will only work correctly if and only if the KeyValuePairs data comes in (and gets stored in the map) _before_ the `updateRolesWithStreams` function runs.

Is that a valid concern?

src/store/roles/useRolesStore.ts
adamgall: It does get closer to what I was thinking, yes! Left another comment below, though.

Da-Colon:
> In the latest commit, is there a race condition here? Seems like this code will only work correctly if and only if the KeyValuePairs data comes in (and gets stored in the map) _before_ the `updateRolesWithStreams` function runs.
> 
> Is that a valid concern?

I don't think so? Since its coming in is being fetched at the same time as the `hatTreeId` which kicks off the retreival of that hats and `setHatsTree` is first called. I believe that these will correctly get set prior to `setHatTree` being called next. 

I thought about update `setHatTreeId` into like `setHatTreeEventData` (The name is why I didn't) to set the loop first. In that I effect. I could just flip These like this for the same effect. 

```
   setHatIdsToStreamIds(getHatIdsToStreamIds(safeEvents, sablierV2LockupLinear, chain.id));
        setHatsTreeId({
          contextChainId: chain.id,
          hatsTreeId: getHatsTreeId(safeEvents, chain.id),
        });
```

Since these are coming from the same `getEvents` called and extracted out. 

I"ve done several refreshes and haven't seen any weird behavior so far.

src/store/roles/useRolesStore.ts
Da-Colon: just didn't opt. wrote everything up top. updated.

mudrila (approved):

mudrila:
> > > In the latest commit, is there a race condition here? Seems like this code will only work correctly if and only if the KeyValuePairs data comes in (and gets stored in the map) _before_ the `updateRolesWithStreams` function runs.
> > > Is that a valid concern?
> > 
> > 
> > I don't think so? Since its coming in is being fetched at the same time as the `hatTreeId` which kicks off the retreival of that hats and `setHatsTree` is first called. I believe that these will correctly get set prior to `setHatTree` being called next.
> > I thought about update `setHatTreeId` into like `setHatTreeEventData` (The name is why I didn't) to set the loop first. In that I effect. I could just flip These like this for the same effect.
> > ```
> >    setHatIdsToStreamIds(getHatIdsToStreamIds(safeEvents, sablierV2LockupLinear, chain.id));
> >         setHatsTreeId({
> >           contextChainId: chain.id,
> >           hatsTreeId: getHatsTreeId(safeEvents, chain.id),
> >         });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Since these are coming from the same `getEvents` called and extracted out.
> > I"ve done several refreshes and haven't seen any weird behavior so far. But I guess thats no saying much.
> 
> @DarksightKellar @mudrila Do either of you have any thoughts on this?

Yeah, I do see Adam's point - @Da-Colon did I get this right that you've changed implementation to what you suggested?

Da-Colon:
yeah I'll go with the thing I suggested of just handling both of these via a single state setter.

Da-Colon:
Check out latest commit @DarksightKellar || @mudrila. look good?

src/store/roles/useRolesStore.ts
```diff
@@ -96,10 +109,21 @@ const useRolesStore = create<RolesStore>()((set, get) => ({
   updateRolesWithStreams: (updatedRoles: DecentRoleHat[]) => {
     const existingHatsTree = get().hatsTree;
     if (!existingHatsTree) return;
+    const streamIdsToHatIdsMap = getStreamIdToHatIdMap();
 
     const updatedDecentTree = {
       ...existingHatsTree,
-      roleHats: updatedRoles,
+      roleHats: updatedRoles.map(roleHat => {
+        const filteredStreamIds = streamIdsToHatIdsMap
```
DarksightKellar, line 117: This here seems like _the thing_ to solve this whole problem, but I don't quite remember what's going on here. Remind me pls?

Everything else looks good to me -- just gotta understand this final piece

Da-Colon: Not sure what you mean? 

So We get the stream/Hat id form keyValue pairs contract. the concern was that roles would be retrieved prior to the map being set.

Sooo. I changed it where it all happens at the same time (setHatTreeId) (setHatIdToStreamId) since they are coming from same place. and set at same time. so we should be good now

PR #2566, opened 2024-11-22
`[PRD]` | `[Issue #2529]` Organization Hierarchy Cache 

### LINKED ISSUE #2529: Once loading first time correctly, we need to cache “Organization” list data
https://www.notion.so/hulking-cosmos-3e8/Q4-2024-QA-Requirements-13e2ee9c5bc280a88458e7a935d17488?pvs=4#1422ee9c5bc2804c88a7c7109f9678f2

Da-Colon:
Closes #2529 

## GIF
screencast 2024-11-22 09-20-44


DarksightKellar (approved):

adamgall:
This implementation uses a cache expiry of 1 week. So at least every 7 days this cache will expire and be re-fetched fresh again.

I'm not sure if that experience is really what we're after here.

But the problem with making the cache expiration longer, is that DAO names can change! So, if a DAO name changes, we'd be displaying the old name until the cache expires.

I dealt with this recently in my "cache favorites" PR. My solution was to:

- set the cache to never expire
- load the data from the cache immediately
- in the background do a fetch and if there have been changes, update the cache

Would something like that be appropriate for this PR?

---

As an aside, we are storing a _lot_ of data for each "Node"; the whole `DaoInfo` object. All we really need are the address and name, right? Wonder if this PR is the place to refactor our Organization page's data requirements.

adamgall (commented):
This implementation uses a cache expiry of 1 week. So at least every 7 days this cache will expire and be re-fetched fresh again.

I'm not sure if that experience is really what we're after here.

But the problem with making the cache expiration longer, is that DAO names can change! So, if a DAO name changes, we'd be displaying the old name until the cache expires.

I dealt with this recently in my "cache favorites" PR. My solution was to:

- set the cache to never expire
- load the data from the cache immediately
- in the background do a fetch and if there have been changes, update the cache

Would something like that be appropriate for this PR?

---

As an aside, we are storing a _lot_ of data for each "Node"; the whole `DaoInfo` object. All we really need are the address and name, right? Wonder if this PR is the place to refactor our Organization page's data requirements.

Da-Colon:
> This implementation uses a cache expiry of 1 week. So at least every 7 days this cache will expire and be re-fetched fresh again.
> 
> I'm not sure if that experience is really what we're after here.
> 
> But the problem with making the cache expiration longer, is that DAO names can change! So, if a DAO name changes, we'd be displaying the old name until the cache expires.
> 
> I dealt with this recently in my "cache favorites" PR. My solution was to:
> 
> * set the cache to never expire
> * load the data from the cache immediately
> * in the background do a fetch and if there have been changes, update the cache
> 
> Would something like that be appropriate for this PR?
> 
> As an aside, we are storing a _lot_ of data for each "Node"; the whole `DaoInfo` object. All we really need are the address and name, right? Wonder if this PR is the place to refactor our Organization page's data requirements.

We can definitely discuss, but this is that change. The view layer component has been updated with only what it needs. 

About DAO Info that is the bare min. it does two calls per safe right now

SafeAPI -> We will need to modules, I could take this out but we will need it back in later anyway
Subgraph -> All the other stuff, DAOName and Heirarchy. 

And thats currently it that is what makes up `DAOInfo`

Da-Colon:
I could break out the Safe Type from `DAOInfo` But I think we are going to do that later in the overhaul and figured why complicated that here?

Da-Colon:
I don't know why I argue you are right, second I took at look at the type I mean duh. one second

Da-Colon:
@adamgall Still need to make the same calls. But created its own type

src/types/fractal.ts
```diff
@@ -204,6 +204,13 @@ export interface DaoInfo extends SubgraphDAOInfo {
   isModulesLoaded?: boolean;
 }
 
+export interface DaoHierarchyInfo {
+  daoName: string | null;
+  nodeHierarchy: NodeHierarchy;
```
adamgall, line 209: This should maybe be a new, smaller type too?

Da-Colon: ah good call.

adamgall (approved):

Da-Colon:
Before this is merged. I am pausing with the thought. So we don't to migrate. might need to go ahead and add in the work for the `type` of governance? so that can get cached as well?

adamgall:
Can bump the CACHE_VERSION for this new local storage store if we ever need to invalidate its cache.

Da-Colon:
> Can bump the CACHE_VERSION for this new local storage store if we ever need to invalidate its cache.
> 
> edit: speaking of which, should probably set a CACHE_VERSION for this so we can invalidate it in the future if/when we need to add more stuff to it, like you're hypothesizing in your previous comment @Da-Colon !
> 
> edit2: just tested, if we don't have a CACHE_VERSION for this new CacheKey, it's all good it just starts at version `undefined`. In the future we can add a CACHE_VERSION and set it to 1 and that will cause old local storage keys to be invalidated.
> 
> edit3: but i mean, it might be good just to explicitly set the version to 1 now in this PR, just for clarity.

100% agreed to set at 1

Da-Colon:
So as you said the expiry is defaults to ONE_WEEK. I'm not really sure what to set it. 

This got me thinking. There are places were caching got us in trouble but saves alot. We could think about a `sync` button that will allow a user to manually refresh cached data? (not for this PR lol)

src/components/DaoHierarchy/DaoHierarchyNode.tsx
```diff
@@ -24,62 +49,104 @@ export function DaoHierarchyNode({
...
+    currentSafe?.address.toLowerCase() === hierarchyNode.safeAddress.toLocaleLowerCase();
   return (
     <Flex
       flexDirection="column"
       alignItems="stretch"
       gap="1.25rem"
       width="100%"
     >
-      <DAONodeInfoCard node={fractalNode} />
+      <Link
```
mudrila, line 112: Seems like we're double-wrapping with link now?

src/components/ui/cards/DAONodeInfoCard.tsx
```diff
@@ -28,39 +30,22 @@ export interface InfoProps extends FlexProps {
...
-      </Flex>
-    );
-  }
 
-  const isCurrentDAO = displayedAddress === currentSafe?.address;
-  const daoName = node.daoName || displayName;
+  const { toggleFavorite, isFavorite } = useAccountFavorites();
 
   return (
     <Link
```
mudrila, line 43: I'd expect this Link wrapping to be removed

mudrila (approved):
Caching looks good - but it seems like we're wrapping info card in link twice

src/components/ui/cards/DAONodeInfoCard.tsx
Da-Colon: Yeah I removed the one i this component

parkermccurley:
Looking great @Da-Colon thank you!

src/types/fractal.ts
```diff
@@ -202,6 +202,18 @@ export interface DaoInfo extends SubgraphDAOInfo {
   isModulesLoaded?: boolean;
 }
 
+export interface DaoHierarchyInfo {
+  daoName: string | null;
+  nodeHierarchy: {
+    parentAddress: Address | null;
+    childNodes: {
+      safeAddress: Address;
+    }[];
```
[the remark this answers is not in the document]
Da-Colon: This is all updated in #2574

PR #2595, opened 2024-12-05
`[Bugfix | Issue #2565]` Network Switching and Context to Store update

Da-Colon:
Closes #2565

## Updates the Network Provider to zustand store
Now switching networks don't cause a massive tree re-render

## Fixes App Crash when navigating to a DAO on a different network via links, refresh or external link
We will need to test with other wallets, but now that 'being connected to a network' is decoupled from the choice in the wallet we don't need to use `switchchain` anymore

Issue #2565, opened 2024-11-22
`Bug` Network Autoswitch not working properly


Da-Colon:
## Problem

Any attempt to use `useSwitchChain` results in nothing happening. This includes Favorites list and pasting a link in the URL if wallet is connected to a different wallet.

Da-Colon:
I'm not really sure what to do here. `switchchain` is just no longer working anymore. I made a bunch of changes and got it work, but then I updated to latest metamask as I had to restart my computer. 

Now switchchain does nothing. I see a notification on the extension but no popup. Metamask now says you no longer need to 'switch' network. 

I'm not really sure atm what needs to be done here. We do need to refactor `NetworkConfig` into a Store but how 'network' works in the app might need to fundamentally change.

PR #2609, opened 2024-12-06
`[Issue #2565]` network config extended

Da-Colon:
Merges with #2595 
Discussed in this comment https://github.com/decentdao/decent-interface/pull/2595#issuecomment-2524193062

This PR will cover any work needed to ensure that the network is properly being used in the App.


src/hooks/utils/useAutomaticSwitchChain.ts
```diff
@@ -8,11 +9,25 @@ export const useAutomaticSwitchChain = ({
...
+  const { switchChain } = useSwitchChain({
+    mutation: {
+      onError: () => {
+        if (addressPrefix !== urlAddressPrefix && urlAddressPrefix !== undefined) {
+          const chainId = getChainIdFromPrefix(urlAddressPrefix);
+          switchChain({ chainId });
+        }
+      },
+    },
+  });
```
Da-Colon, line 21: Funny enough, this kinda reverts back to how the file was before only this now we are attempting the switch chain again onError. 

While, not ideal. this does work and fixes the current network switcher issues we've been having. Ideally I think we would have completely seperate `publicClient` by passing the `id` directly in. and then move all network switching to when a user clicks to send a transaction.

But I'm not sure how that scales at the moment and want to discuss further before we make that change.

So this is the current path of least resistance. currently there are two spots that this is going to be required.

- `useAutomaticSwitchChain`
- `EstablishEssentials` where a new network switcher will exist.

Da-Colon (commented):
Another. I'll create any other PRs for anything pre-existing issues off develop I find while testing so keep an eye out for those. 

And one PR branched from here to updates the DAO search to be cross-chain per @adamgall suggestion. 

- I have another commit to add the SwitchChain to the network switcher in DAO creation coming and this will be good for other testers starting tomorrow
-

Da-Colon:
Okay, I feel like this is good as its going to get in its current form. There is a flash of the network error screen, but that was always there before, but now it actually loads. I think the state refactor will help this get rid of the flash.

adamgall (approved):

src/providers/NetworkConfig/web3-modal.config.ts
```diff
@@ -12,29 +12,32 @@ const supportedWagmiChains = supportedNetworks.map(network => network.chain);
...
+  metadata,
   transports: supportedNetworks.reduce(transportsReducer, {}),
   batch: {
     multicall: true,
   },
 });
 
 if (walletConnectProjectId) {
-  createWeb3Modal({ wagmiConfig, projectId: walletConnectProjectId });
+  createWeb3Modal({ wagmiConfig, projectId: walletConnectProjectId, metadata: metadata });
```
mudrila, line 42: Can use prop shorthand here as well

mudrila (approved):
1 very minor comment, but otherwise - looks good!

PR #2615, opened 2024-12-09
`[Issue | No Issue assigned]` Various loading and error handling cleanup

Da-Colon:
See comments~


src/hooks/DAO/loaders/useHatsTree.ts
Da-Colon, line 1: Because of the need to error handle, wrong network, invalid address, etc. We should not be getting the DAO address from the URL, rather it should come from DAOInfoStore one the first checks have happen and data begins to load. 

This improves load handling between DAOs as well as prevents unnecessary requests

src/hooks/DAO/useKeyValuePairs.ts
Da-Colon, line 1: same as above comment in `useHatsTree`

src/store/roles/rolesStoreUtils.ts
```diff
@@ -134,7 +134,7 @@ export const predictAccountAddress = async (params: {
     tokenId,
   ]);
   if (!(await publicClient.getBytecode({ address: predictedAddress }))) {
-    throw new DecentHatsError('Predicted address is not a contract');
+    return;
```
Da-Colon, line 137: We shouldn't actually throw this. rather just return undefined. We should handle the error of missing smart accounts in the places it happens if needed

src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -297,10 +293,10 @@ const useHatsTree = () => {
   }, [hatsTree, updateRolesWithStreams, getPaymentStreams, streamsFetched]);
 
   useEffect(() => {
-    if (safeAddress && safe?.address && safeAddress !== safe.address && hatsTree) {
+    if (!hatsTreeId && !!hatsTree) {
```
DarksightKellar, line 296: Does `hatsTreeId` go false-y when switching DAOs? Seems this conditional now checks if there's a `hatsTree` but no `hatsTreeId`. Curious on when that will be the case and why

DarksightKellar (commented):
Looks good, just have one question

adamgall (approved):

src/hooks/DAO/loaders/useHatsTree.ts
Da-Colon: yea, honestly wasn't trying to disect the `useHatTree` reset path. Just removing the safeAddress conditional here and had noticed that the hatTreeId was still populating and kicking off request to get the HatTree. no beuno. but not the goal of this particular PR.

Da-Colon: that explaination good or would do you still think its worth a deeper look into to

DarksightKellar: Hmm because I believe the point of this useEffect was to call `resetHatsStore` when the DAO changes. As long as that logic is unchanged I'm cool and prefer this actually

Da-Colon: yep sure is haha. Ill need to update it (just came back here to say that as I ran up against this in test ha)

DarksightKellar (approved):

src/components/ui/menus/SafesMenu/SafesList.tsx
Da-Colon, line 1: In testing found that this component was not being used.

DarksightKellar: Good riddance.

src/hooks/DAO/loaders/useGovernanceContracts.ts
```diff
@@ -23,123 +24,123 @@ export const useGovernanceContracts = () => {
...
+          const possibleLockRelease = getContract({
+            address: govTokenAddress,
+            abi: LockReleaseAbi,
+            client: { public: publicClient },
+          });
+
+          try {
+            const lockedTokenAddress = await possibleLockRelease.read.token();
+            lockReleaseAddress = govTokenAddress;
+            votesTokenAddress = lockedTokenAddress;
```
DarksightKellar, line 83: Can't you just do `votesTokenAddress = await possibleLockRelease.read.token()` on `81` above and be rid of this line?

src/hooks/DAO/loaders/useHatsTree.ts
```diff
@@ -301,12 +296,6 @@ const useHatsTree = () => {
 
     getHatsStreams();
   }, [hatsTree, updateRolesWithStreams, getPaymentStreams, streamsFetched]);
-
-  useEffect(() => {
-    if (safeAddress && safe?.address && safeAddress !== safe.address && hatsTree) {
-      resetHatsStore();
-    }
-  }, [resetHatsStore, safeAddress, safe?.address, hatsTree]);
```
Da-Colon, line 309: moved this to be handled by `useKeyValuePairs` contract, as it loads the hatId and the store doesn't reset properly when done here.

I'm sure we will re-visit all this we when get to the state refactor.

PR #2637, opened 2024-12-16
Refactor `useResolveAddressMultichain` to `useResolveENSName` 

Da-Colon:
@adamgall 

Very similar to `useAddress`. 

Main difference it that the execution the lookup doesn't kickoff via `useEffect` but rather the hook spits out the function to be used as directed. 

if you prefer I just refactor this back and fit `useAddress` in I can do that. I just prefer using a `callback` function to time the execution

### DIFF
--- a/src/hooks/utils/useResolveENSName.ts
+++ b/src/hooks/utils/useResolveENSName.ts
@@ -1,24 +1,21 @@
 import { useState, useCallback } from 'react';
-import { Address, createPublicClient, http, isAddress, getAddress } from 'viem';
+import { Address, createPublicClient, http, isAddress, getAddress, zeroAddress } from 'viem';
 import { normalize } from 'viem/ens';
 import { supportedNetworks } from '../../providers/NetworkConfig/useNetworkConfigStore';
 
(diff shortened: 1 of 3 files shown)


src/hooks/utils/useResolveENSName.ts
```diff
@@ -0,0 +1,63 @@
...
+    });
+    try {
+      const resolvedAddress = await mainnetPublicClient.getEnsAddress({ name: normalizedName });
+      if (resolvedAddress) {
+        returnedResult.resolvedAddress = resolvedAddress;
+        returnedResult.isValid = true;
+      }
+    } catch {
+      // do nothing
+    }
```
adamgall, line 58: I don't think this try/catch is necessary any longer. I believe it was here previously to catch the error in `.getEnsAddress` when the call was made on a network that didn't ahve the ENS contracts deployed (all of them except for mainnet and sepolia IIRC).

adamgall (changes requested):

adamgall (approved):

PR #2641, opened 2024-12-18
`ENG-45` | Fix Address xyz is invalid bug

DarksightKellar:
Closes https://linear.app/decent-labs/issue/ENG-45/addess-xxx-is-invalid

### DIFF
--- a/src/hooks/DAO/loaders/useDecentTreasury.ts
+++ b/src/hooks/DAO/loaders/useDecentTreasury.ts
@@ -170,7 +170,7 @@ export const useDecentTreasury = () => {
       // make unique
       .filter((value, index, self) => self.indexOf(value) === index)
       // turn them into Address type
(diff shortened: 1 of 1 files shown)


adamgall (commented):
Hm, at first glance this change doesn't actually do anything different. Is the bug reproducible?

DarksightKellar:
> Hm, at first glance this change doesn't actually do anything different. Is the bug reproducible?

@adamgall  Yep, see the comment in the issue.

I think the thing is that `getAddress` takes _two_ args, and doing `.map(getAddress);` causes it to not actually run as intended. Assumes only one arg

adamgall:
Yep you're right @DarksightKellar , thank you

adamgall (approved):

Da-Colon (approved):

PR #2642, opened 2024-12-18
`[Bugfix | No Issue]` Add `useEffect` for when wallet is connected to different network

Da-Colon:
Noticed that the wallet wouldn't switch to ethereum on page load and that switching networks via the selection wasn't causing hte network to change

### DIFF
--- a/src/components/DaoCreator/formComponents/EstablishEssentials.tsx
+++ b/src/components/DaoCreator/formComponents/EstablishEssentials.tsx
@@ -108,6 +108,13 @@ export function EstablishEssentials(props: ICreationStepProps) {
     },
   });
 
+  useEffect(() => {
+    if (chain.id !== walletChainID) {
+      const chainId = getChainIdFromPrefix(addressPrefix);
+      switchChain({ chainId });
+    }
+  });
+
   return (
     <>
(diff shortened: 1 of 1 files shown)


DarksightKellar (approved):

adamgall (commented):
@Da-Colon do we need/want a dependency array for this `useEffect`?

adamgall (approved):
Works well for me, thanks @Da-Colon !

src/components/DaoCreator/formComponents/EstablishEssentials.tsx
```diff
@@ -113,7 +113,7 @@ export function EstablishEssentials(props: ICreationStepProps) {
       const chainId = getChainIdFromPrefix(addressPrefix);
       switchChain({ chainId });
     }
-  });
+  }, [chain.id, walletChainID, addressPrefix, switchChain]);
```
DarksightKellar, line 116: With the deps, I'm worried about what might happen if the user switches their wallet network themselves, resulting in `walletChainID` being different from `chain.id` (right?), and triggering a call to `switchChain` to `chainId` from unchanged `addressPrefix`

PR #2645, opened 2024-12-20
`[Eng 50]` | Dao creation network switch

Da-Colon:
Closes https://linear.app/decent-labs/issue/ENG-50/dev-error-creating-an-nft-dao

So the problem is that I placed the useEffect within the EstablishEssentials form rather than higher up. 

This move makes it if the User switches there network at some point later in the process, the app switches it back keeping the user on the selected network. 

I did attempt to move this selection and network check to the button, but with the current setup within `useBuildDAOTx` where it pulls contracts from `useNetworkConfigStore` in which is set to the current network. 

A future refactor will be needed to confirm to a new pattern of more fine control over what network the DAO is deployed on. 

## Testing: 
- Enter Create DAO workflow
- Select Different network from etherem
- Go to end of which DAO creation workflow you are on.
- Before deploying, open wallet and manually switch wallet network
- Go back to page (focus on page), go back to wallet and network should be back to the originally selected network. 

Basically our now app now prevents the user from manually switching thier network via wallet during this workflow. Not ideal to be so connected to that network. but without some refactoring how the contracts for deployment are chosen this is the most effective method to ensure proper deployment

### DIFF
--- a/src/components/DaoCreator/index.tsx
+++ b/src/components/DaoCreator/index.tsx
@@ -1,6 +1,9 @@
 import { Box } from '@chakra-ui/react';
 import { Formik } from 'formik';
+import { useEffect } from 'react';
+import { useChainId, useSwitchChain } from 'wagmi';
 import { useDAOCreateSchema } from '../../hooks/schemas/DAOCreate/useDAOCreateSchema';
+import { useNetworkConfigStore } from '../../providers/NetworkConfig/useNetworkConfigStore';
 import {
   AzoriusERC20DAO,
   AzoriusERC721DAO,
@@ -28,6 +31,8 @@ function DaoCreator({
   mode: DAOCreateMode;
 }) {
   const { totalParentVotingWeight } = useParentSafeVotingWeight();
+  const walletChainID = useChainId();
+  const { chain } = useNetworkConfigStore();
 
   const { createDAOValidation } = useDAOCreateSchema({
     isSubDAO: !!isSubDAO,
@@ -36,6 +41,21 @@ function DaoCreator({
 
   const { prepareMultisigFormData, prepareAzoriusERC20FormData, prepareAzoriusERC721FormData } =
     usePrepareFormData();
+
+  const { switchChain } = useSwitchChain({
+    mutation: {
+      onError: () => {
+        if (chain.id !== walletChainID && !isSubDAO) {
+          switchChain({ chainId: chain.id });
+        }
+      },
+    },
+  });
+  useEffect(() => {
+    if (chain.id !== walletChainID && !isSubDAO) {
+      switchChain({ chainId: chain.id });
+    }
+  }, [chain.id, walletChainID, switchChain, isSubDAO]);
 
   return (
     <Box>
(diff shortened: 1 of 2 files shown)


src/components/DaoCreator/index.tsx
```diff
@@ -37,6 +42,21 @@ function DaoCreator({
...
     usePrepareFormData();
 
+  const { switchChain } = useSwitchChain({
+    mutation: {
+      onError: () => {
+        if (chain.id !== walletChainID && !isSubDAO) {
+          switchChain({ chainId: chain.id });
+        }
+      },
+    },
```
adamgall, line 52: @Da-Colon can you explain what this mutation is and why it's needed?

Da-Colon: So my understanding is that Metamask doesn't require that popup permission to switch networks anymore. Network permission are done within Metamask and if its listed then its available. 

So when `switchchain` is called if the network is allowed then switch should happen. BUT for some reason this doesn't happen as it should. I looked into this before and mentioned a while back that I believe this error is occurs in 
 the current version of `Viem`. 
 
 So during testing. I found that this pattern 'brute' forces past the error and switches the network. Didn't see this problem in Frame (didn't reach the error). 
 
 There are two ways to get rid of this need more than likely. 
 - Update viem to latest, but requires some changes in the app as they change `Address` typing to string for whatever reason.
 - We could also create our own `wallet client` to use in the app. I attempted to set this up here. but requires a bit more thought, a little refactor and additional testing for how network config (contracts) is used in this workflow.

Da-Colon: As far as what the mutation, it taps into the internal tanstack query parameters and I chose to use the `onError` from there. I guess looking at the docs that I could look to see if just using onSettled will work.

Da-Colon:
> Tested against Coinbase and TokenPocket wallet extensions. CB works fine because when the app request the wallet to change network, it just does it. TP has a little problem.
> 
> Steps:
> 
> 1. Connect TP wallet with Ethereum selected.
> 2. Create a DAO in our app. Select "Sepolia" as the network.
> 3. Note, TP shows prompt to change network. If you click "Cancel", it keeps on prompting. Click on "Confirm" now.
> 4. Click "Next"
> 5. In TP, switch network back to "Ethereum". Note, immediately, it shows a badge with "1" label. If you click on it, it is asking you to switch back to Sepolia. However, the badge is not super obvious and user may miss it.
> 6. Create NFT DAO with 0x31408f226E37FBF8715CA6eE45aaB4Ea213bA7A5 as address (this issue should happen with ERC-20 DAO too). You will see the address is red with error because it is now using Ethereum mainNet to check the address.
> 
> Conclusion, there are many wallet variations and different network switching flows, that may still cause two source of truth in our app.
> 
> I think we should ignore the wallet network selection during our flow, and only prompt the wallet to switch network at signing (so only one source of truth with user selected network in the app)

Let's discussed this on a call when available.

adamgall (approved):

PR #2648, opened 2024-12-20
ENG-36 Feature flag pattern. See README.md for usage

johnqh:
See README.md for more details.

It uses an Injection, so if/when we use remote feature flags, we can change the injection in main.tsx without changing any downstream code.


adamgall (commented):
@johnqh please see the file `src/constants/common.ts`, there is some additional "feature flag" code implemented there. I *think* can we probably safely remove all of those, they're probably all on by default at this point. Gotta double check though.

src/helpers/featureFlags.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export class EnvironmentFeatureFlags implements IFeatureFlags {
+  urlFlags: { [key: string]: any } = {};
+  set(key: string, value: any): void {
+    this.urlFlags[key] = value;
+  }
+
+  get(key: string): any {
+    return this.urlFlags[key] ?? import.meta.env[key];
```
DarksightKellar, line 17: I'd like this to instead be `return import.meta.env[key] ?? this.urlFlags[key];` so that we can disable a feature on the environment level and not worry about someone setting a url param to get around it, even if it's a very unlikely scenario

src/main.tsx
```diff
@@ -1,11 +1,12 @@
 import '@fontsource/space-mono';
 import React from 'react';
 import ReactDOM from 'react-dom/client';
-import { RouterProvider } from 'react-router-dom';
+import { RouterProvider, useSearchParams } from 'react-router-dom';
```
DarksightKellar, line 4: unused import `useSearchParams`

mudrila, line 4: Seems like `useSearchParams` is not used in this file?

src/constants/common.ts
```diff
@@ -63,12 +64,7 @@ const features = {
 type FeatureFlag = (typeof features)[keyof typeof features];
 
 export const isFeatureEnabled = (feature: FeatureFlag) => {
-  const featureStatus = import.meta.env[`VITE_APP_FLAG_${feature}`];
-  if (featureStatus === 'ON') {
-    return true;
-  } else {
-    return false;
-  }
+  return FeatureFlags.instance?.get(`VITE_APP_FLAG_${feature}`) === 'ON';
```
mudrila, line 67: Seems like this `=== 'ON'` conflicts with documentation - since in the docs we're referring to the value as boolean, while here we still rely on `ON`.
I'd like to stick to `ON` value - just aesthetic matter, so loosely held preference. But yeah - we need to align implementation

src/pages/home/HomePage.tsx
```diff
@@ -19,6 +21,11 @@ export default function HomePage() {
     }
   }, [safe?.address, action]);
 
+  const [searchParams] = useSearchParams();
+  searchParams.forEach((value, key) => {
```
mudrila, line 25: We probably should wrap this `forEach` into `useEffect` to prevent re-running this loop during every re-rendering. I'm quite sure we wanna run this loop only when `searchParams` gets changed

src/constants/common.ts
DarksightKellar, line 67: Since we're bridging these two feature flag implementations, can we go all the way? I'm thinking:
- Move all feature-flag code from here to `helpers/featureFlags.ts` (including, IMO, `isDevMode` on `70`)
- In `EnvironmentFeatureFlags`, use the `FeatureFlag` type so we're not depending on strings
- Update readme to reflect that adding/removing a feature flag should also happen in `const features` on `58`
- Might as well move `isFeatureEnabled` into `EnvironmentFeatureFlags`. I dunno.

src/helpers/featureFlags.ts
```diff
@@ -0,0 +1,19 @@
+export interface IFeatureFlags {
+  set(key: string, value: any): void; // set from URL params
+  get(key: string): any;
+}
+
+export class FeatureFlags {
+  static instance?: IFeatureFlags;
+}
+
+export class EnvironmentFeatureFlags implements IFeatureFlags {
```
mudrila, line 10: I'm wondering if we wanna run this with class implementation in comparison to other patterns we have in the app. I'd prefer using React context or Zustand store to gain better observability and streamline UI update when search parameters are getting changed. 

Current `isFeatureEnabled` does not correlate with the observability and reactivity that we're gaining through `useSearchParams` - so we either should move logic with search params right into this class, obtaining query parameters using vanilla browser API, or refactor class into Zustand store - to keep reactivity consistent.

DarksightKellar (changes requested):
This is nice. I got a couple or so change requests

mudrila (changes requested):
Love dependency injection - though I think current implementation brings mixing conceptually different approaches - and I feel like we should streamline it to either leverage existing reactivity / observability patterns with Zustand & React Router - or we should keep it "vanilla-only" and remove usage of `useSearchParams`, moving logic of initializing feature flags  from query parameters into constructor of `EnvironmentFeatureFlags`

src/helpers/featureFlags.ts
adamgall: I personally think it would be good to allow both scenarios:
- feature disabled at environment level, enableable at query flag level
- feature enabled at environment level, disableable at query flag level

src/pages/home/HomePage.tsx
adamgall: I'm confused about why this code lives on the HomePage. What if a user loads a DAO directly? Then I think no feature flags are going to be set

adamgall (commented):
John could you also implement a small "other side" implementation, so this PR goes end to end? Something small, like an emoji in the footer or something. But importantly, this should utilize your new feature flag system. I'm just curious about seeing how this works in practice.

adamgall (changes requested):
When using the flag on any url which isn't the HomePage component results in the flag not being set

image

Would recommend you check out the code in src/components/ui/page/Global/index.tsx

src/constants/common.ts
```diff
@@ -1,5 +1,6 @@
 import { useBreakpointValue } from '@chakra-ui/react';
 import { Hex } from 'viem';
+import { FeatureFlags } from '../helpers/featureFlags';
```
adamgall, line 3: Is this unused? (I'm on mobile, hard to tell...)

src/helpers/featureFlags.ts
```diff
@@ -0,0 +1,47 @@
+export enum FeatureFlag {
+  developmentMode = 'VITE_APP_FLAG_DEVELOPMENT_MODE',
+  demoMode = 'VITE_APP_FLAG_DEMO_MODE',
+  configDemo = 'VITE_APP_FLAG_CONFIG_DEMO',
+}
+
+export interface IFeatureFlags {
+  set(key: FeatureFlag, value: any): void; // set from URL params
+  get(key: FeatureFlag): any;
```
adamgall, line 9: Any reason we have these values typed as `any` instead of like `boolean`

src/helpers/featureFlags.ts
```diff
@@ -0,0 +1,47 @@
+export enum FeatureFlag {
+  developmentMode = 'VITE_APP_FLAG_DEVELOPMENT_MODE',
+  demoMode = 'VITE_APP_FLAG_DEMO_MODE',
+  configDemo = 'VITE_APP_FLAG_CONFIG_DEMO',
```
adamgall, line 4: Since these are no longer environment variables (ah, reminder, delete them from the `.env` reference file please), and are controlled via query params, no real reason to keep this verbose syntax imo.

Edit: never mind, if the intention is to allow for query flags to override build time env bars, I guess they will need to match. Or... you can imply/strip the `VITE_APP_` prefix in the query params.

src/main.tsx
```diff
@@ -25,6 +26,8 @@ function DecentRouterProvider() {
 async function initializeApp() {
   await runMigrations();
 
+  FeatureFlags.instance = new EnvironmentFeatureFlags();
+
```
adamgall, line 30: Seeing a singleton get created outside of React like this makes me nervous! Seems like an antipattern to do this is a React codebase.

src/hooks/utils/useCanUserSubmitProposal.ts
```diff
@@ -2,12 +2,12 @@ import { abis } from '@fractal-framework/fractal-contracts';
 import { useCallback, useEffect, useState } from 'react';
 import { Address, getContract } from 'viem';
 import { useAccount, usePublicClient } from 'wagmi';
-import { isDemoMode } from '../../constants/common';
 import { useFractal } from '../../providers/App/AppProvider';
 import { useSafeAPI } from '../../providers/App/hooks/useSafeAPI';
 import { useDaoInfoStore } from '../../store/daoInfo/useDaoInfoStore';
 import { GovernanceType } from '../../types';
 import useVotingStrategiesAddresses from './useVotingStrategiesAddresses';
+import { isDemoMode } from '../../helpers/featureFlags';
```
DarksightKellar, line 10: Linter error here, Just the order of imports

README.md
```diff
@@ -41,6 +41,56 @@ It is crucial to have `Netlify` functions running locally to work with anything
...
+  demoMode = 'VITE_APP_FLAG_DEMO_MODE',
+  configDemo = 'VITE_APP_FLAG_CONFIG_DEMO',
+}
+```
+
+and add a convenience function for FLAG_DEMO:
+
+```typescript
+export const isConfigDemo = () => isFeatureEnabled(FeatureFlag.configDemo);
+```
```
adamgall, line 66: My suggestion is that we delete this pattern and just have callers utilize `isFeatureEnabled(FeatureFlag.feature)`

src/main.tsx
DarksightKellar: Eh I'm not too concerned in this instance because the code has nothing to do with React.

src/helpers/featureFlags.ts
DarksightKellar: I suppose the current `this.urlFlags[key] ?? import.meta.env[key]` fulfils that need then no? Can I resolve this conversation? @adamgall @johnqh

adamgall: Yeah ur right this does work. It's a little strange to override an ON env var with an OFF query param, because technically every single string which is not `"ON"` acts as "off" in the query param flags.

src/helpers/featureFlags.ts
adamgall: @johnqh but if they're coming from Environment Variables or Query Parameters they are strings, right?

src/main.tsx
adamgall: Oh no need to move the EnvironmentFeatureFlags class implementation to another file, I get what's going on here.

I think my concern was more around the fact that by creating this outside of React (outside of a hook, more specifically), we can't tap into `react-router-dom`'s `useLocation()` hook, which gives us access to query params. Importantly, it'll give us access to query params _all the time, whenever they change, regardless of how they change_. In the current implementation, we get access to the query params at page load and that's it. If there's other code in the codebase that, for example, sets a query param to turn on a feature flag, our current implementation here won't pick it up.

README.md
adamgall: Yeah I hear you. It's no big deal either way i guess.

src/helpers/featureFlags.ts
adamgall: I was talking about something slightly different than what you responded to, I think... 
I was actually talking about how you can declare one set of strings (like the full `VITE_APP_ENV_VAR`s), and then programmatically transform them into a "friendlier" string which could be used for the Query Params. I only say that because those big environment variables in the URL look ugly. See https://github.com/decentdao/decent-interface/pull/2673/files#diff-6eeb99f68ddd47adea0a1295d32ea04b3be72f8c13fc698d268950227e41d603R25-R28

---

but, for more context to your response:

Currently none of these flags are set on deployed instances.

Let me take that back.

To be clear, `VITE_APP_FLAG_DEMO_MODE=ON` is set on the `app.demo.decentdao.org` Netlify site. But when we switch over to Cloudflare, we won't have a standalone separate instance of a "demo site url" like we do with Netlify. I do not intend to set the `VITE_APP_FLAG_DEMO_MODE=ON` environment variable on the "Preview" deployments coming from Cloudflare (every branch except the "Production" branch: `main`)

src/main.tsx
adamgall: Ok, that is fair. My next question follows the first then. Will consumer components of this object's `get` function automatically re-render when the underlying value is changed from somewhere else?

src/helpers/featureFlags.ts
adamgall: I am under the impression that from the point of view of all of these cloud provider build platforms, the "environment variables" that i'm setting at the Netlify / Cloudflare dashboard level are all treated as strings.

adamgall (approved):

adamgall (approved):
Looks good to me thank you John!

PR #2657, opened 2025-01-08
`[ENG-84]` Toast not closing

Da-Colon:
## Description
Looks like this was set to `infinity`, removed that.

Closes [ENG-84](https://linear.app/decent-labs/issue/ENG-84/cannot-close-toast)

## Testing
As this particular toast is part of a failure workflow, might take a few tries to reproduce. I have this happen more often when creating Token DAOs, so I'd recommend that, if you are itching to test, otherwise, pretty straight forward fix.

## Screenshots
No Screenshots

### DIFF
--- a/src/pages/create/SafeCreatePage.tsx
+++ b/src/pages/create/SafeCreatePage.tsx
@@ -37,9 +37,7 @@ export function SafeCreatePage() {
       if (daoFound) {
         navigate(DAO_ROUTES.dao.relative(addressPrefix, safeAddress));
       } else {
-        toast.loading(t('failedIndexSafe'), {
-          duration: Infinity,
-        });
+        toast.loading(t('failedIndexSafe'));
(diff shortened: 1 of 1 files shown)


DarksightKellar (approved):

src/pages/create/SafeCreatePage.tsx
```diff
@@ -37,9 +37,7 @@ export function SafeCreatePage() {
       if (daoFound) {
         navigate(DAO_ROUTES.dao.relative(addressPrefix, safeAddress));
       } else {
-        toast.loading(t('failedIndexSafe'), {
-          duration: Infinity,
```
mudrila, line 41: Yeah, this thingy was intentional previously :)
But I know from user feedback that it's rather confusing.

mudrila (approved):

PR #2658, opened 2025-01-08
`[ENG-95]` DAO Create: White screen

Da-Colon:
I

### DIFF
--- a/src/components/DaoCreator/index.tsx
+++ b/src/components/DaoCreator/index.tsx
@@ -1,6 +1,6 @@
 import { Box } from '@chakra-ui/react';
 import { Formik } from 'formik';
-import { useEffect } from 'react';
+import { useEffect, useRef } from 'react';
 import { useChainId, useSwitchChain } from 'wagmi';
 import { useDAOCreateSchema } from '../../hooks/schemas/DAOCreate/useDAOCreateSchema';
 import { useNetworkConfigStore } from '../../providers/NetworkConfig/useNetworkConfigStore';
@@ -51,10 +51,13 @@ function DaoCreator({
       },
     },
   });
+
+  const isAppLoadedRef = useRef(false);
   useEffect(() => {
-    if (chain.id !== walletChainID && !isSubDAO) {
+    if (chain.id !== walletChainID && !isSubDAO && isAppLoadedRef.current) {
       switchChain({ chainId: chain.id });
     }
+    isAppLoadedRef.current = true;
   }, [chain.id, walletChainID, switchChain, isSubDAO]);
 
   return (


src/components/DaoCreator/formComponents/EstablishEssentials.tsx
```diff
@@ -144,7 +130,6 @@ export function EstablishEssentials(props: ICreationStepProps) {
               selectedItem={dropdownItems.find(item => item.selected)}
               onSelect={item => {
                 setCurrentConfig(getConfigByChainId(Number(item.value)));
```
Da-Colon, line 132: just switch network config's network

src/hooks/DAO/useDeployDAO.ts
```diff
@@ -16,10 +16,13 @@ const useDeployDAO = () => {
...
   const {
     addressPrefix,
+    chain,
     contracts: { multiSendCallOnly },
   } = useNetworkConfigStore();
 
-  const { data: walletClient } = useWalletClient();
+  const { data: walletClient } = useWalletClient({
+    chainId: chain.id,
+  });
```
Da-Colon, line 25: This was the key. So basically the error I kept running into was even though `switchChain` was being called the walletClient being referenced in the `deployDAO` callback was still expected the wallet's network. 

I had tried before adding this without `switchChain` and was seeing the same error as I was seeing with switchChain alone. 

combining the two seems to work as the walletClient now is setup to expect the network change, I guessss

DarksightKellar (approved):
LGTM, tested, works for me

src/hooks/DAO/useBuildDAOTx.ts
```diff
@@ -42,7 +43,8 @@ const useBuildDAOTx = () => {
     governanceContracts: { linearVotingErc721Address },
   } = useFractal();
   const user = useAccount();
-  const publicClient = usePublicClient();
+
+  const publicClient = useNetworkPublicClient();
```
mudrila, line 47: I truly feel little bit weird about the fact that our setup leads to inability of simple `usePublicClient` usage - for me that's a signal that our set of classes that performs transactions encoding doesn't suit reactive nature of our application...

src/hooks/useNetworkPublicClient.ts
```diff
@@ -0,0 +1,13 @@
...
+import { useNetworkConfigStore } from '../providers/NetworkConfig/useNetworkConfigStore';
+
+export default function useNetworkPublicClient() {
+  const { chain, rpcEndpoint } = useNetworkConfigStore();
+  return createPublicClient({
+    chain,
+    batch: {
+      multicall: true,
+    },
+    transport: http(rpcEndpoint),
```
mudrila, line 11: With this I'm wondering where we also need to add `ws` in order to being able to listen for event updates - so that we're connected to socket instead of polling

mudrila (approved):
Looks good - though I feel like we should be using `usePublicClient` in ideal world. The fact that it doesn't suit completely our needs gives me a feeling we're doing something incorrectly in networks management

src/hooks/useNetworkPublicClient.ts
Da-Colon: For sure future upgrades.

src/hooks/DAO/useBuildDAOTx.ts
Da-Colon: I don't think I understand your concern? We have outgrown the simple plug and play `usePublicClient`. we could use it by passing in the `chainId` but that requires an extra import. We COULD just wrap the hook in `useNetworkPublicClient` but felt creating it directly remove any hook weirdness?

src/components/DaoCreator/formComponents/EstablishEssentials.tsx
adamgall: I like the whole idea of removing the coupling between a user's connected wallet's selected network, and "the one network" that our app is configured with at that moment.

In that vein, i think this global "current config" thing needs to change. Maybe not in this PR, but just in general. It doesn't make sense to have a "current config" in an app that doesn't have any more concept of a Currently Selected Network.

Also, as an aside, a bug im noticing on current develop. I have Mainnet Decent DAO loaded. But for some reason in Rabby i'm "connected" to Sepolia. (actually the Decent app auto-switched my Rabby network from Mainnet to Sepolia when I connected my wallet while on the Decent App Home Page). So anyway, the Decent DAO Dashboard page doesn't display the "Delegate" or "Create Proposal" buttons even though I'm connected with the right account, _because my Rabby Wallet connection is on Sepolia, not Mainnet_. When I switch Rabby to Mainnet (and refresh), those buttons correctly appear now.

So what I'm getting at is, that even in our DAO loading where we've already tried to decouple the Connected Wallet's Network and the "current config" in the app which we set when a dao fist starts to load and we get it's network from the URL, we are still relying on some wallet-specific network (read data) behaviors and there are bugs under those rocks still.

adamgall (approved):
Originally we were able to rely on this idea that whatever wallet the user was connected with, our app would be connected to that wallet's currently selected network.

Our codebase currently uses `usePublicClient()` (a hook from `wagmi`) to get an instance of a viem public client _which is using the same network as the user's connected wallet_. So basically, much of our app's data-fetching pathways are a direct downstream effect of whatever network our connected user decides to switch their wallet to while connected to our site.

Lots of code has been written to try to fight this behavior. We've written code to try to automatically switch a user's wallet to the one "we want them to be on", in order to load a DAO. We've written code that notices if a user is trying to switch the network in their wallet and stops them from doing so (by switching it back). Lots of this shit has inconsistent behaviors between wallets, too.

But we've been approaching it wrong.

We've been trying to control what network a user's wallet is connected to, because we are actually trying to control the network of the client that `usePublicClient()` gives us.

David's initiative here is a step in the direction where _we don't use `usePublicClient()` to get instances of public clients, because that hook gives us a client that's always tied to the user's currently selected network_.

A better pattern is to not rely on or use `usePublicClient()`, but simply create (and reuse? maybe not) public client instances connected to explicit networks.

cc @johnqh

Da-Colon:
> also, as an aside, a bug im noticing on current develop. I have Mainnet Decent DAO loaded. But for some reason in Rabby i'm "connected" to Sepolia.

Rabby done snuck and bit me. yeah agreed. hmmm I guess this effort continues. Let's talk about next steps tommorrow a little further.  Specifically two points for easy next steps.

> but simply create (and reuse? maybe not) public client instances connected to explicit networks.

Could be as easy as find and replace `usePublicClient` -> `useNetworkPublicClient` and stress test or the (and reuse?)

> In that vein, i think this global "current config" thing needs to change. Maybe not in this PR, but just in general. It doesn't make sense to have a "current config" in an app that doesn't have any more concept of a Currently Selected Network.

I think am smelling what you are cooking.

PR #2659, opened 2025-01-08
[Cloudflare Base] Migration

adamgall:
Closes https://linear.app/decent-labs/issue/ENG-100/mirgate-releases-from-netlify-to-cloudflare

This PR is the "base" Cloudflare branch, to make PR reviews of all of the work necessary for this migration easier.

There will be a number of follow up PRs, all targeting the one before it, all of them ultimately ending up in here. 

The Cloudflare deployment environments are like so:

- Production deployment (tracking this `cloudflare` branch (for now)): https://app.new.decentdao.org
- All other branches get preview deployments at: https://branch-name.decent-interface.pages.dev

Until the first PR is merged into this branch, the production Cloudflare deployment will fail.


mudrila:
@adamgall Can you please link Linear ticket or duplicate rationale behind the cause of migration to Cloudflare?
Not saying I'm against it but I don't really understand necessity for migrating from one static-optimized deployment provider to another, even more optimized - especially in the light of our emerging discussions around having server / API (where Netlify would work better and set up will be simpler, or some other options gonna be even more suitable)

functions/balances/index.ts
```diff
@@ -0,0 +1,83 @@
...
+    },
+  },
+} as const;
+
+type BalanceType = keyof typeof endpoints;
+const ALL_BALANCE_TYPES: BalanceType[] = ['tokens', 'nfts', 'defi'];
+
+export const router = new Hono<{ Bindings: Env }>().use('*', getParams).get('/', async c => {
+  const { address, network } = c.var;
+  const flavors = c.req.queries('flavor') as BalanceType[] | undefined;
```
mudrila, line 66: Is it like a magic key or something? Or some copy-paste left-overs? Feel like `flavors` naming doesn't fit nicely here lol

functions/balances/transformers.ts
```diff
@@ -0,0 +1,38 @@
+import { DefiBalance, NFTBalance, TokenBalance } from '../../src/types/daoTreasury';
+import { DefiResponse, NFTResponse, TokenResponse } from '../shared/moralisTypes';
+
+export function transformTokenResponse(token: TokenResponse): TokenBalance {
+  return {
+    ...token,
+    tokenAddress: token.token_address,
```
mudrila, line 7: `<bruh>` Could be done with `lodash` in 1 liner instead of 3 dedicated functions `</bruh>`

To be serious - think being intentional is actually better here :)

functions/balances/transformers.ts
```diff
@@ -0,0 +1,38 @@
...
+    metadata: nft.metadata ? JSON.parse(nft.metadata) : undefined,
+    tokenUri: nft.token_uri,
+    name: nft.name || undefined,
+    symbol: nft.symbol || undefined,
+    amount: nft.amount ? parseInt(nft.amount) : undefined,
+  };
+}
+
+export function transformDefiResponse(defi: DefiResponse): DefiBalance {
+  return {
```
mudrila, line 35: Do we even need to spread it here? Maybe just returning `defi` straight away?

functions/shared/moralisApi.ts
```diff
@@ -0,0 +1,75 @@
...
+  chain,
+  apiKey,
+  params = {},
+}: MoralisRequestConfig): Promise<T[]> {
+  let allResults: T[] = [];
+  let cursor: string | null = null;
+  const limit = 100;
+
+  do {
+    const chainHex = `0x${parseInt(chain).toString(16)}`;
```
mudrila, line 27: Isn't it pretty much `toHex` from `viem`?

functions/shared/moralisApi.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    // Handle no results
+    if (!data || !data.result) {
+      break;
+    }
+
+    allResults = allResults.concat(data.result);
+    cursor = data.cursor;
+
+    // Add small delay between requests
```
mudrila, line 68: Is it threshold matter?

functions/shared/moralisTypes.ts
```diff
@@ -0,0 +1,64 @@
+import { DefiPosition, NFTMedia } from '../../src/types/daoTreasury';
+
+export type TokenResponse = {
```
mudrila, line 3: Lol - I just realized we ain't using library anymore
Was there a specific reason for that? Guess it wasn't playing nice with Cloudflare, so just wanted to know

package.json
```diff
@@ -63,10 +66,11 @@
     "pretty": "prettier . --write",
     "pretty:check": "prettier . --check",
     "dev": "vite --force",
-    "dev:netlify": "netlify dev",
+    "dev:netlify": "VITE_APP_USE_LEGACY_BACKEND=true netlify dev",
```
mudrila, line 69: Should there be some command for running with Cloudflare locally? I guess since we're migrating from Netlify - why should we keep Netlify functions then?

mudrila: Or that would be next step?

src/providers/App/hooks/useBalancesAPI.ts
```diff
@@ -4,62 +4,94 @@ import { Address } from 'viem';
...
   const {
     chain,
     moralis: { deFiSupported },
   } = useNetworkConfigStore();
   const { t } = useTranslation('treasury');
 
-  const getTokenBalances = useCallback(
-    async (address: Address): Promise<{ data?: TokenBalance[]; error?: string }> => {
+  const fetchBalancesNew = useCallback(
+    async (address: Address, flavors: string[]): Promise<BalanceResponse> => {
```
mudrila, line 24: Again wondering why `flavors` 😄

src/types/daoTreasury.ts
```diff
@@ -87,20 +87,22 @@ type NftMediaItem = {
   url: string;
 };
 
+export type NFTMedia =
```
mudrila, line 90: Is this type reused anywhere? Haven't noticed that
If it's not reused - can we avoid exporting it? To make it explicit that it is local to the `daoTreasury` types

wrangler.toml
```diff
@@ -0,0 +1,50 @@
...
+compatibility_date = "2025-01-08"
+
+[[kv_namespaces]]
+binding = "balances"
+id = "680ec75643924ba4b5d0692aa999c3a1"
+
+[vars]
+BALANCES_CACHE_INTERVAL_MINUTES = "60"
+BALANCES_MORALIS_INDEX_DELAY_MINUTES = "5"
+VITE_APP_ALCHEMY_API_KEY = "TWthr43vQ-Vn50I7t4pr8N8Idui9zdEh"
```
mudrila, line 13: I reeeeeealy don't think we wanna this being exposed in the repo
Should this file be ignored? Probably, since it is generated

mudrila (changes requested):
Generally - looks good
The only big concern is `wrangler.toml`
Then couple of minor questions / comments which are rather optional

functions/balances/index.ts
adamgall: Not a magic key, just changing the way the API works.

On netlify, the schema is:

```
GET /functions/balancesTokens
GET /functions/balancesNFTs
GET /functions balancesDefi
```

But I changed it in the CF implementation so that there's only one `GET /api/balances` endpoint, and it takes (up to multiple) query params called "`flavor`" which can be values of "`tokens`", "`nfts`", "`defi`", and the response will include the requested balances flavors.

I agree flavors is a weird word.

functions/balances/transformers.ts
adamgall: sure

functions/shared/moralisApi.ts
adamgall: Yep good call, changed to `const chainHex = toHex(parseInt(chain));`

functions/shared/moralisApi.ts
adamgall: I assume you're talking about the "small delay between requests" via setTimeout?

Yeah I'm not sure to be honest, the AI wrote this code and this is one piece of it that I didn't question. I should have. I think we start by removing it lol.

functions/shared/moralisTypes.ts
adamgall: Yeah it wasn't playing nicely with Cloudflare.

package.json
adamgall: After this gets merged and we are using CF fully in production, then there will be a cleanup PR to remove all Netlify stuff.

On this branch though, the "npm run dev" command starts up the Cloudflare local environment.

src/providers/App/hooks/useBalancesAPI.ts
adamgall: just a weird word lol plz suggest a better one

src/types/daoTreasury.ts
adamgall: Yeah I pulled it out because I needed it in `./functions/shared/moralisTypes.ts`

wrangler.toml
adamgall: Yeah i mean, it's all public anyway, they're shipped in the React bundle to all users.

But david brought up the same concerns.

There's a difference between having some keys ship quietly in a React bundle

vs

explicitly publishing them on github.

I still need to remove some of these API keys from `wrangler.toml` and rotate them. Is on my TODO list still.

src/providers/App/hooks/useBalancesAPI.ts
mudrila: `items` / `modules` / `features`/ `moralisChunks` / `chunks` - suggesting something like that

src/types/daoTreasury.ts
mudrila: Oh, gotcha - ty!

src/providers/App/hooks/useBalancesAPI.ts
adamgall: Hm fwiw this is a query param name. So it'll be in the url

adamgall:
@DarksightKellar @mudrila @Da-Colon @johnqh

I've factored this work a bit.

This PR now basically tracks the "base" branch for all of this `cloudflare` work. I decided to split it up amongst many PRs is to make it easier to review all of this work.

Please see yourself added as a review to a number of new PRs, which all build on each other.

functions/balances/index.ts
DarksightKellar: `class`? 
`type` (overused so not a fan)

oh 

I just noticed the `BalanceType` so why not just do `balance-type(s)`?

PR #2663, opened 2025-01-10
`[ENG-101]` Fix conditional logic preventing switch chain from working properly

Da-Colon:
Closes [ENG-101](https://linear.app/decent-labs/issue/ENG-101/dao-does-not-load-properly-when-on-wrong-network)

While I believe this setup isn't long (hopefully) for this codebase as discussed in @adamgall 's comments in #2658. it should still work. and was not taking in account the fact that mainnet is the default network for our network config. 

This would cause app to think everything was good to go and not switch the wallet's network...

Soooo this will fix that thanks for flagging @DarksightKellar 

## Testing
- Visit any sepolia DAO. (or be on home page of app and manually switch network to sepolia. and close the tab.

- Then in new tab visit `/home?dao=eth:0xB14AC30AA97057e2A934c47BDc45171335B91Bd8`.
- or just be on Home page and search and visit the above address while connected to sepolia. 

Everything should load correctly

### DIFF
--- a/src/hooks/utils/useAutomaticSwitchChain.ts
+++ b/src/hooks/utils/useAutomaticSwitchChain.ts
@@ -27,11 +27,7 @@ export const useAutomaticSwitchChain = ({
       return;
     }
     const chainId = getChainIdFromPrefix(urlAddressPrefix);
-    if (
-      (addressPrefix !== urlAddressPrefix || chainId !== walletChainId) &&
-      urlAddressPrefix !== undefined &&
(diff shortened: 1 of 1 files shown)


adamgall:
@Da-Colon thanks for identifying and fixing! please be sure to copy the "testing steps" into the Linear issue.

DarksightKellar (approved):

mudrila (approved):

adamgall (approved):

PR #2665, opened 2025-01-14
[Cloudflare] Basic Cloudflare Pages deployment

adamgall:
This branch implements the base deployment of the static Decent app onto Pages. It does not implement any "functions", so the Treasury requests will fail (as they're targeting Netlify functions which don't exist in the Cloudflare environment).

Please view the site at https://cf-add-base-cloudflare-deplo.decent-interface.pages.dev/


src/insights/hotjar.ts
adamgall, line 1: This was `undefined` (not set) on Cloudflare preview, and we weren't catching it properly.

src/insights/hotjar.ts
```diff
@@ -2,8 +2,8 @@ import Hotjar from '@hotjar/browser';
 
 export const initHotjar = () => {
   if (
-    import.meta.env.VITE_APP_HOTJAR_SITE_ID === '' ||
-    import.meta.env.VITE_APP_HOTJAR_VERSION === ''
+    import.meta.env.VITE_APP_HOTJAR_VERSION === undefined ||
+    import.meta.env.VITE_APP_HOTJAR_SITE_ID === undefined
```
DarksightKellar, line 6: Should we not check for both?

DarksightKellar (approved):

src/insights/hotjar.ts
adamgall: what are you talking about here?

PR #2674, opened 2025-01-15
Issue/env feature flags 3

adamgall:
Special `demo_mode` query param, with local storage persistence.

`?demo_mode=on` to turn it on. You'll see a button in the Footer.

Click the button (or do `?demo_mode=off`) to turn it off.

Also in this PR are a wrangler version bump and wrangler config cleanup to stop a warning in the console. And deleting an unused button which doesn't need to be in the codebase.


src/components/Roles/forms/RoleFormPaymentStream.tsx
```diff
@@ -216,27 +215,6 @@ export function RoleFormPaymentStream({ formIndex }: { formIndex: number }) {
...
-                    },
-                    startDate: addMinutes(nowDate, 1),
-                    endDate: addMinutes(nowDate, 10),
-                  });
-                }}
-              >
-                Ze stream ends in 10!
-              </Button>
-            )}
             {canBeCancelled && (
```
adamgall, line 218: This should have been deleted in your branch already, forgot to tell you. It's not needed any longer.

adamgall:
> The demo code are under "flag_demo". How does this hook into "flag_demo"?

Good eye! I _just_ pushed a couple more commits to address that oversight.

PR #2677, opened 2025-01-17
`[ENG-113 | Public Client]` Swap wagmi's `usePublicClient` with custom client

Da-Colon:
> Part 1 of 3.
    - Swap Public Client <------
    - Swap Wallet Client
    - Find and Update any other hooks

Swap out of usePublicClient with `useNetworkPublicClient` which keeps control of what network is interacted with based on the currently set network config.

## Other Updates
- removed `!publicClient` conditionals as no longer the case
- memoized the `createPublicClient` return (was causing problems otherwise)
- found a few places where public client was being used as a dependency but not being used.


src/hooks/utils/useAddress.ts
```diff
@@ -45,7 +38,7 @@ const useAddress = (addressInput: string) => {
       setIsLoading(false);
       return;
     }
-
+    // @todo this should be mainnet?
```
adamgall, line 41: at first glance, yes

adamgall (approved):
looks good ship it

mudrila (approved):

DarksightKellar (approved):
🤌🏾

PR #2679, opened 2025-01-17
`[ENG-113 | Various]` Other Wagmi Hooks

Da-Colon:
I thought there would be more. But some other places that needed the chain id passed in.

### DIFF
--- a/src/components/ui/forms/ABISelector.tsx
+++ b/src/components/ui/forms/ABISelector.tsx
@@ -27,9 +27,9 @@ interface IABISelector {
 
 export default function ABISelector({ target, onChange }: IABISelector) {
   const [abi, setABI] = useState<ABIElement[]>([]);
-  const { etherscanAPIUrl } = useNetworkConfigStore();
+  const { etherscanAPIUrl, chain } = useNetworkConfigStore();
   const { t } = useTranslation('common');
-  const { data: ensAddress } = useEnsAddress({ name: target?.toLowerCase() });
+  const { data: ensAddress } = useEnsAddress({ name: target?.toLowerCase(), chainId: chain.id });
   const client = useNetworkPublicClient();
 
   useEffect(() => {
--- a/src/hooks/utils/useGetAccountName.ts
+++ b/src/hooks/utils/useGetAccountName.ts
@@ -1,6 +1,7 @@
 import { useEffect, useState } from 'react';
 import { Address } from 'viem';
 import { useEnsName } from 'wagmi';
+import { useNetworkConfigStore } from '../../providers/NetworkConfig/useNetworkConfigStore';
 
 export const createAccountSubstring = (account: string) => {
   return `${account.substring(0, 6)}...${account.slice(-4)}`;
@@ -16,10 +17,12 @@ export const createAccountSubstring = (account: string) => {
  * @todo Should switch to object for props
  */
 export const useGetAccountName = (account?: Address | null, truncate?: boolean) => {
+  const { chain } = useNetworkConfigStore();
   if (truncate === undefined) truncate = true;
 
   const { data: ensName } = useEnsName({
     address: !!account ? account : undefined,
+    chainId: chain.id,
   });
 
   const [accountSubstring, setAccountSubstring] = useState<string>();


Da-Colon (commented):
So on top of these there is two other places I believe that chainId is passed into a ENS hook...

Should these point to mainnet only?

adamgall:
> So on top of these there is two other places I believe that chainId is passed into a ENS hook...
> 
> Should these point to mainnet only?

I think we should be defaulting all ENS calls and lookups to mainnet specifically, yes. thoughts @decentdao/engineering?

src/components/ui/forms/ABISelector.tsx
```diff
@@ -27,9 +27,9 @@ interface IABISelector {
 
 export default function ABISelector({ target, onChange }: IABISelector) {
   const [abi, setABI] = useState<ABIElement[]>([]);
-  const { etherscanAPIUrl } = useNetworkConfigStore();
+  const { etherscanAPIUrl, chain } = useNetworkConfigStore();
   const { t } = useTranslation('common');
-  const { data: ensAddress } = useEnsAddress({ name: target?.toLowerCase() });
+  const { data: ensAddress } = useEnsAddress({ name: target?.toLowerCase(), chainId: chain.id });
```
adamgall, line 32: Yeah so in the ABI selector here, i think it is correct to use the `chain` from `useNetworkConfigStore`. We don't want to default to Mainnet for this one.

src/hooks/useNetworkPublicClient.ts
```diff
@@ -9,7 +9,9 @@ export default function useNetworkPublicClient() {
       createPublicClient({
         chain,
         batch: {
-          multicall: true,
+          multicall: {
+            wait: 500,
+          },
```
adamgall, line 14: sure let's run it

src/hooks/utils/useGetAccountName.ts
```diff
@@ -16,10 +17,12 @@ export const createAccountSubstring = (account: string) => {
  * @todo Should switch to object for props
  */
 export const useGetAccountName = (account?: Address | null, truncate?: boolean) => {
+  const { chain } = useNetworkConfigStore();
   if (truncate === undefined) truncate = true;
 
   const { data: ensName } = useEnsName({
     address: !!account ? account : undefined,
+    chainId: chain.id,
```
adamgall, line 25: But here though, I think this is a place to use mainnet specifically for this call, not the chain from useNetworkConfigStore.

we're doing reverse lookups here, so we're trying to get human readable names. just display purposes only.

i think it's safe to leave this change as is, or hardcode mainnet. but i like leaving it as it is.

adamgall (approved):

src/hooks/useNetworkPublicClient.ts
mudrila: This will eventually mean 500ms waiting time for each next request since the last point of adding previous request in the batch - I think it's too large. I'd suggest reducing it to 200-300

src/hooks/utils/useGetAccountName.ts
mudrila: On some chains ENS is supported separately - like Sepolia / OP
So I'd let it fail under the hood instead of always displaying ENS from mainnet

mudrila (approved):
Overall - looks good, though I'd suggest lowering wait time for batching

src/hooks/useNetworkPublicClient.ts
Da-Colon: played with it some more agreed.

src/hooks/utils/useGetAccountName.ts
Da-Colon: This is what I was thinking about. maybe this should be a seperate issue. as we do have hardcoded mainnet is other places. maybe we should take a second and reconcile these different places.

adamgall: Yeah good call @Da-Colon . Although is this change here necessary for this PR? If not, can you remove it from this PR and create an issue?

Da-Colon: yeah the changes here keep things as they were in the app. I think its safe to leave this and update it from here.

DarksightKellar: Agree with @mudrila's take here

DarksightKellar (approved):

src/hooks/utils/useGetAccountName.ts
Da-Colon: Issue created https://linear.app/decent-labs/issue/ENG-137/ens-updates

PR #2681, opened 2025-01-21
Remove netlify

adamgall:
The rest of the stuff needed to fully remove netlify from codebase.


src/components/DaoCreator/StepController.tsx
adamgall, line 1: This is the only actual code change in this PR. This change happened in the `prepare-toasts` branch. That branch included this bugfix. The bug is that we were telling ALL TOASTS to dismiss, regardless of where they were initiated. In this specific context we should only be telling the disconnect toast to dismiss.

We also don't need to store the toast id in a ref, as far as i can tell.

PR #2685, opened 2025-01-23
`[Eng 137]` wagmi `useEns~` custom wrapper functions

Da-Colon:
Closes [ENG-137](https://linear.app/decent-labs/issue/ENG-137/ens-updates)


src/providers/NetworkConfig/useNetworkConfigStore.ts
```diff
@@ -9,6 +9,7 @@ interface NetworkConfigStore extends NetworkConfig {
 }
 
 export const supportedNetworks = Object.values(networks).sort((a, b) => a.order - b.order);
+export const supportedEnsNetworks = [networks.mainnet.chain.id, networks.sepolia.chain.id];
```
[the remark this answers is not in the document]
adamgall: I would rather see this as a boolean prop on the NetworkConfig objects

src/hooks/useNetworkEnsAddress.ts
```diff
@@ -0,0 +1,20 @@
...
+interface UseNetworkEnsAddressProps {
+  name?: string;
+  chainId?: number;
+}
+
+export function useNetworkEnsAddress(props: UseNetworkEnsAddressProps | undefined) {
+  const { chain } = useNetworkConfigStore();
+  const ensNetworkOrMainnet = supportedEnsNetworks.includes(props?.chainId ?? chain.id)
+    ? chain.id
+    : mainnet.id;
```
adamgall, line 17: Hm, just want to make sure I understand this ternary here.

then a call to this function might look like this:
- props.chainId = mainnet
- chain.id (from newtork config store hook) = sepolia

but then we'd be targeting sepolia in the wagmi hook call, despite passing in mainnet, to the outer hook's props. is that right?

src/hooks/useNetworkEnsAvatar.ts
```diff
@@ -0,0 +1,20 @@
...
+
+interface UseNetworkEnsAvatarProps {
+  name?: string;
+  chainId?: number;
+}
+
+export function useNetworkEnsAvatar(props: UseNetworkEnsAvatarProps | undefined) {
+  const { chain } = useNetworkConfigStore();
+  const ensNetworkOrMainnet = supportedEnsNetworks.includes(props?.chainId ?? chain.id)
+    ? chain.id
```
adamgall, line 16: Same logical questions here as previous comment

src/hooks/useNetworkEnsName.ts
```diff
@@ -0,0 +1,21 @@
...
+
+interface UseNetworkEnsNameProps {
+  address?: Address;
+  chainId?: number;
+}
+
+export function useNetworkEnsName(props: UseNetworkEnsNameProps | undefined) {
+  const { chain } = useNetworkConfigStore();
+  const ensNetworkOrMainnet = supportedEnsNetworks.includes(props?.chainId ?? chain.id)
+    ? chain.id
```
adamgall, line 17: Same question

src/hooks/useNetworkEnsAddress.ts
```diff
@@ -0,0 +1,20 @@
+import { mainnet } from 'viem/chains';
+import { useEnsAddress } from 'wagmi';
+import {
+  supportedEnsNetworks,
+  useNetworkConfigStore,
+} from '../providers/NetworkConfig/useNetworkConfigStore';
+
+interface UseNetworkEnsAddressProps {
+  name?: string;
```
adamgall, line 9: What does it mean that this can be optional? How could we do an ENS Address lookup without a name?

src/hooks/useNetworkEnsAvatar.ts
```diff
@@ -0,0 +1,20 @@
+import { mainnet } from 'viem/chains';
+import { useEnsAvatar } from 'wagmi';
+import {
+  supportedEnsNetworks,
+  useNetworkConfigStore,
+} from '../providers/NetworkConfig/useNetworkConfigStore';
+
+interface UseNetworkEnsAvatarProps {
+  name?: string;
```
adamgall, line 9: Same question as previous comment

src/hooks/useNetworkEnsName.ts
```diff
@@ -0,0 +1,21 @@
+import { Address } from 'viem';
+import { mainnet } from 'viem/chains';
+import { useEnsName } from 'wagmi';
+import {
+  supportedEnsNetworks,
+  useNetworkConfigStore,
+} from '../providers/NetworkConfig/useNetworkConfigStore';
+
+interface UseNetworkEnsNameProps {
+  address?: Address;
```
adamgall, line 10: Same question as previous comment

src/providers/NetworkConfig/useNetworkConfigStore.ts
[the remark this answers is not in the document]
Da-Colon: good call

src/hooks/useNetworkEnsAddress.ts
Da-Colon: o opps I see my mistake. 

Also I think instead of this safe guard I should throw an error. This isn't user entered and would be a mistake on our side if this isn't right.

Da-Colon: updated.

src/hooks/useNetworkEnsAddress.ts
Da-Colon: So to answer why the types are optional, since these are extending wagmi hooks, I kept the expected param types the same for these three files

Da-Colon:
@adamgall Updated and answer your questions. Soon as you approve these changes I'll bleed these through to the follow up PR. and make the same changes (I use the same ensormainnetchainid mistake there)

src/hooks/useNetworkEnsAddress.ts
Da-Colon: Well your comment made me rethink this. 'We' did something wrong if the network used in this hook is incorrect. so its a runtime safety for us. I could type safety instead?

DarksightKellar (approved):
lgtm

adamgall (approved):

PR #2686, opened 2025-01-23
`[Eng 137]` ENS publicClient async usage updates

Da-Colon:
Closes [ENG-137](https://linear.app/decent-labs/issue/ENG-137/ens-updates).

So as this PR shows we have two different patterns in our app, not saying a bad thing but pointing it out. 

We have a `hook` pattern, which on component mount the lookup is performed.

And updated in this PR, we have the `async` pattern, where we need to make the call more explicitly either mostly in use in form validation and form submission preparation. 

This PR brings these 'async' instances to use the same function with the supported ens network used.


src/hooks/useNetworkEnsName.ts
Da-Colon, line 1: main update

src/hooks/useNetworkEnsAddress.ts
Da-Colon, line 1: main update

src/components/DaoCreator/hooks/usePrepareFormData.ts
```diff
@@ -120,52 +121,50 @@ export function usePrepareFormData() {
     }: AzoriusERC20DAO<BigIntValuePair> & FreezeGuardConfigParam): Promise<
       AzoriusERC20DAO | undefined
     > => {
-      if (publicClient) {
```
Da-Colon, line 123: This large section was just me removing this conditional

src/hooks/DAO/useSearchDao.ts
```diff
@@ -42,17 +45,27 @@ export const useSearchDao = () => {
...
+
+      const resolvedAddressesSet = new Set(resolvedAddresses);
+      const mappedAddressesWithChainIds: ResolvedAddressWithChainId[] = [];
+      for (const network of supportedNetworks) {
+        for (const address of resolvedAddressesSet) {
+          mappedAddressesWithChainIds.push({ address, chainId: network.chain.id });
+        }
       }
+
+      await findSafes(mappedAddressesWithChainIds);
```
Da-Colon, line 68: A little bit of a refactor/update. So now it.

- attempts to resolve the address on supported ENS networks
- creates a Set of unique addresses
- creates an array of addresses with chainIds by looping through supported networks and then the unique addresses to check each of the supported networks for each of the addresses for a safe.

src/hooks/useNetworkEnsAddress.ts
```diff
@@ -18,3 +20,25 @@ export function useNetworkEnsAddress(props?: UseNetworkEnsAddressProps) {
 
   return useEnsAddress({ name: props?.name, chainId: ensNetworkOrMainnet });
 }
+
+export function useNetworkEnsAddressAsync() {
```
DarksightKellar, line 24: `useGetEnsAddress`, maybe? Doesn't really fit "async". IMO usage of this hook is more on the "sync" side -- you get a getter function that is called and must wait for a return.

src/hooks/useNetworkEnsName.ts
```diff
@@ -19,3 +21,23 @@ export function useNetworkEnsName(props?: UseNetworkEnsNameProps) {
 
   return useEnsName({ address: props?.address, chainId: ensNetworkOrMainnet });
 }
+
+export function useNetworkEnsNameAsync(props?: UseNetworkEnsNameProps) {
```
DarksightKellar, line 25: Same suggestion on naming here

DarksightKellar (commented):
A naming suggestion

src/hooks/useNetworkEnsAddress.ts
Da-Colon: I'm using a similar naming pattern to wagmi. `async` This tells that this version of the hook has a callback that returns a promise. 

hmm

DarksightKellar: Wait yeah you're right. It _is_ more async. Ah alright I'm good then

src/hooks/useNetworkEnsName.ts
adamgall: I'm wondering, should we be using `useEnsName` from `wagmi` at all any more? My gut says no.

src/hooks/DAO/useSearchDao.ts
```diff
@@ -42,17 +45,27 @@ export const useSearchDao = () => {
...
-        await findSafes(
-          supportedNetworks.map(network => ({
-            address: resolvedAddress,
-            chainId: network.chain.id,
-          })),
-        );
-      } else {
+      const resolvedAddresses: Address[] = [];
+      for (const chainId of supportedEnsNetworks) {
+        const { resolvedAddress, isValid } = await resolveENSName(input, chainId);
```
adamgall, line 50: Do these async `resolveENSName` calls happen in sequence? We should make sure they happen in parallel.

src/hooks/useNetworkPublicClient.ts
```diff
@@ -1,21 +1,25 @@
 import { useMemo } from 'react';
 import { createPublicClient, http } from 'viem';
 import { useNetworkConfigStore } from '../providers/NetworkConfig/useNetworkConfigStore';
+interface UseNetworkPublicClientProps {
+  chainId?: number;
```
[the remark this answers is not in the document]
adamgall: Actually it looks like you can delete this entire props interface. `useNetworkPublicClient` is never called with any params.

```ts
export default function useNetworkPublicClient() {
  const { chain, rpcEndpoint } = useNetworkConfigStore();

  const publicClient = useMemo(
    () =>
      createPublicClient({
        chain,
        batch: {
          multicall: {
            wait: 200,
          },
        },
        transport: http(rpcEndpoint),
      }),
    [chain, rpcEndpoint],
  );
  return publicClient;
}
```

adamgall (changes requested):
eh i should have marked my last review as "Request changes". Doing that here. Please see my last couple sets of comments.

adamgall (approved):

PR #2708, opened 2025-02-03
Eng 61 update subgraph endpoints

adamgall:
Closes https://linear.app/decent-labs/issue/ENG-61/update-subgraph-endpoints
Closes https://linear.app/decent-labs/issue/ENG-97/remove-network-dependencies-from-build-process
Closes https://linear.app/decent-labs/issue/ENG-182/deployed-a-polygon-dao-does-not-load


src/components/DaoHierarchy/DaoHierarchyNode.tsx
```diff
@@ -31,6 +31,36 @@ import { BarLoader } from '../ui/loaders/BarLoader';
...
+    hierarchy: DAOHierarchyNode[];
+    proposalTemplatesHash: string | null;
+  }[];
+}
+
+interface DAOQueryVariables {
+  safeAddress: string;
+}
+
+export interface HierarchyChild {
```
DarksightKellar, line 60: Doesn't seem like this needs to be exported

package.json
```diff
@@ -65,11 +65,7 @@
     "dev": "vite --force",
     "preview": "vite preview",
     "build": "vite build && vite build --mode page-function",
-    "build:cf": "vite build && vite build --mode page-function",
-    "graphql:build": "graphclient build",
-    "graphql:dev-server": "graphclient serve-dev",
-    "test": "vitest --dir=test",
-    "postinstall": "npm run graphql:build"
```
Da-Colon, line 72: 🔥

src/components/DaoHierarchy/DaoHierarchyNode.tsx
```diff
@@ -181,8 +192,8 @@ export function DaoHierarchyNode({
         setHierarchyNode(_node);
       });
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
```
Da-Colon, line 185: 👀 sweet if we don't need this anymore. tested Myosin good stress tester for org page.

src/graphql/index.ts
```diff
@@ -1,10 +1,57 @@
...
+
+  const url = import.meta.env.DEV
+    ? `https://api.studio.thegraph.com/query/${config.space}/${config.slug}/version/latest`
+    : `https://gateway.thegraph.com/api/${theGraphAPIKey}/subgraphs/id/${config.id}`;
+
+  // Check if we already have a client for this URL
+  const cachedClient = clientCache.get(url);
+  if (cachedClient) {
+    return cachedClient;
+  }
```
Da-Colon, line 18: Why do we cache the client? looking at the documentation, and is this instead of create a context provider?

src/hooks/DAO/loaders/useFractalGovernance.ts
```diff
@@ -146,14 +158,15 @@ export const useFractalGovernance = () => {
     type,
   ]);
 
+  const proposalsLoadKey = useRef<string>();
```
Da-Colon, line 161: ➕

src/store/roles/rolesStoreUtils.ts
```diff
@@ -346,7 +337,7 @@ const getPaymentStreams = async (
     });
 
     const streamsWithCurrentWithdrawableAmounts: SablierPayment[] = await Promise.all(
-      formattedLinearStreams.map(async stream => {
+      formattedLinearStreams.map(async (stream: any) => {
```
Da-Colon, line 340: you mean to leave as type `any`?

src/store/roles/rolesStoreUtils.ts
```diff
@@ -268,30 +268,21 @@ const getRoleHatTerms = async (
...
+    if (!streamQueryResult.data?.streams.length) {
       return [];
     }
     const secondsTimestampToDate = (ts: string) => new Date(Number(ts) * 1000);
     const lockupLinearStreams = streamQueryResult.data.streams.filter(
-      stream => stream.category === 'LockupLinear',
+      (stream: { category: string }) => stream.category === 'LockupLinear',
     );
-    const formattedLinearStreams = lockupLinearStreams.map(lockupLinearStream => {
+    const formattedLinearStreams = lockupLinearStreams.map((lockupLinearStream: any) => {
```
Da-Colon, line 285: you mean to leave as type `any`?

src/graphql/index.ts
adamgall: Decided to implement this lil caching mechanism (thanks for the suggestion @mudrila), specifically _because_ I've removed the context provider.

Wherever we need to utilize a graphql client, we create (or get from cache) an instance of it on the fly, instead of using the "one" that would be in a Provider context.

Decided to do it like this because of how diverse our graphql clients are. Effectively the "url" property of a `urql` client instance is the thing that makes it unique, and given that there are 3 dimension in which this "url" can differ (dev vs prod (2), decent vs sablier (2), all the networks (5)), there are 2 * 2 * 5 = 20 different potential endpoints for the urql client. Rather than just include _one_ of those client instances in a Provider context, i said fuggit and we just create one (or pull from cache) as needed.

I didn't really do any profiling to see how expensive it is to create these different clients, but the caching layer here is so light it seemed like a simple thing to implement.

src/store/roles/rolesStoreUtils.ts
adamgall: Ah shit, good point. I'm realizing now that there's no typing for the result of these queries...

sigh, I'll need to get back into this.

adamgall: oke pushed some more code

src/store/roles/rolesStoreUtils.ts
adamgall: fixed

Da-Colon (approved):
Good deal. Send it

Da-Colon:
This also fixes an issue that was found when doing a subgraph lookup on polygon https://linear.app/decent-labs/issue/ENG-182/deployed-a-polygon-dao-does-not-load

PR #2714, opened 2025-02-05
`[ENG-190]` Nonce typing from Safe incorrect (develop)

Da-Colon:
Closes [ENG-190](https://linear.app/decent-labs/issue/ENG-190/multisig-cannot-execute-proposal)

Traced the weird conditional failure all the way back to the nonce we receive from Safe's API. The API's `SafeInfoResponse` type for nonce `number` but actually typed a `string` on the response...Something has changed as of like yesterday?

This change is basically causing any `===` comparison to fail comparing against the proposal's nonce

Edit: Updated the SafeApi SDK to latest reveals this change to `SafeInfoResponse`. nonce is indeed a string


adamgall (approved):

Da-Colon:
@adamgall I need to rebase this to develop and remove the bump in version

PR #2718, opened 2025-02-06
Fix DeFi Positions fetching and transforming

adamgall:
Closes https://linear.app/decent-labs/issue/ENG-196/moralis-defi-positions-api-is-currently-broken

- Update Moralis API to support non-paginated endpoints
- Enhance DefiResponse transformer to map Moralis response to a more structured format
- Add fetchMoralisDefi helper function for DeFi-specific API calls
- Improve type definitions for DeFi-related responses

### DIFF
--- a/functions/shared/moralisApi.ts
+++ b/functions/shared/moralisApi.ts
@@ -13,6 +13,7 @@ interface MoralisRequestConfig {
   chain: string;
   apiKey: string;
   params?: Record<string, string>;
+  isPaginated?: boolean;
 }
 
 export async function fetchMoralis<T>({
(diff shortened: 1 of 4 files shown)


functions/shared/moralisApi.ts
```diff
@@ -57,16 +59,24 @@ export async function fetchMoralis<T>({
...
-    const data: MoralisResponse<T> = await response.json();
+    const data = await response.json();
 
-    // Handle no results
-    if (!data || !data.result) {
-      break;
+    // Handle paginated vs non-paginated responses
+    if (isPaginated) {
+      const paginatedData = data as MoralisResponse<T>;
+      if (!paginatedData || !paginatedData.result) break;
```
DarksightKellar, line 67: In case something weird happens like Moralis changing their API for whatever reason, an error log here before the `break` would be helpful for faster bug ID'ing. Cz this shouldn't ever happen ideally right?

adamgall: correct. updated.

Da-Colon (approved):
SIck. honestly didn't realize that section existed there. Tested and reviewed

PR #2721, opened 2025-02-07
Show banner for when user delegated tokens after a proposal was made

DarksightKellar:
Screenshot 2025-02-07 at 17 19 13

See issue for testing instructions

I've also:
- Removed the ability to hide voting power
- Done some refactoring to make it it easier to implement redesigned proposal summary cards
- Removed an apparently redundant `expandedView`


src/types/daoProposal.ts
```diff
@@ -62,15 +62,15 @@ export type ERC721ProposalVote = {
...
   {
-    label: 'no',
+    label: 'Reject',
     value: 0,
   },
   {
-    label: 'abstain',
+    label: 'Abstain',
     value: 2,
   },
```
adamgall, line 75: What are these changes for?

adamgall: The lowercase "yes/no/abstain" are translation string keys though right?

DarksightKellar: Oh... dang. Duh. Will update, thanks

src/components/Proposals/ProposalSummary.tsx
```diff
@@ -125,201 +361,40 @@ export function AzoriusProposalSummary({ proposal }: { proposal: AzoriusProposal
...
-      alignItems="flex-start"
-      variant="text"
-      color="celery-0"
-      _active={{ color: 'celery--2' }}
-      onClick={toggleShowVotingPower}
-    >
-      {showVotingPower ? proposalVotingWeight : t('show')}
-    </Button>
-  );
+  const notEnoughVotingPowerAtTheTimeOfProposalCreation =
```
Da-Colon, line 366: now if thats not an example of clear variable naming I don't know what is.

src/components/Proposals/ProposalSummary.tsx
```diff
@@ -73,7 +313,6 @@ export function AzoriusProposalSummary({ proposal }: { proposal: AzoriusProposal
             address: tokenAddress,
             client: publicClient,
           });
-          // @todo We should be checking proposal state - if it's active, we should use the latest block, otherwise we should calculate the voting weight based on the startBlock and deadlineMs
```
Da-Colon, line 76: is this TODO no longer relavent?

src/components/Proposals/ProposalSummary.tsx
DarksightKellar: lmaooo

src/components/Proposals/ProposalSummary.tsx
DarksightKellar: hm I don't remember. I'll take a look

adamgall (changes requested):
yes/no/abstain translation strings need to be updated

src/types/daoProposal.ts
DarksightKellar: done

src/components/Proposals/ProposalSummary.tsx
DarksightKellar: Actually this todo seems weird. Why would we ever want to use the latest block to check voting weight on a proposal? Is it not always determined by the weight _at the time of proposal creation_? And who cares about proposals not active here? If inactive, voting weight has no relevance, no? Am I missing sth?

@decentdao/engineering

adamgall: I agree that the comment seems off to me. Unless I'm missing something, if the intention is to determine the voting power of an address _within the context of a proposal_, we would always want to determine voting power at that proposal's creation block.

However, if the intention is to determine the voting power of an address _outside the context of a specific proposal, for example, on the home page or proposal's list page, something like "Your current voting power is xxx", then yeah we would want to use the current block as that anchor.

adamgall:
@DarksightKellar you didn't add `'approve'` or `'reject'` translation strings in the translations files. I also see that the "Breakdown" section which shows the horizontal bar graphs are still labeled as "yes/no/abstain", instead of the language used in the new radio selection "approve/reject/abstain". @decentdao/design

DarksightKellar:
> @DarksightKellar you didn't add `'approve'` or `'reject'` translation strings in the translations files. I also see that the "Breakdown" section which shows the horizontal bar graphs are still labeled as "yes/no/abstain", instead of the language used in the new radio selection "approve/reject/abstain". @decentdao/design 

@adamgall They _are_ there already. Dunno who added them or when, but they're in "common.json" but never used apparently. Also the breakdown section maintains the yes/no copy

adamgall:
> > @DarksightKellar you didn't add `'approve'` or `'reject'` translation strings in the translations files. I also see that the "Breakdown" section which shows the horizontal bar graphs are still labeled as "yes/no/abstain", instead of the language used in the new radio selection "approve/reject/abstain". @decentdao/design
> 
> @adamgall They _are_ there already. Dunno who added them or when, but they're in "common.json" but never used apparently.
> 
> Also the breakdown section maintains the yes/no copy (edit: ah I just noticed you tagged design. I assumed the copy diff was intentional, but it might have been missed, tbf)

Oh yes I see, sorry for jumping the gun on those translation strings.

Still looking for @decentdao/design to confirm that the "Breakdown" should use "yes/no/abstain" while the vote buttons use "approve/reject/abstain".

src/components/Proposals/ProposalSummary.tsx
DarksightKellar: In the context of proposal summary (this file) then, I think only the first scenario matters. Is there ever a user flow where the second is intended? The user's always looking at their voting power when about to vote

adamgall: > In the context of proposal summary (this file) then, I think only the first scenario matters.

Yep

> BTW is there ever a user flow where the second is intended?

Offhand, I don't think so.

nicolaus-sherrill:
@adamgall @DarksightKellar disregard "approve", "reject" text strings for now. We did not work on that copy in this effort so that is likely an artifact of design-to-prod gap we're still closing.

adamgall (approved):

nicolaus-sherrill (approved):

DarksightKellar:
Gonna merge this

PR #2727, opened 2025-02-11
Fix polygon

DarksightKellar:
Fixes a regression(?) on polygon causing polygon dao subgraph data to not load. Happens only on development since a different url is used in deployed sites.

Also fix an infinite reload bug directly resulting from the failed data load.

### DIFF
--- a/src/providers/NetworkConfig/networks/polygon.ts
+++ b/src/providers/NetworkConfig/networks/polygon.ts
@@ -28,7 +28,7 @@ export const polygonConfig: NetworkConfig = {
   isENSSupported: false,
   decentSubgraph: {
     space: 71032,
-    slug: 'fractal-base-polygon',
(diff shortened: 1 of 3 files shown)


src/providers/NetworkConfig/networks/polygon.ts
```diff
@@ -28,7 +28,7 @@ export const polygonConfig: NetworkConfig = {
   isENSSupported: false,
   decentSubgraph: {
     space: 71032,
-    slug: 'fractal-base-polygon',
+    slug: 'fractal-polygon',
```
adamgall, line 31: Yeah wow looks like I introduced this bug 10 months ago!

b26c2db0517fb208775940f1619ecdcefac6c032

adamgall (approved):

Da-Colon (approved):

PR #2732, opened 2025-02-18
Isolate and co-locate dev mode code

DarksightKellar:
Inspired by [this comment](https://github.com/decentdao/decent-interface/pull/2707#discussion_r1958702131). Isolate and co-locate dev mode code.


src/utils/dev/DevSet10MinPaymentStream.tsx
```diff
@@ -0,0 +1,26 @@
+import { Button } from '@chakra-ui/react';
+import { useFormikContext } from 'formik';
+import { isFeatureEnabled } from '../../helpers/featureFlags';
+import { RoleFormValues } from '../../types/roles';
+
+export function DevSet10MinPaymentStream({ formIndex }: { formIndex: number }) {
+  const { values, setFieldValue } = useFormikContext<RoleFormValues>();
+
+  return isFeatureEnabled('flag_dev') ? (
```
DarksightKellar, line 9: Doing this on the component level for cleaner and leaner looking code. Avoid scattering `isFeatureEnabled('flag_dev') &&` everywhere.

src/utils/dev/devModeConstants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEV_VOTING_PERIOD_DAYS = 0.0021;
+export const DEV_TIMELOCK_PERIOD_DAYS = 0;
```
DarksightKellar, line 2: `DEV-` prefix on these exports for that tiny extra clarity when scanning code

adamgall (commented):
I don't really agree with the intention of this PR.

Is it really better if we move the "dev mode behavior" implementation away from where it's actually be used?

DarksightKellar:
> I don't really agree with the intention of this PR.
> 
> Is it really better if we move the "dev mode behavior" implementation away from where it's actually be used?

@adamgall  Do you disagree with the isolation bit, the co-location bit, or both?

Reasons on
- Isolation: really just to avoid mucking up implementation code one actually cares about
- Co-location: make it super straightforward to add, but especially delete, dev mode code. Plus I don't wanna look at dev files while browsing the directory. 

Both are really just a readability/scannability thing.

TBH I'm not gonna defend this to the grave, but why do you think it would be worse (or not better) to do this?

adamgall:
I'm also not gonna defend my position very much. I don't have any good reasons other than it was just a gut feeling of mine.

@Da-Colon @johnqh what are your thoughts?

Da-Colon:
So for me the main pro is readability; with specifically larger UI or buttons that are intertwined with implementation code.  As well as ability to quickly test things on develop/prod if need be.

Personally I create these all the time but I just stash them locally, as these kinda things are only useful to me at the time I'm testing something particular. But don't keep them because they can depreciate pretty easily and just another thing to update. So I can take it or leave it, but I'll probably delete if I have to update it :D

Any variable that is particular to dev mode should 100% be a named constant. Shouldn't have these floating values without any explanations.

adamgall (approved):
Heya @DarksightKellar, I took the liberty of pushing a couple more commits to this branch:

- Removed `flag_yelling`, was just used as an example before actual real flags showed up
- Moved the "10 minutes" constant for "dev mode payment stream duration" to the "dev mode constants" file.
- Removed the "dev mode default timelock duration" constant. Timelock duration now defaults to 1 day.
- Renamed the "dev mode voting duration" constant to reference "minutes" instead of "days", so that the value is more human readable ("3 minutes" vs "0.0021 days")

I am ok with this PR for now then :)

DarksightKellar:
> Heya @DarksightKellar, I took the liberty of pushing a couple more commits to this branch:
> 
> * Removed `flag_yelling`, was just used as an example before actual real flags showed up
> * Moved the "10 minutes" constant for "dev mode payment stream duration" to the "dev mode constants" file.
> * Removed the "dev mode default timelock duration" constant. Timelock duration now defaults to 1 day.
> * Renamed the "dev mode voting duration" constant to reference "minutes" instead of "days", so that the value is more human readable ("3 minutes" vs "0.0021 days")
> 
> I am ok with this PR for now then :)

LGTM!

PR #2741, opened 2025-02-24
Add app down banner

DarksightKellar:
Screenshot 2025-02-24 at 17 37 28

Screenshot 2025-02-24 at 17 37 39

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "decent-interface",
-  "version": "0.7.0",
+  "version": "0.7.1",
   "private": true,
   "dependencies": {
     "@amplitude/analytics-browser": "^2.11.1",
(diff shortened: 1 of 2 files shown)


adamgall:
Just looking at the code on my phone right now, but first thought is that the slide effect is probably overkill

adamgall:
@nicolaus-sherrill @parkermccurley @johnqh any suggestions for what the copy should say here?

DarksightKellar:
> Just looking at the code on my phone right now, but first thought is that the slide effect is probably overkill

@adamgall lool I know, I know, I just couldn't help it. Doesn't look bad though and meant to be reverted anyway, so

Also, I've updated description with screenshots

adamgall:
> Doesn't look bad though

and yeah seeing it in action, the slide is nice

PR #2743, opened 2025-02-25
Load Safe Data from Contracts instead of Safe API

adamgall:


src/hooks/DAO/useSearchDao.ts
```diff
@@ -25,8 +25,15 @@ export const useSearchDao = () => {
 
   const findSafes = useCallback(
     async (resolvedAddressesWithChainId: { address: Address; chainId: number }[]) => {
+      /*
+      This function only checks if the address is on any of the EVM networks. 
+      The same safe could of on multiple networks
```
DarksightKellar, line 30: Hmm I don't think so. See `mappedAddressesWithChainIds` setup on lines 70-74. There will be repeated addresses for each network and the search will happen for each.

PR #2746, opened 2025-02-26
Safe subgraph queries

adamgall:
Beginnings of implementation of Subgraph Query to determine whether or not an address is a Safe

### DIFF
--- a/src/providers/App/hooks/useSafeAPI.ts
+++ b/src/providers/App/hooks/useSafeAPI.ts
@@ -21,6 +21,8 @@ import {
 } from 'viem';
 import GnosisSafeL2Abi from '../../../assets/abi/GnosisSafeL2';
 import { SENTINEL_ADDRESS } from '../../../constants/common';
+import { createDecentSubgraphClient } from '../../../graphql';
+import { SafeQuery } from '../../../graphql/SafeQueries';
 import { SafeWithNextNonce } from '../../../types';
 import { NetworkConfig } from '../../../types/network';
 import { useNetworkConfigStore } from '../../NetworkConfig/useNetworkConfigStore';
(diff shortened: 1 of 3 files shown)


src/providers/App/hooks/useSafeAPI.ts
```diff
@@ -141,31 +143,40 @@ class EnhancedSafeApiKit extends SafeApiKit {
...
+    //     safeAddress,
+    //     '/transactions/creation',
+    //   );
+
+    //   return { ...response, singleton: response.masterCopy };
+    // } catch (error) {
+    //   console.error('Error fetching getSafeCreationInfo from safe-client:', error);
+    // }
+
+    // TODO: this is what needs to be tested
```
adamgall, line 168: This is a good place to start.

Find out where in the codebase is calling this `getSafeCreationInfo` function, to trigger it to run, and test that we can get the correct data from these subgraph graphql calls.

You can use the Playground to verify that data is actually there and available:

- [Sepolia](https://thegraph.com/explorer/subgraphs/GNsEo1grZ1dUyyPbPnAutkDtx51t1XN8DhgMHThgo7b2?view=Query&chain=arbitrum-one)
- [Mainnet](https://thegraph.com/explorer/subgraphs/EwAjJixmcVMcxoj1qt1o2oae1rbvvZNdfshkfovLrbxp?view=Query&chain=arbitrum-one)
- [Base](https://thegraph.com/explorer/subgraphs/BWv5n6PPqv99y6vAez5M9jNvmKagpSpKprYPoKMb8z4o?view=Query&chain=arbitrum-one)

Check out the query in `SafeQueries.ts` to see the one we're using.

When I left off, the subgraph calls were returning strange "Message: Not Found" errors.

DarksightKellar: I'm not getting the error now. Seems to work for me, so I can search on mainnet on sepolia.

This DAO: `0xF222823699a84A6Dcd47A1A860554A5B9DAaaf47` should exist on Base though, but the query returns nothing

adamgall: Were you able to figure out why, @DarksightKellar ?

adamgall:
Closing this PR in favor of https://github.com/decentdao/decent-interface/pull/2747

PR #2755, opened 2025-03-01
Delete unused types

adamgall:
Remove some unused types from the codebase.

No functional changes.


src/types/fractal.ts
```diff
@@ -302,6 +301,11 @@ export interface VotingStrategyAzorius extends VotingStrategy {
   strategyType?: VotingStrategyType;
 }
 
+interface BIFormattedPair {
+  value: bigint;
+  formatted?: string;
+}
+
```
adamgall, line 308: This was the only thing which was used in `votingFungibleToken.ts`, so I just moved it over to this file (it's only usage)

DarksightKellar (approved):

PR #2764, opened 2025-03-06
ENG-334 isOwner fails cause the owners addresses have different cases from wallet address

johnqh:


adamgall (commented):
@johnqh, i would recommend closing this PR in favor of mine: https://github.com/decentdao/decent-app/pull/2765

The underlying issue here is that the `safe-transaction-service` (which just re-opened up on sepolia) is now returning data in a different shape than we expect.

concretely, here, it used to return checksummed addresses, now it returns all lowercased addresses (or maybe vice versa).

my PR addresses these issues at the source, by simply ripping out all calls to `safe-transaction-service` in places where we have robust fallbacks to either (or both) `safe-client` and reading onchain data.

in addition to fixing the issue that this PR addresses, my PR fixes another bug I ran into with multisig transactions, from the same root cause of data type mismatches (`nonce` as a string vs number). I wouldn't be surprised if there are other weird bugs in our app resulting from modified objects from the newly improved `safe-transaction-service`.

adamgall:
@johnqh thanks for approving https://github.com/decentdao/decent-app/pull/2765, i'll take that to mean you are ok with that PR superseding this one, so I'll close this PR.

PR #2771, opened 2025-03-11
Feature/eng 331 Use iframe injected (safe app) way in a new template

twodam:
### Summary: 
Use iframe to access dApps, provide Safe address as if it's connected to that site and then intercept final transaction data which will be attached to proposal.
Can try with this path: `/proposals/new/iframe?dao=eth:0xD26c85D435F02DaB8B220cd4D2d398f6f646e235& flag_iframe_template=on`

### Changes:

- new template that can load dApps in iframe and parse transactions in form
- new component SafeInjectIframeCard that does mostly core function, including communicate with dApp in iframe, intercept `eth_sendTransactions` and return true to tell dApp their request has been confirmed.
- extract useABI hook from ABISelector component
- update related string in `proposalTemplate.json` for i18n
- add an analytic event for opening iframe template page

### Screenshot:
image
image

### Limitation: 
This approach works because dApps support the Safe app, and it fail to work if that dApp set CSP and our domain is not in that list, check error below.
image

### Progress

- [X] integrate cowswap widget directly
- [X] port SafeInjectIframeCard from nance
- [X] intercept final transaction
- [X] put in correct formik values
- [X] test proposal publishing with a sepolia DAO that has assets (https://app.decentdao.org/proposals/0xc71eb239e7ea40ded935cec5c23665dd502f426281a9fb55c5788c1324ffb0f8?dao=sep:0x4E413477Cbdc696b568f4b963448cB645D96323b)
- [ ] get design and improve styling
- [ ] get review


src/components/SafeInjectIframeCard/helpers/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+    return generateId(10);
+  }
+
+  return new Date().getTime().toString(36);
+};
+
+export function getChainIdFromName(chainName: string | undefined) {
+  const chainMap: { [key: string]: number } = {
+    mainnet: mainnet.id,
+    gnosis: gnosis.id,
```
jigglyjams, line 27: why gnosis?

twodam: this was old code from nance which is no longer needed, I will remove this function

jigglyjams:
when i checkout this branch and then `npm install` it breaks some `Address` types. For example `Avatar.tsx`

maybe we need to migrate `@walletconnect/web3wallet` to new version per there notice

https://docs.reown.com/walletkit/upgrade/from-web3wallet-web

twodam:
this PR will be split in multiple smaller PRs.

PR #2779, opened 2025-03-18
`[ENG-250]` Allow Error text to show as design in organization page

Da-Colon:
Just unlocked the UI that was already there to handle when data isn't returned. I guess its hard without some extra logic to decide tell if this is a 'import' with no graph data and a error happening with the graph request. 

## Screenshot
localhost_3000_hierarchy_dao=base_0x4F6F99828B204bBe6F7a2D014166537e37aCb82C(Desktop)


Da-Colon:
@decentdao/design @johnqh This is kinda a catch/all error message...we should probably update the copy at least. Thoughts?

PR #2780, opened 2025-03-18
`[ENG-238]` Update proposal submission voting strategy error handling

Da-Colon:
Closes [ENG-238](https://linear.app/decent-labs/issue/ENG-238/sub-dao-cannot-create-proposals)

There was a check for `isProposer` that was being called on all 'voting strategies'. This includes `ERC20FreezeVoting` which doesn't have this method. wrapped this section in a try catch.


Da-Colon:
This fixes the issues but has me curious why this is here in the first place? There is no `isProposer` method on it so why is it in the loop?

PR #2783, opened 2025-03-19
`gasless-voting-4` | Enable Gasless on new dao

DarksightKellar:
Closes ENG-152
Closes ENG-142


package.json
```diff
@@ -55,7 +56,7 @@
     "sonner": "^1.5.0",
     "urql": "^4.2.1",
     "use-between": "^1.3.5",
-    "viem": "2.23.2",
+    "viem": "^2.23.2",
```
Da-Colon, line 59: wasn't this caret removed on purpose?

adamgall: the carrot does currently exist on `develop`, which IMO is fine. the carrot just means that _if_ vite's minor/patch versions need to be automatically bumped, in order to be compatible with some other package, then it can be.

the problems we were running into before is that currently the permissionless package (v0.2.37) NEEDS viem to be anchored at 2.23.2. so we did need to remove the carrot in order to get that specific version installed, but after we got viem 2.23.2 installed, the carrot should have been added back. there shouldn't have been any commits with the carrot removed.

i think the carrot should exist because at some point in the future permissionless might relax its strict dependency on viem 2.23.2, and if/when that happens, we shouldn't be artificially restricting viem to be stuck on 2.23.2.

Da-Colon: 👍🏽 Sounds good

src/hooks/utils/useInstallVersionedVotingStrategy.ts
```diff
@@ -180,21 +181,21 @@ export const useInstallVersionedVotingStrategy = () => {
       );
 
       const encodedStrategySetupData = encodeFunctionData({
-        abi: abis.LinearERC721Voting,
+        abi: LinearERC721VotingV1Abi, // @todo: use the deployed abi
         functionName: 'setUp',
         args: [encodedStrategyInitParams],
       });
 
-      const deployWhitelistingVotingStrategyTx = {
```
Da-Colon, line 188: confused me for a second, but probably fixing a copy/paste rename

src/pages/dao/settings/general/SafeGeneralSettingsPage.tsx
```diff
@@ -147,10 +151,27 @@ export function SafeGeneralSettingsPage() {
...
+      }
+
+      const paymasterCode = await publicClient.getCode({
+        address: predictedPaymasterAddress,
+      });
 
-      if (!paymasterAddress) {
-        // Create a new paymaster
+      if (!paymasterCode || paymasterCode === '0x') {
+        // Paymaster does not exist, deploy a new one
```
Da-Colon, line 174: I guess this answers my question, with whether the `getPaymasterAddress` method just returns a predicted address. Cool beans.

PR #2784, opened 2025-03-20
`ENG-361` take dApp transaction, decode and then create a proposal

twodam:
### Changes

- new modal `ConfirmTransactionModal` that list all transactions from dApp, navigate to proposalWithAction page if user choose to continue.
- new callback `onTransactionsReceived` in props of `SafeInjectProvider`
- updated `IframeModal` to decode transactions within `onTransactionReceived` and then open `ConfirmTransactionModal`
- new action and action card for `DAPP_INTEGRATION`, which display dapp name. (it's now only app url, once we got `dapp.json` imported in later PR, we can display app name easily.)

#### 1. Confirm transaction modal
image

#### 2. Create proposal page with action attached (dApp transactions)
image

### DIFF
--- a/src/components/ProposalBuilder/ProposalActionCard.tsx
+++ b/src/components/ProposalBuilder/ProposalActionCard.tsx
@@ -7,6 +7,7 @@ import { useFractal } from '../../providers/App/AppProvider';
 import { useProposalActionsStore } from '../../store/actions/useProposalActionsStore';
 import { CreateProposalAction, ProposalActionType } from '../../types/proposalBuilder';
 import { Card } from '../ui/cards/Card';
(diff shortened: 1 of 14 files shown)


jigglyjams (commented):
I dont see how to get the CoW swap info into the proposal. What am I missing?

jigglyjams (approved):

twodam:
> I dont see how to get the CoW swap info into the proposal. What am I missing?
> 
> I get this when I use the iframe from template image

we talked in Slack, here's notes for the record:

- check Etherscan API Key in `.env.local`
- check if node is correct version by using `nvm use`

PR #2791, opened 2025-03-21
`[ENG-250]` Show organization page with imported DAO

Da-Colon:
Closes [ENG-250](https://linear.app/decent-labs/issue/ENG-250/imported-sarcophagus-safe-parent-dao-loads-fine-but-not-subdao#comment-6c4e9614)

Che made a good point that this wasn't quite complete. looking at the code further noticed some redundant error handling going on.

## Screenshot
localhost_3000_hierarchy_dao=base_0x4F6F99828B204bBe6F7a2D014166537e37aCb82C(Desktop) (1)

### DIFF
--- a/src/components/DaoHierarchy/DaoHierarchyNode.tsx
+++ b/src/components/DaoHierarchy/DaoHierarchyNode.tsx
@@ -126,10 +126,6 @@ export function DaoHierarchyNode({
           safeAddress: _safeAddress,
         });
 
-        if (queryResult.error) {
-          throw new Error('Query failed');
-        }
-
         if (!queryResult.data) {
           return {
             parentAddress: null,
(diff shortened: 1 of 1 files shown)


src/components/DaoHierarchy/DaoHierarchyNode.tsx
```diff
@@ -126,10 +126,6 @@ export function DaoHierarchyNode({
           safeAddress: _safeAddress,
         });
 
-        if (queryResult.error) {
```
DarksightKellar, line 129: Is this handled somewhere else or?

Da-Colon: Actually yeah let me take a closer look at this and make sure this is taken care of correctly

Da-Colon: So yeah I put back this. but deleted the other early return. We were already handling the case of no data below.

DarksightKellar (approved):

adamgall (approved):

PR #2798, opened 2025-03-24
`gasless-voting-6` | cleanup 2

DarksightKellar:
Allow connected wallet to refill paymaster directly without having to go through a proposal
Closes ENG-403

### DIFF
--- a/src/components/GaslessVoting/GaslessVotingToggle.tsx
+++ b/src/components/GaslessVoting/GaslessVotingToggle.tsx
@@ -3,6 +3,7 @@ import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate } from 'react-router-dom';
 import { getContract } from 'viem';
+import { useWalletClient } from 'wagmi';
 import { EntryPoint07Abi } from '../../assets/abi/EntryPoint07Abi';
 import { DETAILS_BOX_SHADOW } from '../../constants/common';
 import { DAO_ROUTES } from '../../constants/routes';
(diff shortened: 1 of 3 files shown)


src/components/GaslessVoting/GaslessVotingToggle.tsx
```diff
@@ -147,13 +148,31 @@ export function GaslessVotingToggleDAOSettings(props: GaslessVotingToggleProps)
   }, [entryPointv07, paymasterAddress, publicClient]);
 
   const { addAction } = useProposalActionsStore();
+  const { data: walletClient } = useWalletClient();
```
adamgall, line 151: We have a hook called something like `useNetworkWalletClient()` that should be used instead of `useWalletClient()`.

src/components/ui/modals/GaslessVoting/RefillGasTankModal.tsx
```diff
@@ -19,175 +18,196 @@ interface RefillGasFormValues {
...
+                    <BigIntInput
+                      {...field}
+                      value={field.value?.bigintValue}
+                      onChange={value => {
+                        setFieldValue('inputAmount', value);
+                      }}
+                      parentFormikValue={values.inputAmount}
+                      decimalPlaces={balance?.decimals || 0}
+                      placeholder="0"
+                      maxValue={!isDirectDeposit ? balance?.value || 0n : undefined}
```
adamgall, line 96: what does an `undefined` max value mean here?

adamgall (changes requested):

src/components/GaslessVoting/GaslessVotingToggle.tsx
DarksightKellar: Ah yes good catch, thanks. Updated

src/components/ui/modals/GaslessVoting/RefillGasTankModal.tsx
DarksightKellar: Removes the "Max" trailing button. Since the user is paying for gas, doesn't make sense to allow to input their max balance

adamgall: what happens if a user enters a value which is greater than their wallet's balance?

DarksightKellar: Validation fails and the submit button is disabled

adamgall: awesome thank you

adamgall (approved):

PR #2806, opened 2025-03-26
Update package dependencies

adamgall:
- remove unused "graphql" and "@types/lodash.isequal"
- add "@walletconnect/types", which is used in `src/components/SafeInjectIframe/hooks/useWalletConnect.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "evm-proxy-detection": "^1.1.0",
     "formik": "^2.2.9",
     "framer-motion": "^6.5.1",
-    "graphql": "^16.10.0",
     "hono": "^4.6.16",
     "i18next": "^23.10.1",
     "i18next-browser-languagedetector": "^6.1.5",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -110,7 +109,6 @@
     "jsdom": "^24.0.0",
     "prettier": "^3.3.3",
     "typescript": "^5.6.3",
-    "vitest": "^1.2.2",
-    "wrangler": "^3.102.0"
```
Da-Colon, line 114: is this cloudflare related? that I assume was never used?

adamgall: Yeah `wrangler` is a Cloudflare package, which is used for interacting with Cloudlfare services from the cli. It's not needed as a dependency in this project though.

Da-Colon (approved):

DarksightKellar (approved):

twodam (approved):

PR #2817, opened 2025-03-28
`[ENG-425-plus]` reject transaction flow&UX for multsig DAO

Da-Colon:
Merge with: https://github.com/decentdao/decent-app/pull/2816

## Reviewing:
Best way to review is follow by each commit:

- Removes the rejection proposals from Dashboard and removes related components and types
- improves logic around `SignerDetails` logic. As discussed in Slack, the most recent confirmation chosen will be what decides what `Badge` is shown
- updates return type
- If no rejection proposal exists, create a proposal to create a new one.
- clean up signing logic, and if rejection confirmations are passed threshold, show rejection button.

## Screenshots
- Proposal Created
localhost_3000_proposals_0x81664e2d86bbb20512d22f4a2db86b138cb6a882a1df376667ed9f814c788d0f_dao=sep_0x40584Ef308e7317BcaF086dB60cCA926c08d81c1(Desktop) (1)
- Rejection successful, buttons disabled
localhost_3000_proposals_0x81664e2d86bbb20512d22f4a2db86b138cb6a882a1df376667ed9f814c788d0f_dao=sep_0x40584Ef308e7317BcaF086dB60cCA926c08d81c1(Desktop) (2)
- Other member visits
localhost_3000_proposals_0x81664e2d86bbb20512d22f4a2db86b138cb6a882a1df376667ed9f814c788d0f_dao=sep_0x40584Ef308e7317BcaF086dB60cCA926c08d81c1(Desktop) (3)
- Other member rejects and execution shows
localhost_3000_proposals_0x0a2efa22bca3c29cb2f1294a23c124a05ebfd7d4c19f2ccb63a77894552ec55a_dao=sep_0x40584Ef308e7317BcaF086dB60cCA926c08d81c1(Desktop)

## Notes:
- Made 2 temp UI executive decisions. Using the reds and variant available. I updated the 'Reject' button to be the 'danger' variant, which for whatever reason did not have disabled properties. (See Screenshot 2). @decentdao/design can beat me up later and we can create a ticket to update. (I'll commit a comment w/ issue after review)

- Last open question is in the final screenshot. There are some left over edges dealing with execution of proposal/rejection.


src/components/Proposals/MultisigProposalDetails/TxActions.tsx
```diff
@@ -302,6 +379,28 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
     proposal.state === FractalProposalState.TIMELOCKED ||
     !isActiveNonce;
 
+  if (isRejectedProposalPassThreshold) {
```
Da-Colon, line 382: @decentdao/engineering @cbajandas So the open question is when to show the Rejection Execution and When to show Proposal Execution.

As it currently stands: Race to finish

- When a proposal normal threshold is reached all other buttons are removed and no longer can 'reject'. 
- When a rejection threshold is reached all other buttons are removed and no longer can 'approve'.

This opens the floor. 

- Should we show 'reject' option even when approve threshold is reached?
- Should we show 'approve' option even when reject is reached?
- If yes to either of those? Should we show both Execution buttons when both thresholds are reached?

- Or we cool with leaving as is?

Da-Colon: This may be a work done after this PR I think.

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
```diff
@@ -245,11 +261,72 @@ export function TxActions({ proposal }: { proposal: MultisigProposal }) {
...
+            safeTx.data,
+            safeTx.operation,
+            BigInt(safeTx.safeTxGas),
+            BigInt(safeTx.baseGas),
+            BigInt(safeTx.gasPrice),
+            safeTx.gasToken,
+            safeTx.refundReceiver,
+            signatures,
+          ]),
+        failedMessage: t('failedExecute', { ns: 'transaction' }),
```
twodam, line 318: this new `executeRejectedProposal` function is almost identical with `executeTransaction` except it has different messages. can we modify `executeTransaction` to accept both `proposal` and `messages` parameters, then re-use this function?

### Refactor? not now, just taking a note.
I think this component has too many function lines, we could also extract these functions to be put a file under `helpers`, so we have a short and clear visual component file. this can be done later in future.
and also noted that these functions has been using variables in addition to parameters, it can mess the logic when someone trying to understand too.

src/hooks/DAO/proposal/useProposals.ts
```diff
@@ -23,16 +23,19 @@ export function useProposalsSortedAndFiltered({
...
-        const dataB = new Date(b.eventDate).getTime();
-        if (sortBy === SortBy.Oldest) {
-          return dataA - dataB;
-        }
-        return dataB - dataA;
-      });
+    return (
+      [...(proposals || [])]
+        // filters out "rejection" multisig proposals
+        .filter(proposal => filters.includes(proposal.state!) && proposal.transaction?.data)
```
twodam, line 29: is it better to add full condition for detecting rejection proposal? like what we do in `isMultisigRejectionProposal` function.
transaction without data can be an native ETH transfer, or do we want to assume that transfer will all be wrapped in `multisend` as long as user are creating proposal on Decent interface?

what happens if user queue a no-data(ETH transfer?) transaction on Safe interface and go to Decent? will that transaction be missing in our interface?

src/components/Activity/ActivityDescriptionGovernance.tsx
```diff
@@ -23,21 +22,6 @@ const formatId = (proposalId: string) => {
   }
 };
 
-function OnChainRejectionMessage({ activity }: { activity: FractalProposal }) {
```
twodam, line 26: which part of interface used this? is this an indicator above proposal title to tell user it's a rejection or?

twodam (changes requested):
are we sure to hide all empty-data transaction proposal from the list? are they all must be rejection proposal? (detail in above review comment)

src/components/Activity/ActivityDescriptionGovernance.tsx
Da-Colon: This was for rendering the Text for the Card on the DaoDashboard like: Proposal to Reject: ProposalId. Which is no longer needed.

src/hooks/DAO/proposal/useProposals.ts
Da-Colon: ah fog, forgot about native-transfers. yeah can add a check for a value that isn't 0. I guess I can readd something to the interface and do the check back in useSafeTransactions?

src/components/Proposals/MultisigProposalDetails/TxActions.tsx
Da-Colon: Agreed. Normally, considering hotfix and it would require more updates to the `ButtonProps` I created a seperate function and called it a day.

src/hooks/DAO/proposal/useProposals.ts
Da-Colon: and then, went a little further and clean up the typing and reuse the same function `isMultisigRejectionProposal` across both instances

PR #2831, opened 2025-04-02
`Eng-435` | Use v1 voting and paymaster abis from fractal-contracts

DarksightKellar:
Closes Eng-435


src/models/AzoriusTxBuilder.ts
adamgall, line 1: Github is saying the `AbiFunction` import is defined but never used. Delete it?

DarksightKellar: I missed this when cherry picking lines to include in this PR. It's removed in the next

adamgall: Can you please get it into this PR?

adamgall (approved):

PR #2832, opened 2025-04-02
`ENG-406` estimate UserOperation gas to replace fixed value

twodam:


DarksightKellar (approved):
LGTM to merge as is. I've got a PR in the works that explores the multipliers and attempts to DRY up some repeat code

PR #2835, opened 2025-04-02
Add cast vote fallback pending hiFi UX design

DarksightKellar:
Adds an automatic call to `castVote` on `castGaslessVote` error, after a 5-second delay.
Closes ENG-492


src/i18n/locales/en/proposal.json
```diff
@@ -191,5 +191,6 @@
   "streamTransferableHelper": "If enabled, the stream can be transferred to another address by the stream's recipient",
   "proposalSummaryNotEnoughVotingPower": "Your voting power did not meet requirements to vote on this proposal when it was created.",
   "proposalSummaryNotEnoughVotingPowerSecondary": "Delegate voting tokens to your address to be eligible to vote on future proposals.",
-  "castVoteError": "Failed to send vote"
+  "castVoteError": "Failed to send vote",
+  "castVoteErrorTempAutoFallback": ". Falling back to paid voting..."
```
adamgall, line 195: nitpick: why not just add the period to `castVoteError`

DarksightKellar: Because I'd like a clean break when it's time for that translation key to be removed.

PR #2836, opened 2025-04-02
Expand types and typescript errors

jigglyjams:
This tsconfig flag fully expands types and typescript errors in IDE

before:
image

after (you get a scroll bar):
image

credit:
https://stackoverflow.com/questions/62059942/how-to-extract-a-large-type-with-vs-code-without-the-19-more-truncate


adamgall (approved):
Thank you this is legendary

PR #2862, opened 2025-04-08
If gasless feature not enabled on dao create, deploy old strategies

DarksightKellar:
Closes [ENG-544](https://linear.app/decent-labs/issue/ENG-544/if-gasless-voting-is-disabled-and-user-deploys-a-new-dao-deploy-the)


src/models/DaoTxBuilder.ts
```diff
@@ -141,7 +141,7 @@ export class DaoTxBuilder extends BaseTxBuilder {
       txs.push(azoriusTxBuilder.buildCreateTokenTx());
     }
 
-    txs.push(azoriusTxBuilder.buildDeployStrategyTx());
+    txs.push(azoriusTxBuilder.buildDeployStrategyTx(enableGaslessVoting));
```
DarksightKellar, line 144: I _briefly_ considered using the env feature flag for this boolean instead of the toggle's value, but, eh.

- This is all temporary anyway
- The v1 strategies will be deployed if the DAO chooses to enable gasless voting in the future
- Functionally from the user perspective, nothing really changes.

PR #2886, opened 2025-04-14
`[ENG-213]` fix: `useAzroiusListeners`unmounts when event fires

Da-Colon:
Closes [ENG-213](https://linear.app/decent-labs/issue/ENG-213/after-a-proposal-is-created-the-proposal-list-is-not-updated-until-a)

Just doubles the time we wait before we lookup call the `getProposalsVotes` which is causing the failure when ERC20_Votes hasn't been mined yet

### DIFF
--- a/src/hooks/DAO/loaders/governance/useAzoriusListeners.ts
+++ b/src/hooks/DAO/loaders/governance/useAzoriusListeners.ts
@@ -115,7 +115,7 @@ export const useAzoriusListeners = () => {
           // which include the `proposalId` error out because the RPC node (rather, the block it's on)
           // doesn't see this proposal yet (despite the event being caught in the app...).
           const averageBlockTime = await getAverageBlockTime(publicClient);
-          await new Promise(resolve => setTimeout(resolve, averageBlockTime * 1000));
+          await new Promise(resolve => setTimeout(resolve, averageBlockTime * 1000 * 2));
 
           const typedTransactions = log.args.transactions.map(t => ({
             ...t,


adamgall:
How reproducible is the issue that this PR fixes?

Da-Colon:
> How reproducible is the issue that this PR fixes?

For me, I've had trouble with this quite often, where it shows the error where `getProposalVotes` hasn't been mined and the proposal never loads in. Now granted double might be overkill. but maybe 1.5x longer might catch the majority of these times of failures

adamgall:
> > How reproducible is the issue that this PR fixes?
> 
> For me, I've had trouble with this quite often, where it shows the error where `getProposalVotes` hasn't been mined and the proposal never loads in. Now granted double might be overkill. but maybe 1.5x longer might catch the majority of these times of failures

Sure, I've seen it as well.

Just wondering, how reproducible is that issue? Can you get it to happen consistently by editing that timeout down to 0 seconds, for example?

I just want to make sure that increasing that timeout is the thing that will actually lower the probability of the issue happening.

Da-Colon:
> > > How reproducible is the issue that this PR fixes?
> > 
> > 
> > For me, I've had trouble with this quite often, where it shows the error where `getProposalVotes` hasn't been mined and the proposal never loads in. Now granted double might be overkill. but maybe 1.5x longer might catch the majority of these times of failures
> 
> Sure, I've seen it as well.
> 
> Just wondering, how reproducible is that issue? Can you get it to happen consistently by editing that timeout down to 0 seconds, for example?
> 
> I just want to make sure that increasing that timeout is the thing that will actually lower the probability of the issue happening.

Ah I see what are you saying, For sure I'll do some more testing

Da-Colon:
I've tested: 0s, averageBlockTime, averageBlockTime + 5s, averageBlockTime + 10s, averageBlockTime x 1.5, averageBlockTime x 2. tried 5 different proposals each timing and waited up to 2 minutes to see if anything happens. refreshing before creating the next proposal

The results of this points to two different issues.

## The issue I went after first
`ERC20_Votes haven't been mined yet`:
First the issue that was attempt to be tacked by waiting additional average block time before attempting to call the strategy contract for the votes. Up to `ABT + 5s`, it was a pretty consistent with only 1 loading, 1 failing with an error, rest failing without showing error.

`ABT + 10s` and `ABT x 1.5` both resulted in 2 loading and rest failing with no error seen, while `ABT x 2` 3 loading and 2 resulting in failing without the error. (with a 6 run resulting in a success)

## The behind the scenes issue
These `failing without showing error` was curious. I placed a console log within the `onLogs` handler and found that when it `fails without error` the console also never shows up. Not really sure what's going on here. 

While increasing the setTimeout to be longer does seem increate the change of success,  it looks like somewhere between 1.25 and 1.5 is probably all we need since I wasn't seeing the error in the `ABT + 10s` trails.

As far as the other issue. Kinda just marking my current findings as I continue dig into this further.

Da-Colon:
I'm going to close this PR for now. Ticket is still active and will revisit when @mudrila zustand work is further along

PR #2887, opened 2025-04-14
`[ENG-309]` fix: Throw error causing proposals not to load

Da-Colon:
Closes [ENG-309](https://linear.app/decent-labs/issue/ENG-309/proposal-overview-transaction-details-dont-load-for-dao-targeted#comment-9ef39810)

### DIFF
--- a/src/utils/azorius.ts
+++ b/src/utils/azorius.ts
@@ -251,7 +251,7 @@ export const mapProposalCreatedEventToProposal = async (
       transactionHash = executedEvent?.transactionHash;
     } else {
       // @dev Proposal with 0 transactions goes straight into EXECUTED state, but since executeProposal event wasn't fired - it can't be found
-      throw new Error('Proposal state is EXECUTED, but no execution event found');
+      transactionHash = createdEventTransactionHash;
     }
   } else {
     transactionHash = createdEventTransactionHash;


adamgall:
Not seeing any difference in this PR, compared with the issue that was originally reported by che: https://linear.app/decent-labs/issue/ENG-309/proposal-overview-transaction-details-dont-load-for-dao-targeted#comment-9ef39810

Screenshot 2025-04-14 at 2 04 48 PM

Da-Colon:
Now granted this is a some rare state to be in. But technically someone could do this. 

image

So I got here while I was working on the proposal builder. removing the transaction builder from the workflow and I was able to create a proposal without with other data (just filled out the metadata) and ended up with a proposal like above. I have since patched this in the first PR I believe. 

But looking into error, I don't believe we should throw here as it causes the rest of the proposals to fail to load

adamgall:
Are you confirming that this PR does not fix ENG-309?

Da-Colon:
> Are you confirming that this PR does not fix ENG-309?

You are correct, I was looking into that when I discovered this. my bad meant to open a new ticket.

Da-Colon:
updated.

adamgall (approved):

PR #2896, opened 2025-04-15
`[ENG-375]` add asset drop down menu in place of streamed token address

twodam:
### Summary

* Use asset dropdown menu instead of contract address input
* Display a toast when there's no available asset to stream
* Assets filter by `!asset.possibleSpam && !asset.nativeToken && parseFloat(asset.balance) > 0`.

#### Asset dropdown
image

#### No available asset
image

### DIFF
--- a/src/components/ProposalBuilder/ProposalStream.tsx
+++ b/src/components/ProposalBuilder/ProposalStream.tsx
@@ -55,7 +55,7 @@ export function ProposalStream({
 
   const safeAddress = safe?.address;
   const fungibleNonNativeAssetsWithBalance = assetsFungible.filter(
-    asset => !asset.nativeToken && parseFloat(asset.balance) > 0,
+    asset => !asset.possibleSpam && !asset.nativeToken && parseFloat(asset.balance) > 0,
   );
   const selectedAssetIndex = fungibleNonNativeAssetsWithBalance.findIndex(
     asset => asset.tokenAddress === stream.tokenAddress,
--- a/src/components/ProposalBuilder/ProposalStream.tsx
+++ b/src/components/ProposalBuilder/ProposalStream.tsx
@@ -9,14 +9,18 @@ import {
   Flex,
   HStack,
   IconButton,
+  Select,
   Text,
   VStack,
 } from '@chakra-ui/react';
 import { CaretDown, CaretRight, MinusCircle, Plus } from '@phosphor-icons/react';
+import { Field, FieldAttributes, FieldProps } from 'formik';
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { erc20Abi, formatUnits, getContract, isAddress } from 'viem';
+import { useCurrentDAOKey } from '../../hooks/DAO/useCurrentDAOKey';
 import useNetworkPublicClient from '../../hooks/useNetworkPublicClient';
+import { useStore } from '../../providers/App/AppProvider';
 import { useDaoInfoStore } from '../../store/daoInfo/useDaoInfoStore';
 import { Stream } from '../../types/proposalBuilder';
 import { scrollToBottom } from '../../utils/ui';
@@ -43,9 +47,19 @@ export function ProposalStream({
   const [tokenBalanceFormatted, setTokenBalanceFormatted] = useState('');
   const [expandedIndecies, setExpandedIndecies] = useState<number[]>([0]);
   const { safe } = useDaoInfoStore();
+  const { daoKey } = useCurrentDAOKey();
+  const {
+    treasury: { assetsFungible },
+  } = useStore({ daoKey });
   const { t } = useTranslation(['proposal', 'common']);
 
   const safeAddress = safe?.address;
+  const fungibleNonNativeAssetsWithBalance = assetsFungible.filter(
+    asset => !asset.nativeToken && parseFloat(asset.balance) > 0,
+  );
+  const selectedAssetIndex = fungibleNonNativeAssetsWithBalance.findIndex(
+    asset => asset.tokenAddress === stream.tokenAddress,
+  );
 
   useEffect(() => {
     const fetchFormattedTokenBalance = async () => {
@@ -72,6 +86,7 @@ export function ProposalStream({
 
     fetchFormattedTokenBalance();
   }, [safeAddress, publicClient, stream.tokenAddress]);
+
   return (
     <AccordionPanel p={0}>
       <VStack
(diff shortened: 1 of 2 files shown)


src/components/ProposalBuilder/ProposalStream.tsx
```diff
@@ -72,30 +86,63 @@ export function ProposalStream({
...
+                iconSize="1.5rem"
+                icon={<CaretDown />}
+                data-testid="stream.tokenAddress"
+                onChange={e => {
+                  console.debug(
+                    'Asdsad',
+                    index,
+                    Number(e.target.value),
+                    fungibleNonNativeAssetsWithBalance[Number(e.target.value)],
+                  );
```
adamgall, line 126: remove this please, generally as a rule in this codebase we don't commit `console` logs

src/components/ProposalBuilder/ProposalStream.tsx
```diff
@@ -72,30 +86,63 @@ export function ProposalStream({
...
+                  });
+                }}
+                value={selectedAssetIndex}
+              >
+                {fungibleNonNativeAssetsWithBalance.map((asset, i) => (
+                  <option
+                    key={i}
+                    value={i}
+                  >
+                    {asset.symbol}
```
adamgall, line 139: Would it be good from a product perspective to also show the contract address here?

example: `WBTC (0x2260fac5e5542a773aa44fbcfedf7c193bc2c599)`

or maybe even a shortened version of it:

example: `WBTC (0x2260...c599)` (see `createAccountSubstring()` function in codebase for example of how to shorten)

src/components/ProposalBuilder/ProposalStream.tsx
```diff
@@ -43,9 +47,19 @@ export function ProposalStream({
...
   const { safe } = useDaoInfoStore();
+  const { daoKey } = useCurrentDAOKey();
+  const {
+    treasury: { assetsFungible },
+  } = useStore({ daoKey });
   const { t } = useTranslation(['proposal', 'common']);
 
   const safeAddress = safe?.address;
+  const fungibleNonNativeAssetsWithBalance = assetsFungible.filter(
+    asset => !asset.possibleSpam && !asset.nativeToken && parseFloat(asset.balance) > 0,
```
adamgall, line 58: Does `possibleSpam` filter out a DAO's native token? If so, that would be bad.

Maybe we should "hardcode" to always allow the DAO's native governance token to pass this filter.

src/pages/dao/proposal-templates/SafeProposalTemplatesPage.tsx
```diff
@@ -46,6 +48,10 @@ export function SafeProposalTemplatesPage() {
 
   const safeAddress = safe?.address;
   const { openSendAssetsModal } = useSendAssetsActionModal();
+  const hasAvailableAssetsForSablierStream =
+    assetsFungible.filter(
+      asset => !asset.possibleSpam && !asset.nativeToken && parseFloat(asset.balance) > 0,
```
adamgall, line 53: Same `possibleSpam` comment as above

adamgall (changes requested):

src/components/ProposalBuilder/ProposalStream.tsx
twodam: I think it's like a blacklist where people could report scam for certain token contract https://docs.moralis.com/2.0/web3-data-api/evm/erc20-spam-detection
yeah, I can make sure native governance token pass this check.

I used this because I saw there's many spam token in ShutterDAO which is pretty obvious, they all have website url in the token symbol.
image

src/components/ProposalBuilder/ProposalStream.tsx
twodam: oh, I must have forgot about this

src/components/ProposalBuilder/ProposalStream.tsx
twodam: display full address link below the input?
image

src/components/ProposalBuilder/ProposalStream.tsx
twodam: Safe interface are using whitelist: https://help.safe.global/en/articles/40784-default-token-list-local-hiding-of-spam-tokens

twodam: updated.

src/hooks/utils/useTokenListFilter.ts
```diff
@@ -0,0 +1,50 @@
...
+
+interface TokenListFilterOptions {
+  includeNativeToken?: boolean;
+  includeZeroBalanceToken?: boolean;
+}
+
+/**
+ * Filter out spam tokens. DAO's governance tokens will be included.
+ */
+export function useTokenListFilter(options: Partial<TokenListFilterOptions> = {}) {
```
DarksightKellar, line 15: Why `Partial` here? All `TokenListFilterOptions` props are optional anyway

src/pages/dao/proposals/new/sablier/SafeSablierProposalCreatePage.tsx
```diff
@@ -33,11 +35,13 @@ export function SafeSablierProposalCreatePage() {
   const { daoKey } = useCurrentDAOKey();
   const {
     governance: { type },
+    treasury: { assetsFungible },
   } = useStore({ daoKey });
   const {
     addressPrefix,
     contracts: { sablierV2Batch, sablierV2LockupTranched },
   } = useNetworkConfigStore();
+  const tokenListFilter = useTokenListFilter();
```
DarksightKellar, line 44: Forgive this _nitpick_, but, I prefer verbs for functions like this. So `filterTokenList`, which reads more like a "thing do-er".

Can take it further and be more descriptive, something like: `filterSpamTokens`

DarksightKellar (approved):

DarksightKellar (changes requested):
Minor changes requested, then good to go

src/pages/dao/proposals/new/sablier/SafeSablierProposalCreatePage.tsx
twodam: yeah, make sense.

src/hooks/utils/useTokenListFilter.ts
twodam: updated.

adamgall (approved):
Thanks for addressing my comments ser

PR #2900, opened 2025-04-15
Remove `maxFeePerGas`; get `maxPriorityFeePerGas` directly from Alchemy

DarksightKellar:
- `maxPriorityGasFee` is now fetched directly from Alchemy's API. Using that without any modifiers
- Removed `maxFeePerGas`. According to [this doc](https://docs.alchemy.com/docs/maxpriorityfeepergas-vs-maxfeepergas#when-to-use-maxpriorityfeepergas-vs-maxfeepergas-a-hrefwhen-to-use-max-priority-fee-per-gas-vs-max-fee-per-gas-idwhen-to-use-max-priority-fee-per-gas-vs-max-fee-per-gasa:~:text=maxFeePerGas%20%3D%20baseFeePerGas%20%2B%20maxPriorityFeePerGas-,When%20to%20use%20maxPriorityFeePerGas%20vs.%20maxFeePerGas%3F,-The%20most%20guaranteed), doing this lets Alchemy tune that based on recent blocks. This works out better that playing whack-a-mole with stupid multipliers.

### DIFF
--- a/src/hooks/DAO/proposal/useCastVote.ts
+++ b/src/hooks/DAO/proposal/useCastVote.ts
@@ -10,6 +10,7 @@ import { EntryPoint07Abi } from '../../../assets/abi/EntryPoint07Abi';
 import { useStore } from '../../../providers/App/AppProvider';
 import { useNetworkConfigStore } from '../../../providers/NetworkConfig/useNetworkConfigStore';
 import { useDaoInfoStore } from '../../../store/daoInfo/useDaoInfoStore';
+import { fetchMaxPriorityFeePerGas } from '../../../utils/gaslessVoting';
 import useNetworkPublicClient from '../../useNetworkPublicClient';
 import { useNetworkWalletClient } from '../../useNetworkWalletClient';
 import { useTransaction } from '../../utils/useTransaction';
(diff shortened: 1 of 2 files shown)


src/hooks/DAO/proposal/useCastVote.ts
```diff
@@ -198,12 +184,22 @@ const useCastVote = (proposalId: string, strategy: Address) => {
       transport: http(rpcEndpoint),
     });
 
+    const minimumMaxPriorityFeePerGas = await fetchMaxPriorityFeePerGas(networkConfig);
+    const { maxPriorityFeePerGas: maxPriorityFeePerGasEstimate } =
+      await publicClient.estimateFeesPerGas();
+    const maxPriorityFeePerGasEstimateBuffered = maxPriorityFeePerGasEstimate * (13n / 10n);
```
DarksightKellar, line 190: Honestly not sure how useful this buffer is. Is it truly necessary at this point? 

Or maybe we apply the buffer directly to the thing (`maxPriorityFeePerGas`) we're passing into the UserOP?

DarksightKellar: So I imagine my second point holds true then? Apply the buffer to `maxPriorityFeePerGas`, instead of to the values we're choosing from (`maxPriorityFeePerGasEstimate` VS `minimumMaxPriorityFeePerGas`).

So, we pick the larger of the 2 and add a buffer to the one picked.

Basically I don't see the point in buffering `maxPriorityFeePerGasEstimate` if there's a chance it won't be used anyway. Or that the buffer makes it _slightly_ more than `minimumMaxPriorityFeePerGas`, but then when it's time for it to be mined, network conditions have worsened and it's too small. In which case it would have been better to pick `minimumMaxPriorityFeePerGas` and buffer that one instead. No?

DarksightKellar: I've gone ahead and pushed a commit with this change. Lemme know what you think

DarksightKellar: Spoke with @johnqh. Decided buffer is not needed in this case of priority fees.

src/utils/gaslessVoting.ts
```diff
@@ -103,3 +105,31 @@ export const getVoteSelectorAndValidator = (
...
+    }
+
+    const data = await response.json();
+    return BigInt(data.result as Hex);
+  } catch (error) {
+    console.error('Error fetching maxPriorityFeePerGas from Alchemy:', {
+      chain: networkConfig.chain.name,
+      error: error instanceof Error ? error.message : 'Unknown error',
+    });
+    throw error;
```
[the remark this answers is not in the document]
DarksightKellar: Done, see latest commit

src/hooks/DAO/proposal/useCastVote.ts
```diff
@@ -198,12 +181,22 @@ const useCastVote = (proposalId: string, strategy: Address) => {
       transport: http(rpcEndpoint),
     });
 
+    const minimumMaxPriorityFeePerGas = await fetchMaxPriorityFeePerGas(networkConfig);
```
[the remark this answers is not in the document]
DarksightKellar: See update in latest commit

PR #2906, opened 2025-04-17
`[ENG-680]` Remove Actions bugfix 

Da-Colon:
## Description
I realized that the way I originally set this up was causing more problems than it solved and moving the `ActionsExperience` actually had fixed the dupe issue we had saw with the `reEnableInitialization` on the Formik form. After some trial and error with trying a more direct approach, and some testing I think its safe from dupes and we don't need to manually groom the transactions. 

@johnqh This needs to make it in next release with the other work to ensure no this feature goes out working correctly and allowing users to remove created transactions without refresh.


src/store/actions/useProposalActionsStore.ts
```diff
@@ -18,7 +18,18 @@ const initialProposalActionsStore: ProposalActionsStoreData = {
...
+      const newAction = {
+        ...action,
+        transactions: action.transactions.map(transaction => ({
+          ...transaction,
+          // Generate a unique actionId for each transaction based on the action type and index
+          actionId: `${action.actionType}_${state.actions.length}`,
+        })),
+      };
+      return { actions: [...state.actions, newAction] };
+    }),
```
Da-Colon, line 32: This handles when actions are added outside of the Proposal Builder form

Da-Colon:
need to rethink some things.

PR #2910, opened 2025-04-18
chore: update @fractal-framework/fractal-contracts to version 1.6.0

adamgall:
and adjust gasless voting minimum stake values in network configurations

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@chakra-ui/react": "^2.8.2",
     "@cloudflare/workers-types": "^4.20241230.0",
     "@fontsource/space-mono": "^5.0.19",
-    "@fractal-framework/fractal-contracts": "^1.6.0-rc.5",
+    "@fractal-framework/fractal-contracts": "^1.6.0-rc.7",
     "@hatsprotocol/modules-sdk": "^1.4.0",
     "@hatsprotocol/sdk-v1-core": "^0.9.0",
     "@hatsprotocol/sdk-v1-subgraph": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "@chakra-ui/react": "^2.8.2",
     "@cloudflare/workers-types": "^4.20241230.0",
     "@fontsource/space-mono": "^5.0.19",
-    "@fractal-framework/fractal-contracts": "^1.6.0-rc.7",
+    "@fractal-framework/fractal-contracts": "^1.6.0",
     "@hatsprotocol/modules-sdk": "^1.4.0",
     "@hatsprotocol/sdk-v1-core": "^0.9.0",
     "@hatsprotocol/sdk-v1-subgraph": "^1.0.0",
--- a/src/hooks/DAO/proposal/useCastVote.ts
+++ b/src/hooks/DAO/proposal/useCastVote.ts
@@ -269,18 +269,9 @@ const useCastVote = (proposalId: string, strategy: Address) => {
     };
 
     estimateGaslessVoteGas().catch(() => {
-      //const error = e as EstimateUserOperationGasErrorType;
-      //console.warn('error', error.message);
       setCanCastGaslessVote(false);
     });
-  }, [
-    accountAbstraction,
-    paymasterAddress,
-    prepareCastVoteData,
-    prepareGaslessVoteOperation,
-    publicClient,
-    rpcEndpoint,
-  ]);
+  }, [accountAbstraction, paymasterAddress, prepareGaslessVoteOperation, publicClient]);
 
   const castGaslessVote = useCallback(
     async ({
--- a/src/providers/NetworkConfig/networks/mainnet.ts
+++ b/src/providers/NetworkConfig/networks/mainnet.ts
@@ -140,7 +140,7 @@ export const mainnetConfig: NetworkConfig = {
   ],
   gaslessVoting: {
     maxPriorityFeePerGasMultiplier: 100n, // @todo: Might be many times higher (or lower), depending on returned value from `estimateFeesPerGas`
-    bundlerMinimumStake: 10_0000_0000_0000_0000n, // 0.1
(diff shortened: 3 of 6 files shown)


src/providers/NetworkConfig/networks/polygon.ts
adamgall, line 1: No gasless voting or min staking on polygon yet

src/providers/NetworkConfig/networks/mainnet.ts
```diff
@@ -140,7 +140,7 @@ export const mainnetConfig: NetworkConfig = {
   ],
   gaslessVoting: {
     maxPriorityFeePerGasMultiplier: 100n, // @todo: Might be many times higher (or lower), depending on returned value from `estimateFeesPerGas`
-    bundlerMinimumStake: 10_0000_0000_0000_0000n, // 0.1
+    bundlerMinimumStake: 1_000_000_000_000_000_000n, // 1 ETH
```
adamgall, line 143: Want 1 ETH stake on Mainnet.

src/hooks/DAO/proposal/useCastVote.ts
```diff
@@ -269,18 +269,9 @@ const useCastVote = (proposalId: string, strategy: Address) => {
...
     });
-  }, [
-    accountAbstraction,
-    paymasterAddress,
-    prepareCastVoteData,
-    prepareGaslessVoteOperation,
-    publicClient,
-    rpcEndpoint,
-  ]);
+  }, [accountAbstraction, paymasterAddress, prepareGaslessVoteOperation, publicClient]);
```
adamgall, line 274: Had some unneeded items in this dep list

package.json
```diff
@@ -8,7 +8,7 @@
     "@chakra-ui/react": "^2.8.2",
     "@cloudflare/workers-types": "^4.20241230.0",
     "@fontsource/space-mono": "^5.0.19",
-    "@fractal-framework/fractal-contracts": "^1.6.0-rc.5",
+    "@fractal-framework/fractal-contracts": "^1.6.0-rc.7",
```
adamgall, line 11: THESE ARE TEH CONTRACTS. LESGOOOOO

src/providers/NetworkConfig/networks/mainnet.ts
twodam: so it did require that much on single paymaster to work on mainnet?

src/providers/NetworkConfig/networks/sepolia.ts
```diff
@@ -134,7 +134,7 @@ export const sepoliaConfig: NetworkConfig = {
   ],
   gaslessVoting: {
     maxPriorityFeePerGasMultiplier: 100n, // @todo: Might be many times higher (or lower), depending on returned value from `estimateFeesPerGas`
-    bundlerMinimumStake: 10_0000_0000_0000_0000n, // 0.1
+    bundlerMinimumStake: 10_000_000_000_000_000n, // stake 0.1 ETH
```
twodam, line 137: this will be 0.01 ETH

twodam (changes requested):

src/providers/NetworkConfig/networks/sepolia.ts
adamgall: Yes good call out, I will fix in the morning

src/providers/NetworkConfig/networks/mainnet.ts
adamgall: Truthfully, I have only tested with 1 ETH stake on mainnet.

For some reason I was under the impression the rundler required a 1 ETH stake, but...

These docs here do say that the rundler only requires 0.1 ETH: https://docs.alchemy.com/docs/bundler-services

Is that why you set it to 0.1 ETH initially?

twodam: yeah, there's config said `1 eth` and doc said `0.1 eth`. do you think maybe we can **test once with 0.1 ETH** stake? it may save user some funds (0.9 ETH) from being kept in staking.

src/providers/NetworkConfig/networks/sepolia.ts
adamgall: You're right we don't _need_ to, but @johnqh and I decided to, for "accurate testing" purposes, leave this config set to a non zero number to better represent the UX flows that will happen on mainnet.

Although for the sake of preserving* our Sepolia ETH I would like to reduce this to 0.01 or 0.001 ETH.

*it's true that this ETH is never lost, but it's just a bit annoying to get it back.

src/providers/NetworkConfig/networks/mainnet.ts
adamgall: I agree with @twodam and will test with 0.1 ETH today

package.json
adamgall: actually i just released `1.6.0`, so those are the actual contracts lesgo

src/providers/NetworkConfig/networks/sepolia.ts
DarksightKellar: Ah yeah makes sense. Also agree with reducing stake required. `0.001` sounds good

package.json
```diff
@@ -8,7 +8,7 @@
     "@chakra-ui/react": "^2.8.2",
     "@cloudflare/workers-types": "^4.20241230.0",
     "@fontsource/space-mono": "^5.0.19",
-    "@fractal-framework/fractal-contracts": "^1.6.0-rc.5",
+    "@fractal-framework/fractal-contracts": "^1.6.0",
```
DarksightKellar, line 11: ✨🔥🔥

DarksightKellar (approved):

src/providers/NetworkConfig/networks/mainnet.ts
adamgall: Just tested a stake of 0.1 ETH on mainnet, and would you look at that, it works.

Excellent suggestion @twodam!

cc: @johnqh @DarksightKellar

PR #2912, opened 2025-04-18
Refactor buildContractCall for ABI type safety

DarksightKellar:
Had this sitting on my machine, totally forgot about it. Was doing some pruning and only accidentally did not delete the branch along with old dead branches.

Anyway, context:
While doing gasless voting work aeons ago, I noticed that I could pass in garbage function names and args into `buildContractCall`, completely breaking the usefulness of the type-safe ABIs we're using.

This PR addresses that.


src/helpers/crypto.ts
```diff
@@ -167,45 +165,31 @@ export const buildSafeAPIPost = async (
...
   return buildSafeTransaction(
     Object.assign(
       {
-        to: contractAddress,
-        data,
+        to: target,
+        data: encodedFunctionData,
         operation,
-        nonce,
+        nonce: 0,
```
DarksightKellar, line 186: Let's leave this to future us to add a `nonce` param if we ever need it.

src/helpers/crypto.ts
```diff
@@ -167,45 +165,31 @@ export const buildSafeAPIPost = async (
...
   );
 };
 
-export const buildContractCall = (
-  contractAbi: Abi,
-  contractAddress: Address,
-  method: string,
-  params: any[],
-  nonce: number,
-  delegateCall?: boolean,
```
DarksightKellar, line 197: Removed `nonce` and `delegateCall` as they are never actually being used beyond their defaults of `0`.

src/models/AzoriusTxBuilder.ts
```diff
@@ -195,77 +195,108 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
...
-        abis.LinearERC20Voting,
-        this.linearERC20VotingAddress!,
-        'setAzorius', // contract function name
-        [this.predictedAzoriusAddress],
-        0,
-        false,
-      );
+      if (!this.predictedAzoriusAddress || !this.linearERC20VotingAddress) {
+        throw new Error('Linear ERC20 addresses not set');
+      }
```
adamgall, line 200: Can you split this into two separate conditions, for error message accuracy?

src/models/AzoriusTxBuilder.ts
```diff
@@ -195,77 +195,108 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
...
-        abis.LinearERC721Voting,
-        this.linearERC721VotingAddress!,
-        'setAzorius', // contract function name
-        [this.predictedAzoriusAddress],
-        0,
-        false,
-      );
+      if (!this.predictedAzoriusAddress || !this.linearERC721VotingAddress) {
+        throw new Error('Linear ERC721 addresses not set');
+      }
```
adamgall, line 213: Can you split this into two conditions, for error message accuracy?

src/models/AzoriusTxBuilder.ts
```diff
@@ -373,39 +419,40 @@ export class AzoriusTxBuilder extends BaseTxBuilder {
...
-  public signatures = (): string => {
-    return (
-      '0x000000000000000000000000' +
+  public signatures = (): Hex => {
+    return ('0x000000000000000000000000' +
       this.multiSendCallOnly.slice(2) +
       '0000000000000000000000000000000000000000000000000000000000000000' +
-      '01'
-    );
+      '01') as Hex;
```
adamgall, line 455: Instead of casting `as Hex`, can you do something like this instead?

```ts
  public signatures = (): Hex => {
    const signatureData =
      '0x000000000000000000000000' +
      this.multiSendCallOnly.slice(2) +
      '0000000000000000000000000000000000000000000000000000000000000000' +
      '01';
    if (!isHex(signatureData)) {
      throw new Error('Invalid signature data');
    }
    return signatureData;
  };
```

src/models/FreezeGuardTxBuilder.ts
```diff
@@ -126,60 +137,88 @@ export class FreezeGuardTxBuilder extends BaseTxBuilder {
...
   public buildSetGuardTx(abi: Abi, address: Address): SafeTransaction {
-    return buildContractCall(abi, address, 'setGuard', [this.freezeGuardAddress], 0, false);
+    return buildContractCall({
+      target: address,
+      encodedFunctionData: encodeFunctionData({
+        functionName: 'setGuard',
+        args: [this.freezeGuardAddress],
+        abi: abi,
+      }),
+    });
```
adamgall, line 191: There's no abi type safety here. It looks like there is only one usage of this function, and it passes the azorius ABI into this call. Can you just remove the `abi` param from this function and hardcode in the Azorius ABI?

or, you could just change the type of the `abi` param from `Abi` to `typeof abis.Azorius`.

Either way you'll notice that `this.freezeGuardAddress` might be undefined and get a nice red squiggle.

src/models/FreezeGuardTxBuilder.ts
```diff
@@ -126,60 +137,88 @@ export class FreezeGuardTxBuilder extends BaseTxBuilder {
...
-    );
+    if (!this.freezeGuardCallData) {
+      throw new Error('Freeze guard call data not set');
+    }
+
+    return buildContractCall({
+      target: this.zodiacModuleProxyFactory,
+      encodedFunctionData: encodeFunctionData({
+        functionName: 'deployModule',
+        args: [this.getGuardMasterCopyAddress(), this.freezeGuardCallData!, this.saltNum],
```
adamgall, line 218: Can remove the "!" on this line

src/models/MultisigTxBuilder.ts
```diff
@@ -18,27 +18,25 @@ export class MultisigTxBuilder {
...
-    return (
-      '0x000000000000000000000000' +
+  public signatures = (): Hex => {
+    return ('0x000000000000000000000000' +
       this.multiSendCallOnlyAddress.slice(2) +
       '0000000000000000000000000000000000000000000000000000000000000000' +
-      '01'
-    );
+      '01') as Hex;
   };
```
adamgall, line 26: Same suggestion as in the other extremely similar `signatures()` function in `AzoriusTxBuilder.ts`.

What do you think of just moving this over to (one of the two lmao) `crypto.ts` helper/utils files?

adamgall (changes requested):

src/models/FreezeGuardTxBuilder.ts
DarksightKellar: Removed the param to use Azorius abi directly

src/models/MultisigTxBuilder.ts
DarksightKellar: I moved this over into the crypto file, as `buildSignatures`, and had both `AzoriusTxBuilder` and  `MultisigTxBuilder` route their `signatures` calls through it.

Also merged the 2 `crypto.ts` files.

src/helpers/crypto.ts
```diff
@@ -219,3 +207,111 @@ const encodeMetaTransaction = (tx: MetaTransaction): string => {
...
+      parametersFixed!!.push([]);
+    } else if (param.startsWith('[') && param.endsWith(']')) {
+      parametersFixed!!.push(
+        param
+          .substring(1, param.length - 1)
+          .split(',')
+          .map(p => (p = p.trim())),
+      );
+    } else if (param.startsWith('(')) {
+      // This is part of tuple param, we need to re-assemble it. There should be better solution to this within splitIgnoreBrackets with regex.
```
DarksightKellar, line 263: This stuff is brought in over from another `crypto.ts` file.

src/helpers/crypto.ts
```diff
@@ -203,3 +204,16 @@ const encodeMetaTransaction = (tx: MetaTransaction): string => {
 export const encodeMultiSend = (txs: MetaTransaction[]): Hex => {
   return `0x${txs.map(tx => encodeMetaTransaction(tx)).join('')}`;
 };
+
+export const buildSignatures = (multiSendCallOnlyAddress: Address): Hex => {
+  const signatureData = ('0x000000000000000000000000' +
+    multiSendCallOnlyAddress.slice(2) +
+    '0000000000000000000000000000000000000000000000000000000000000000' +
+    '01') as Hex;
```
adamgall, line 215: ah one more nitpick, remove the `as Hex` from this line

adamgall (approved):

PR #2917, opened 2025-04-18
chore(types): Enforce consistency for Account Abstraction config

adamgall:
Closes [ENG-689](https://linear.app/decent-labs/issue/ENG-689/add-stronger-type-safety-between-account-abstraction-contracts-and)

## PR Description

This PR introduces explicit configuration for Account Abstraction (AA) support on a per-network basis and refactors the `NetworkConfig` type for improved type safety.

### Changes

1.  **Added `accountAbstractionSupported` Flag:**
    *   A new boolean flag, `accountAbstractionSupported`, has been added to the `NetworkConfig` type (`src/types/network.ts`).
    *   Each network configuration file (`src/providers/NetworkConfig/networks/*.ts`) has been updated to include this flag, specifying whether AA is supported on that network (e.g., `true` for Mainnet and Sepolia, `false` for Base, Optimism, Polygon).

2.  **Refactored `NetworkConfig` Type:**
    *   The `NetworkConfig` type in `src/types/network.ts` has been refactored into a discriminated union based on the `accountAbstractionSupported` flag.
    *   **If `accountAbstractionSupported` is `true`:** The type requires the `contracts.accountAbstraction` object and the `bundlerMinimumStake` property to be defined.
    *   **If `accountAbstractionSupported` is `false`:** The type ensures that `contracts.accountAbstraction` and `bundlerMinimumStake` are undefined.

### Benefits

*   **Improved Type Safety:** Makes it explicit at the type level which networks support AA features, preventing potential runtime errors when accessing AA-specific properties like `contracts.accountAbstraction` or `bundlerMinimumStake` on unsupported networks.
*   **Clearer Configuration:** Provides a clear and explicit way to enable or disable AA features for each network.

This change ensures that components relying on AA features can safely access the necessary configuration details only when supported by the current network.

### DIFF
--- a/src/providers/NetworkConfig/useNetworkConfigStore.ts
+++ b/src/providers/NetworkConfig/useNetworkConfigStore.ts
@@ -3,10 +3,10 @@ import { NetworkConfig } from '../../types/network';
 import { networks } from './networks';
 import { mainnetConfig } from './networks/mainnet';
 
-interface NetworkConfigStore extends NetworkConfig {
+type NetworkConfigStore = NetworkConfig & {
   getConfigByChainId: (chainId?: number) => NetworkConfig;
   setCurrentConfig: (config: NetworkConfig) => void;
-}
+};
 
 export const supportedNetworks = Object.values(networks).sort((a, b) => a.order - b.order);
 export const supportedEnsNetworks = Object.values(networks)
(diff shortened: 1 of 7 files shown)


src/providers/NetworkConfig/useNetworkConfigStore.ts
```diff
@@ -3,10 +3,10 @@ import { NetworkConfig } from '../../types/network';
 import { networks } from './networks';
 import { mainnetConfig } from './networks/mainnet';
 
-interface NetworkConfigStore extends NetworkConfig {
+type NetworkConfigStore = NetworkConfig & {
   getConfigByChainId: (chainId?: number) => NetworkConfig;
   setCurrentConfig: (config: NetworkConfig) => void;
-}
+};
```
adamgall, line 9: This was a necessary change, because `NetworkConfig` is no longer static.

PR #2919, opened 2025-04-18
fix(i18n): Explicitly specify namespaces for useTranslation

adamgall:
Fixes [ENG-691](https://linear.app/decent-labs/issue/ENG-691/localization-string-parsing-issues-in-various-places)

Updates various components and helpers to explicitly provide the relevant namespace(s) when calling the `useTranslation` hook from `react-i18next`.

Previously, some calls relied on the default namespace. Specifying namespaces improves clarity, helps with managing translations, and potentially enables better code splitting or loading strategies for translation files.

I asked the AI to help identify which "pages" are affected by these changes, and this is what it gave me, which looks more or less accurate:

> In summary, the pages most likely affected are:
> - DAO Proposal Dapps Page
> - DAO Proposal Dapp Detail Page
> - DAO Proposal Templates Page
> - DAO Activity/Transactions Page
> - DAO Proposal List Page
> - DAO Proposal Detail Page (both Azorius and Snapshot variants)
> - Pages with sortable lists (like Treasury/Assets, Members if they exist and use this component)
> - Potentially parts of the DAO Dashboard or Settings if they use components like EtherscanLink or ActivityAddress.
> - Technically all pages are affected by the TopErrorFallback change, but only during error states.

### DIFF
--- a/src/components/Activity/ActivityAddress.tsx
+++ b/src/components/Activity/ActivityAddress.tsx
@@ -14,7 +14,7 @@ export function ActivityAddress({
   addComma?: boolean;
 }) {
   const { displayName } = useGetAccountName(address);
-  const { t } = useTranslation();
+  const { t } = useTranslation('common');
   return (
     <EtherscanLink
       type="address"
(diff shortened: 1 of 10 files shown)


cbajandas (changes requested):
Spotted this place in the app (/proposal-templates) where this broke with this PR:

image

It's fine on develop but messed up on the branch.

cbajandas (approved):

PR #2930, opened 2025-04-20
`[ENG-698]` fix(strategy): Explicitly check for AA config when setting up gasless Hats strategies

adamgall:
Fixes [ENG-698](https://linear.app/decent-labs/issue/ENG-698/ensure-aa-config-exists-before-setting-up-gasless-hats-strategies)

Refactors the logic for encoding setup parameters for Hats-whitelisted voting strategies (both ERC20 and ERC721) in `useCreateRoles` and `SafePermissionsCreateProposal`.

Previously, the code implicitly relied on a non-null assertion (`!`) on the `accountAbstraction` configuration when `gaslessVotingFeatureEnabled` was true. This could lead to runtime errors if the feature was enabled on a network lacking the necessary AA configuration, without an explicit check beforehand.

This change introduces an explicit check:
1.  If `gaslessVotingFeatureEnabled` is true, it now verifies that `accountAbstraction` configuration exists.
2.  If AA config is missing when gasless is enabled, it throws a clear error (`Account abstraction not found`).
3.  If both are present, it uses the V1 (gasless-compatible) setup parameters.
4.  If `gaslessVotingFeatureEnabled` is false, it uses the standard setup parameters.

This ensures more robust error handling and guarantees that the V1 strategy setup (requiring `lightAccountFactory`) is only attempted when both the feature flag is enabled and the required network configuration is present. It aligns the logic for handling this scenario consistently across `useCreateRoles` and `SafePermissionsCreateProposal`.

### DIFF
--- a/src/hooks/utils/useCreateRoles.ts
+++ b/src/hooks/utils/useCreateRoles.ts
@@ -186,31 +186,41 @@ export default function useCreateRoles() {
             allowFailure: false,
           });
 
-        if (gaslessVotingFeatureEnabled && !accountAbstraction) {
-          throw new Error('Account abstraction is not enabled');
+        let encodedStrategyInitParams: Hex;
+
+        if (gaslessVotingFeatureEnabled) {
+          if (!accountAbstraction) {
+            throw new Error('Account abstraction not found');
+          }
+          encodedStrategyInitParams = encodeAbiParameters(
+            parseAbiParameters(linearERC20VotingWithWhitelistV1SetupParams),
+            [
+              safeAddress, // owner
+              votesToken.address, // governance token
+              moduleAzoriusAddress, // Azorius module
+              existingVotingPeriod,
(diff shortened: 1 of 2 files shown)


DarksightKellar (approved):

adamgall:
> It seems previous implementation has `gaslessVotingFeatureEnabled && !accountAbstraction` check already? So the changes here is using `if...else` instead of conditional operator?

The ultimate goal was to remove the `!` from the `accountAbstraction!.lightAccountFactory` usage.

But on a call with @DarksightKellar and @Da-Colon just now, we noticed there were some bugs in both the original implementation and the one that you just reviewed. I just pushed a commit to fix them.

We _should not be throwing errors_ in this code at all. We were previously throwing an error if the "gasless voting feature flag" is on, but the network config doesn't support gasless voting. That's not what should happen.

The latest commit fixes these bugs.

PR #2934, opened 2025-04-21
Feat/eng 700 remove gasless voting kvpairs key

adamgall:
Fixes [ENG-700](https://linear.app/decent-labs/issue/ENG-700/change-the-meaning-of-gasless-toggle)

This is WIP.

Although the code compiles, it's buggy. Logic is not right.

Before this work can continue, code needs to be written (in another PR), which determines the state of the Paymaster (based on one of the 5 discrete states that the "Paymaster feature" can be in, as per the PRD).

After that code exists, it makes sense to pick this back up.

### DIFF
--- a/src/hooks/utils/useCreateRoles.ts
+++ b/src/hooks/utils/useCreateRoles.ts
@@ -102,7 +102,7 @@ export default function useCreateRoles() {
       moduleAzoriusAddress,
     },
   } = useStore({ daoKey });
-  const { safe, subgraphInfo, gaslessVotingEnabled, paymasterAddress } = useDaoInfoStore();
+  const { safe, subgraphInfo, paymasterAddress } = useDaoInfoStore();
   const { hatsTree, hatsTreeId, getHat } = useRolesStore();
   const {
     addressPrefix,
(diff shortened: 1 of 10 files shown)


src/hooks/utils/useCreateRoles.ts
```diff
@@ -255,7 +255,7 @@ export default function useCreateRoles() {
         };
 
         const optionallyWhitelistWhitelistingStrategyOnPaymaster = [];
-        if (gaslessVotingEnabled && paymasterAddress) {
+        if (gaslessVotingFeatureEnabled && paymasterAddress) {
```
adamgall, line 258: This is the line that caused me to pause on this PR.

Previously, we were deciding to whitelist the new strategy on the paymaster based on whether gasless voting was enabled on the DAO or not, which came from the KVPairs flag.

But we don't have that any more (this PR removes that flag/state), and we don't have anything to replace it with yet.

Need to write the code that can replace the KVPairs flag before we rip out the KVPairs flag.

adamgall:
Stale, closing

PR #2936, opened 2025-04-22
`[ENG-714]` release tested dApps

twodam:
Include:
* Origin Unified Defi
* Balancer
* Balancer Pool Creator
* Spark
* ParaSwap
* Jumper Exchange

Metatdata fetched from https://safe-client.safe.global/v1/chains/1/safe-apps

image

### DIFF
--- a/src/assets/dapps.json
+++ b/src/assets/dapps.json
@@ -1,12 +1,91 @@
 [
   {
-    "id": 1,
+    "id": 74,
     "url": "https://swap.cow.fi",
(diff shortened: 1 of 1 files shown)


src/assets/dapps.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "id": 1,
+    "id": 74,
```
twodam, line 3: using id from safe-app json for now.

PR #2940, opened 2025-04-22
`[ENG=679]` feat: Enhance transaction builder action card messages

Da-Colon:
## Screenshots
image

image

- actions.json: Add localized strings for single and multiple transaction actions
- ProposalActionCard.tsx: Update rendering logic to use new localized strings


Da-Colon:
Still very much open to suggestions. @cbajandas @johnqh

PR #2946, opened 2025-04-23
`[ENG-547]` use number of unit(day/hour/minute) for duation

twodam:
### Summary

- new component `DurationUnitStepperInput` to accept input based on time units while returning seconds value
- replace `BigIntInput` with new component for "Tranche Duration"

#### Duration input based on time unit
image

#### Preview duration is still based on seconds, no change
image


src/components/ProposalBuilder/ProposalStream.tsx
```diff
@@ -353,22 +355,22 @@ export function ProposalStream({
...
-                                <BigIntInput
-                                  isRequired
-                                  value={tranche.duration.bigintValue}
-                                  placeholder={(SECONDS_IN_DAY * 365).toString()}
-                                  decimalPlaces={0}
-                                  min={index === 0 ? '1' : undefined}
-                                  step={1}
-                                  onChange={value =>
+                                <DurationUnitStepperInput
+                                  minSeconds={trancheIndex === 0 ? 1 : 0}
```
twodam, line 359: it should be `trancheIndex` instead of `index (stream index)`.

> "trancheDurationHelperFirstTranche": "At least 1 second for the first tranche.",

PR #2956, opened 2025-04-24
`[ENG-722]` Can't submit stream proposal with more than one stream

twodam:
### Issue
`stream.tokenAddress` is required field, but it's by default empty with new stream added by user. So proposal creation is disabled when more than one stream is created.
### Reason
Asset dropdown menu display first token as selected, but the `tokenAddress` field is not synced with first token.
### Solution
So I set the default token address when adding new stream.


DarksightKellar (approved):

DarksightKellar:
Merge conflict's been resolved. Merging.

PR #3000, opened 2025-05-05
`[ENG-798]` Update recipient amount with new decimal when switching to different asset

twodam:


DarksightKellar (approved):
This was an easy fix but was it an easy _find_? Curious

twodam:
> This was an easy fix but was it an easy _find_? Curious

it took some time to reproduce the issue, because the issue only occurred with switching to asset with different decimal (like ETH to USDC/WBTC).
and then it's easy to add a log to print out the `totalAmount` and find out what's the issue.

PR #3039, opened 2025-05-19
`[ENG-883]` Update to Etherscan V2 by May 31, 2025

twodam:
### Summary

- Update url of etherscan API to V2 which unify all chain access with same url
- All chain will use same `VITE_APP_ETHERSCAN_MAINNET_API_KEY` for Etherscan API V2, no longer need to specify different key for different network.
- Migration doc: https://docs.etherscan.io/etherscan-v2/v2-quickstart

image

### DIFF
--- a/.env
+++ b/.env
@@ -59,16 +59,8 @@ VITE_APP_WALLET_CONNECT_PROJECT_ID=""
 # VITE_APP_HOTJAR_VERSION=""
 # Sentry DSN URL, not used locally
 # VITE_APP_SENTRY_DSN_URL=""
-# ABI selector, used on Base
-# VITE_APP_ETHERSCAN_BASE_API_KEY=""
-# ABI selector, used on Mainnet
+# ABI selector, Etherscan API V2 key, used on all supported Chains
 # VITE_APP_ETHERSCAN_MAINNET_API_KEY=""
-# ABI selector, used on Optimism
-# VITE_APP_ETHERSCAN_OPTIMISM_API_KEY=""
-# ABI selector, used on Polygon
-# VITE_APP_ETHERSCAN_POLYGON_API_KEY=""
-# ABI selector, used on Sepolia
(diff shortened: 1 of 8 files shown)


adamgall (commented):
It seems like we can consolidate down to just one api key, right? Like we no longer need a different key for each network, we can just use the "mainnet" key for all networks now.

twodam:
> It seems like we can consolidate down to just one api key, right? Like we no longer need a different key for each network, we can just use the "mainnet" key for all networks now.

updated all to use `VITE_APP_ETHERSCAN_MAINNET_API_KEY`

.env
```diff
@@ -59,16 +59,8 @@ VITE_APP_WALLET_CONNECT_PROJECT_ID=""
 # VITE_APP_HOTJAR_VERSION=""
 # Sentry DSN URL, not used locally
 # VITE_APP_SENTRY_DSN_URL=""
-# ABI selector, used on Base
-# VITE_APP_ETHERSCAN_BASE_API_KEY=""
-# ABI selector, used on Mainnet
+# ABI selector, Etherscan API V2 key, used on all supported Chains
 # VITE_APP_ETHERSCAN_MAINNET_API_KEY=""
```
adamgall, line 63: Suggest to rename this to remove "mainnet" from the name

adamgall (approved):
Approving but have one more suggestion about renaming the remaining env var. also, I assume this has been manually tested?

.env
twodam: ha, I was hesitated to do this which may break builds on cloud. but yeah we could do this to clarify more.

twodam: yeah, I tried with Mainnet, Base and Polygon for a quick ABI test with random contract.

adamgall: Oh yeah you're right about needing to update the vars on Cloudflare. Forget it for now

PR #3050, opened 2025-05-22
`[ENG-909 ~ Release]` Gasless voting Various fixes

Da-Colon:


src/components/Proposals/ProposalActions/CastVote.tsx
```diff
@@ -95,7 +95,7 @@ export function CastVote({ proposal }: { proposal: FractalProposal }) {
 
   const castGaslessVoteCallback = useCallback(async () => {
     (async () => {
-      if (!canVoteForFree || !selectedVoteChoice) {
+      if (!canVoteForFree || selectedVoteChoice === undefined) {
```
DarksightKellar, line 98: This goddamned bug kept popping up when I was building and testing the settings edit form. I think it's time we enforced some explicitness about what we're checking.

src/hooks/DAO/proposal/useCastVote.ts
```diff
@@ -245,6 +245,10 @@ const useCastVote = (proposalId: string, strategy: Address) => {
         return;
       }
 
+      if (typeof canCastGaslessVote === 'boolean') {
+        return;
+      }
```
DarksightKellar, line 250: Huh?

Oh do you mean `if (canCastGaslessVote !== undefined)`?
What's this solving exactly? Some sorta render loop?

DarksightKellar: Talked with David. This is the way to do it lol. `if (canCastGaslessVote === undefined)` will skip the first and all subsequent renders. `if (canCastGaslessVote !== undefined)` will not solve the re-render problem. Interesting edge case here ha.

DarksightKellar (approved):

PR #3054, opened 2025-05-23
`[ENG-501]` use Viewer with same package as current markdown editor

twodam:
### Summary

Replace `ReactMarkdown` viewer to `Viewer` from `@toast-ui/react-editor`, to keep editor and viewer are from same package/library.

### Screenshots (Before vs Now)

image
image
image


DarksightKellar (approved):
lgtm

PR #3071, opened 2025-05-28
`ENG-826` | Remove Modal.NONE

DarksightKellar:
While re-working the modal experience, I did not like the way ModalType.NONE was being shoehorned into a "close modal" functionality. This has led to scenarios where visually there's no modal on-screen, but there actually IS one, just invisible -- NONE, thus causing other attempts to open some other modal to simply not work.

I've attempted to fix this in the past by closing the modal in the NONE switch handler. This PR attempts a more comprehensive and explicit fix.

I've completely removed the NONE type, and exposed an explicit `close` function from `useDecentModal`.


Da-Colon:
Have you had a chance to test this and is this ready for review now?

Da-Colon (approved):
Nice

PR #3072, opened 2025-05-28
Fix modal behavior

DarksightKellar:
Addresses this comment: https://linear.app/decent-labs/issue/ENG-842/fix-permissions-settings-ux-flows#comment-3b5a4b77

### DIFF
--- a/src/components/ui/modals/AddStrategyPermissionModal.tsx
+++ b/src/components/ui/modals/AddStrategyPermissionModal.tsx
@@ -62,6 +62,7 @@ export function AddStrategyPermissionModal({
             backgroundColor: 'white-alpha-04',
           }}
           onClick={() => {
+            closeModal();
             if (!isSettingsV1Enabled) {
               navigate(
                 DAO_ROUTES.settingsPermissionsCreateProposal.relative(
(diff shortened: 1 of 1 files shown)


src/pages/dao/settings/permissions/SafePermissionsSettingsContent.tsx
```diff
@@ -97,7 +94,8 @@ export function SafePermissionsSettingsContent() {
             >
               <BarLoader />
             </Box>
-          ) : (!votesToken || noErc20Address) && (!erc721Tokens || !noErc721Address) ? (
```
Da-Colon, line 100: This doesn't mess with DAOs who have already whitelisted view of the permissions settings?

Da-Colon (approved):

src/pages/dao/settings/permissions/SafePermissionsSettingsContent.tsx
DarksightKellar: Well adding the `--WithHatsWhitelistingAddress` to the checks was introduced by me and led to the bug in the first place. This diff is just reverting to prod. It appears whitelisting permissions were never shown here, and can only be "seen" by going to the role that has create proposal permission.

PR #3073, opened 2025-05-30
`[ENG-1001]` Bugfix, Safe's updated api route and missing response property

Da-Colon:
So discovered that Safe's Api has updated its `multsig-transactions` and `data-decoder` routes

`multsig-transactions` no longer returns the `dataDecoded` property in its returns, You must secondly request this information.

`data-decoder` route has updated, or maybe we were just missing it, but is suppose to include the `to` value in the `body` of the post request, this was missing, not sure why/how this was working before or sometimes without it, as per the docs.


Da-Colon:
I just saw that on PROD that proposals are not loading. I believe this bug is the cause. I need to switch this to a hotfix release tomomorrow.

PR #3124, opened 2025-06-11
`[ENG-1143 ~ Release]` Phew Roles loading updated

Da-Colon:
Some more testing revealed a little more left to do. I opted to push this closer the pattern in the global store without actually moving and refactoring code into the global store.

### DIFF
--- a/src/store/roles/useRolesStore.ts
+++ b/src/store/roles/useRolesStore.ts
@@ -48,17 +48,22 @@ const useRolesStore = create<RolesStore>()((set, get) => ({
     return matches[0];
   },
   setHatKeyValuePairData: args => {
-    const { hatsTreeId, contextChainId, streamIdsToHatIds } = args;
+    const { hatsTreeId, contextChainId, streamIdsToHatIds, daoKey } = args;
     for (const { hatId, streamId } of streamIdsToHatIds) {
       streamIdToHatIdMap.set(streamId, hatId);
     }
-    set(() => {
-      // if `hatsTreeId` is null or undefined,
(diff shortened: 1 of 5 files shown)


src/store/roles/useRolesStore.ts
```diff
@@ -48,17 +48,22 @@ const useRolesStore = create<RolesStore>()((set, get) => ({
...
+    set(state => {
+      if (hatsTreeId === null || hatsTreeId === undefined) {
+        return {
+          ...initialHatsStore,
+        };
       }
-      return { hatsTreeId, streamsFetched: false, contextChainId };
+      return {
+        ...state,
+        hatsTreeId: { [daoKey]: hatsTreeId },
```
DarksightKellar, line 63: Looks like there'll only ever be one instance of `{ 'daoKey': 'treeId' }`, and every time `setHatKeyValuePairData`, it's overwritten with `{ 'daoKeyNew': 'treeIdNew' }`?

Assuming this is a tighter fix so that `hatsTreeId[daoKey]` will always return the latest set data? 

Fine with this until the move to global store

DarksightKellar (approved):

src/store/roles/useRolesStore.ts
Da-Colon: wdym? I think yes thats more or less correct. the typing is kinda weird for hatsTreeId as it travels from. even with conditionals undefined and null can sneak through for some reason. I can discuss in person but yeah it all works now and it sort of mimics the normal store now at least with this hatTreeId.

PR #3129, opened 2025-06-12
`[ENG-1148]` Fix first load issues

Da-Colon:
So this solves a few different problems (I'll try to find the tickets if they exist). But basically in the first second the app loads in. its default network configuration is set to mainnet (currently no way around it unless we allow the network to be `undefined` for a second while the URL is checked. 

These two updates allows the information gathers (publicClient and safeAPI) to settle on the correct network when refreshing or navigating straight to a DAO that isn't on `mainnet`


DarksightKellar (approved):
Suggestion: set up

PR #3156, opened 2025-06-20
`[ENG-1215]` Update Calendar with New UI

Da-Colon:
new calendar

### DIFF
--- a/src/assets/css/Calendar.css
+++ b/src/assets/css/Calendar.css
@@ -1,6 +1,6 @@
 .react-calendar {
-  max-width: 100%;
-  background: #221d25;
+  width: 265px;
+  backdrop-filter: blur(10px);
   border-radius: 0.75rem;
   border: 0;
   font-family: 'DM Sans';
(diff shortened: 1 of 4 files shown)


src/assets/css/Calendar.css
Da-Colon, line 1: Went through and:

- Updated color values to using theme values via CSSVar
- Removed unused CSS classes
- Adjust all the styles to match the current designs

src/components/ui/forms/DatePicker.tsx
Da-Colon, line 1: This component got a work over, There was a bug in our the `useDisclosure` was being used outside the context of the menu. 

- I removed all the old extra UI components 
- rearranged the way the components were structured So that the Desktop and Mobile views are
- I removed the 'mobile' view the desktop view is now okay on mobile + we aren't expecting folks to use with mobile. 
image

src/components/ui/forms/DatePicker.tsx
```diff
@@ -182,51 +154,33 @@ export function DatePicker({
...
-          minDate={minDate}
-          maxDate={maxDate}
-          formatShortWeekday={(_, date) => date.toString().slice(0, 2)}
-          prevLabel={<Icon as={CaretLeft} />}
-          nextLabel={<Icon as={CaretRight} />}
-          next2Label={null}
-          prev2Label={null}
-          tileContent={({ date }) =>
-            isToday(date) ? <TodayBox isTodaySelected={isTodaySelected} /> : null
-          }
```
Da-Colon, line 226: See above but this is now `null` as now this is handled via the Calendar.css

twodam (approved):

PR #3169, opened 2025-07-01
`[ENG-1189]` fix: enable tooltip only when nonce is not active 

Da-Colon:


DarksightKellar (approved):
how'd this happen lol? Leftover from previous debugging efforts maybe?
EDIT: oh wait that's "false" to isDisabled not "true". nvm

PR #3179, opened 2025-07-09
 Added logic to pop modals and close all modals in nested scenarios.

DarksightKellar:


twodam (changes requested):
I’m not convinced this approach is the right direction. It feels like a quick and dirty solution, which I personally don’t prefer. Let’s hold off on it for now until I have a better understanding of [the issue](https://decentdao.slack.com/archives/C06QZ68L5SA/p1752055172002399).

Da-Colon:
Closing to work on closer later

PR #3183, opened 2025-07-10
[ENG-1277] Fix automation workflow for `pull` trigger

cbajandas:
This makes two changes:

1. Updates the workflow to make it so that the test results comment can be posted on a PR post merge. The original workflow was written expecting the trigger to be `on: pull-request`, making the target of the comment implicit. The update here attempts to locate the PR associated with a push and adds the comment to it, retroactively.
2. Updates the workflow to add a link to the build artifacts to the test results comment. I think this might be more necessary with the workflow not being used as a check on the PR. If I'm wrong about this, the change will still make it easier and is worthwhile if it works.

### DIFF
--- a/.github/workflows/ui-automation.yaml
+++ b/.github/workflows/ui-automation.yaml
@@ -46,9 +46,13 @@ jobs:
           commit: ${{ github.sha }}
           state: merged
 
+      - name: Prepend artifact link to summary
+        run: |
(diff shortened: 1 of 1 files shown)


Da-Colon (approved):

cbajandas:
I am hoping that we'll know this worked when this PR is merged, as hypothetically that will trigger the workflow. Though perhaps it won't go off until the next PR is merged.

PR #3186, opened 2025-07-10
[ENG-1277] Use alternative github action to find pull request

cbajandas:
This may not work given that we delete the branch after it is merged. It's a matter of timing, I suppose. If the action finds the PR before the branch is deleted, then it may still work. But somehow I doubt it. It's worth a shot before going back and totally rethinking how this is going to work.

### DIFF
--- a/.github/workflows/ui-automation.yaml
+++ b/.github/workflows/ui-automation.yaml
@@ -9,6 +9,8 @@ jobs:
     name: E2E Selenium Automation
     runs-on: ubuntu-latest
     continue-on-error: true
+    permissions:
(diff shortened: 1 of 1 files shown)


cbajandas:
I realized I could test this more effectively on the automation repo. I'll post a set of changes that I've at least confirmed works on that repo before I ask for another review/merge.

cbajandas:
Post merge automation is stupid.

Since this was sitting collecting dust, I went back and figured out how to do pre merge automation which is the gold standard (as far as I am concerned). New PR incoming.

PR #3192, opened 2025-07-18
[`ENG-1297`] Update automation workflow to run automation pre merge

cbajandas:
Adding new approach to the automation workflow. This uses a script to locate the preview url for a PR and runs the automation against that. I believe this will be a check on the PR versus a hidden action like before, meaning it will run whenever there are new commits to the branch associated with the PR.

What I am hoping here is that since this doesn't require a merge, I can test it with a PR draft before submitting it for review since I can't test this on the automation repo (it doesn't have any deployment to key off of).


jigglyjams (approved):
I don't fully understand how this fits in with your other automation stuff but the table looks great and would like to see it in my PR :)

PR #3201, opened 2025-07-28
`[ENG-1301]` Bundle staking transactions

twodam:
Depends on #3200, pls merge it first and then this.

### Summary

- Deploy staking contract if not existed
- Update related values if existed


src/components/ui/utils/AssetSelector.tsx
```diff
@@ -15,7 +15,10 @@ interface AssetSelectorProps {
   onlyNativeToken?: boolean;
   onSelect?: (addresses: string[]) => void;
   canSelectMultiple?: boolean;
-  defaultSelectedAddresses?: string[];
+  /**
+   * Can't unselected these pre-selected assets
+   */
+  lockedSelections?: string[];
```
jigglyjams, line 21: did this change to `lockedSelections` because you cannot remove them at the contract level?

jigglyjams (approved):

src/components/ui/modals/SafeSettingsModal.tsx
```diff
@@ -1191,13 +1197,147 @@ export function SafeSettingsModal({
...
+      throw new Error('Staking are not set');
+    }
+    const stakingValues = updatedValues.staking!;
+
+    const stakingContract = governance.stakedToken;
+
+    const changeTitles = [];
+    const transactions: CreateProposalTransaction[] = [];
+
+    if (stakingValues.deploying) {
```
jigglyjams, line 1219: should we check that there isnt one already deployed as well?

src/components/ui/utils/AssetSelector.tsx
twodam: yes

PR #3217, opened 2025-08-06
`[ENG-1361]` prevent default so dropdown works properly

jigglyjams:
Before you could click anywhere within the assert selector div to activate it, this made it hard to close the asset selector

Was being caused by the label

image

### DIFF
--- a/src/components/ui/forms/LabelWrapper.tsx
+++ b/src/components/ui/forms/LabelWrapper.tsx
@@ -25,7 +25,11 @@ function LabelWrapper({
 }: LabelWrapperProps) {
   return (
     <Box position="relative">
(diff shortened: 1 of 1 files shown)


cbajandas:
omg, the automation caught a bug.

On the preview branch, you can't click the radio buttons to switch between governance types during DAO creation:
image

Those options are right below the 'Network' dropdown menu which uses a component that looks suspiciously like the one we've been messing with.

twodam:
Maybe we could move AssetSelector out of LabelWrapper to avoid modifying the shared component, which might affect other top-level components.

twodam:
close for now, we can have new PR once resolved.

PR #3254, opened 2025-08-25
`[ENG-1361]` fix AssetSelector click area by moving it out of LabelComponent wrapper

twodam:
### Summary

To avoid click area too big for AssetSelector inside Staking setting tab, move it out of label wrapper.


jigglyjams (approved):
Beautiful fix

PR #3261, opened 2025-08-27
`[ENG-1397]` Implement stake/unstake logic

Da-Colon:
Note: Something is up with listeners, will have to refresh after submiting staking/unstaking transactions. approval should update automatically


src/components/DaoStaking/StakeCard.tsx
```diff
@@ -123,15 +136,57 @@ export default function StakeCard() {
   const { t } = useTranslation('staking');
   const { daoKey } = useCurrentDAOKey();
   const {
-    governance: { stakedToken, votesToken },
+    governance: { stakedToken, votesToken, erc20Token },
   } = useDAOStore({ daoKey });
+  const unstakedToken = erc20Token ?? votesToken;
```
twodam, line 141: Would it be better to determine which token to use based on the governance type?

twodam (changes requested):

twodam (approved):

PR #3281, opened 2025-09-03
`[ENG-1444]` Merge Approval into single click

Da-Colon:


src/components/DaoStaking/StakeCard.tsx
```diff
@@ -224,24 +198,67 @@ export default function StakeCard() {
...
+          client: walletClient,
+        });
+        // Use multicall to approve and stake in single transaction
+        contractCall({
+          contractFn: () => tokenContract.write.approve([stakedToken.address, amount.bigintValue!]),
+          pendingMessage: t('approvePending', { amount: formattedAmount, symbol: tokenSymbol }),
+          successMessage: t('approveSuccess', { amount: formattedAmount, symbol: tokenSymbol }),
+          failedMessage: t('approveError'),
+          // Approve and stake in single transaction
+          successCallback: async () => {
```
twodam, line 218: Aren’t those actually two transactions when you put `contractCall` into the `successCallback`? That would require the user to sign twice. Should we consider bundling them into a single `multicall`?

src/components/DaoStaking/StakeCard.tsx
```diff
@@ -224,24 +198,67 @@ export default function StakeCard() {
...
+                successMessage: t('stakingSuccess', {
+                  amount: formattedAmount,
+                  symbol: tokenSymbol,
+                }),
+                failedMessage: t('stakingError'),
+                successCallback: () => {
+                  formikHelpers.resetForm();
+                },
+              });
+            }, 2300);
```
twodam, line 235: Why do we need a timeout when it’s already wrapped in the success callback?

twodam (changes requested):

jigglyjams (commented):
The `Available` on the `Stake` tab does not update once the approve and stake actions have been completed. The  `Available` on the `Unstake` tab does update

src/components/DaoStaking/StakeCard.tsx
Da-Colon: Also good call out. So inside of the `contractCall` there is a timeout before the `pendingContractCall` is switched to false. So when I do a second contract call instead of switching to `true` again. it switches to `false` because of the inside timeout. 

Let me revisit this and see if I can come up with a better solution, I didn't want to touch the inside of `useTransaction` for this use case.

Da-Colon:
> The `Available` on the `Stake` tab does not update once the approve and stake actions have been completed. The `Available` on the `Unstake` tab does update

weird updated for me

staking

src/components/DaoStaking/StakeCard.tsx
Da-Colon: sorry also no way to batch them into a single call. I'm going to update the `useTransactions` to remove that timeout.

Da-Colon: Ha actually. much better solution, thanks for challenging me. just use separate instances of `useTransaction` duh. smooth implementation now without timeout.

twodam: > sorry, also no way to batch them into a single call, in our current setup. SplitsClient does give us access to multisend contract on our supported networks but probably shouldn't rely on it beyond Splits feature.

I might be missing something, but since we’re invoking our own contract functions `approve` and `stake`, how does this relate to the Splits feature? Also, isn’t `multisend` not really part of Splits either?

PR #3287, opened 2025-09-04
`[ENG-1443]` Bugfix `BigIntInput`

Da-Colon:
I actually had ran into this issue when I was using it for staking. `parentFormikValue` was added as a property with a optional status but not passing this value was breaking things. 

## Screenshot
localhost_3000_home_dao=sep%3A0x5BB57D7010e961620105af11b92A8E05d54BD28A page=1 size=10


cbajandas:
I've tried this out on the preview deployment a couple different ways and I am still seeing the same error. Everything seems to be up to date, though, so far as I can tell.

jigglyjams (approved):

PR #3318, opened 2025-09-16
[`ENG-1495`] Update 'Safe' to 'Project' or 'Treasury'

cbajandas:
This makes sweeping changes to the English language files to change the term 'Safe' to either 'Project' or 'Treasury', depending on the context.
* The more specific term 'Safe{Wallet}' was not modified
* In cases where the term 'child Safe' was used, the text was updated to 'child DAO' as part of what would be a two step process, such that the planned for change to the terms 'subDAO' and 'child DAO' are updated


cbajandas:
I did not mean to request copilot's review on this lol... thanks for the slightly more comprehensive summary, though, I guess.

cbajandas:
Roles test failures make sense because they apparently rely upon a label that uses the word "Safe":
```typescript
TimeoutError: Waiting for element to be located By(css selector, [aria-label="Toggle your favorite Safes."])
``` 
This can either be fixed by changing the label when this gets merged to develop to match the new terminology or updated in advance to something different.

